Issue #875, opened 2026-03-24
Migration tooling — self-hosted to hosted


msywulak:
## Summary

Build tooling for users to migrate from a self-hosted Atlas instance to the hosted SaaS at app.useatlas.dev. Export and import conversations, semantic layers, learned patterns, and settings.

## Context

Self-hosted Atlas is the free AGPL option. As the hosted SaaS launches, some self-hosted users will want to migrate to managed infrastructure. This should be a smooth, well-documented path — not a reason to stay self-hosted.

## Acceptance Criteria

- [ ] `atlas export` CLI command — dumps conversations, semantic layer YAMLs, learned patterns, settings to a portable format (JSON/tar)
- [ ] `atlas import` or hosted import endpoint — ingests export bundle into a hosted workspace
- [ ] Conversation history preserved (messages, metadata, timestamps)
- [ ] Semantic layer entities imported (with conflict resolution for entities that already exist)
- [ ] Learned patterns imported (approved patterns only)
- [ ] Settings mapped where applicable (model config, agent settings)
- [ ] Connection strings NOT exported (user re-enters credentials in hosted UI)
- [ ] Migration guide in docs with step-by-step walkthrough
- [ ] Dry-run mode that shows what would be imported without committing

PR #1199, opened 2026-04-03
fix: add export + migrate-import to completions test registry

msywulak:
## Summary
- CLI added `export` and `migrate-import` commands (via #875 migration tooling) but the bidirectional completions test wasn't updated
- Test expected 13 commands but found 15, breaking CI on main

## Test plan
- [x] `bun test packages/cli/src/__tests__/completions.test.ts` — 22 pass, 0 fail


msywulak:
Already fixed in PR #1204 (commit ebbcbc10). Issue #1200 is closed.

PR #1448, opened 2026-04-16
fix: scaffold types.ts — define unpublished exports locally

msywulak:
## Summary
- Deploy Validation CI is failing because `types.ts` re-exports `ADMIN_ROLES` and `ATLAS_MODES` from `@useatlas/types`, but the published `0.0.10` package doesn't include them yet (added in #1442)
- Turbopack validates all re-exports at build time, so the scaffold Docker build fails regardless of whether the exports are consumed
- Post-processes `types.ts` during template sync: replaces the combined re-export with local const definitions
- Also updates ROADMAP entry for scaffold layout override

## Test plan
- [ ] Deploy Validation CI passes (Scaffold docker + nextjs-standalone)
- [ ] Remove this workaround after publishing `@useatlas/types@0.0.11`


msywulak:
Auto-closed by #1458 (Closes #1448). Consumer refs now point at `@useatlas/types@^0.0.11`, which exposes `ADMIN_ROLES` / `ATLAS_MODES` / `ConnectionStatus` / `./mode` directly from npm — the template-sync workaround is no longer needed. Deploy Validation went green on #1458.

PR #1494, opened 2026-04-17
fix(auth): isolate middleware test from internal DB (#1483)

### LINKED ISSUE #1483: bug: pre-existing flaky middleware test 'mode managed with valid session returns authenticated'
## Summary
The middleware test `authenticateRequest() > mode 'managed' with valid session returns authenticated` (`packages/api/src/lib/auth/__tests__/middleware.test.ts:137`) intermittently fails when run in isolation:

```
expect(received).toBe(expected)
Expected: true
Received: false
```

## Detected on
Discovered on a clean checkout of `main@cf6dbc23` (no other changes) while running CI gates for #1472.

## Hypothesis
Likely a mock state leak — `mockValidateManaged.mockResolvedValueOnce(...)` may already be consumed by a previous test that triggered managed-mode validation. Worth confirming whether the test reproduces deterministically or only after specific test orderings.

## Reproduction
```bash
bun test packages/api/src/lib/auth/__tests__/middleware.test.ts
```

## Discovered via
CI gate run for the #1472 fix branch — unrelated to that change.

msywulak:
## Summary

Fixes the intermittent failure in `authenticateRequest() > mode 'managed' with valid session returns authenticated` (`packages/api/src/lib/auth/__tests__/middleware.test.ts:137`).

**Root cause (different from issue hypothesis):** The managed-auth path calls `checkSSOEnforcement` → `isSSOEnforcedForDomain`, which issues a real Postgres query whenever `DATABASE_URL` is set. When the DB isn't reachable (or transient), the fail-closed `catch` returns `{ authenticated: false, status: 500 }`, surfacing as `Expected: true / Received: false`. Mock state was a red herring — the validator mock is set up correctly via `mockResolvedValueOnce` and returns the authenticated result; the failure happens downstream in the SSO enforcement DB call.

**Fix:** Unset `DATABASE_URL` in the `authenticateRequest()` `beforeEach` (and restore in `afterEach`) so `hasInternalDB()` short-circuits and SSO enforcement returns `null` without network I/O. Mirrors the pattern already used in `chat.test.ts`, `query.test.ts`, `health.test.ts`, `dashboards.test.ts`.

## Test plan

- [x] Reproduced on `main` (Postgres not running): test fails with the exact `connect ECONNREFUSED 127.0.0.1:5432` `FiberFailure` shown in #1483.
- [x] After fix, ran 5 consecutive isolated invocations of the file — all 33 tests pass each time.
- [x] `bun run lint` ✓
- [x] `bun run type` ✓
- [x] `bun run test` ✓ (227 API files, 17 CLI, 56 web, 25 EE, 8 react — all pass)
- [x] `bun x syncpack lint` ✓
- [x] `SKIP_SYNCPACK=1 bash scripts/check-template-drift.sh` ✓

Closes #1483.

### DIFF
--- a/.claude/research/ROADMAP.md
+++ b/.claude/research/ROADMAP.md
@@ -1207,7 +1207,7 @@ Parent: #757. Replace per-platform interaction plugins with a single `@useatlas/
 - [x] Signup-connect error-isolation tests double-counting Next.js dev-tools alert (#1488, PR #1489 — scoped queries to `<main>`)
 - [x] Obsidian plugin doc — stale `/api/chat` → `/api/v1/chat` (#1491, surfaced by docs audit)
 - [x] Migrate `internalQuery` call sites from `Effect.promise` to `Effect.tryPromise` (#1468, PR #1493 — 37 sites across 15 routes; extracted `queryEffect<T>()` helper in `@atlas/api/lib/effect` with normalized catch; `makeQueryEffectMock` test helper; found during #1465 silent-failure review)
-- [ ] Pre-existing flaky middleware test `mode 'managed' with valid session returns authenticated` (#1483 — not a 1.2.0 blocker; milestone stripped, tracked standalone)
+- [x] Pre-existing flaky middleware test `mode 'managed' with valid session returns authenticated` (#1483 — fixed by isolating from internal DB so SSO enforcement check short-circuits)
 
 ---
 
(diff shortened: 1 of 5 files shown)


msywulak:
Addressed all review findings in b6320af7:

1. **Tightened the comment** to drop rename-fragile function names.
2. **Added `_setSSOEnforcementOverride()`** test-only hook in `middleware.ts` (mirrors existing `_setValidatorOverrides` pattern) so SSO branches can be tested without `mock.module` or a real DB.
3. **Added two SSO enforcement tests** in `middleware.test.ts`:
   - Managed-auth on an SSO-enforced domain blocks login with `403 + ssoRedirectUrl` propagated.
   - SSO enforcement check throwing fails closed with `500` (the branch the original flake was incidentally exercising).
4. **Defensively isolated `DATABASE_URL`** in `byot-integration.test.ts` and `managed.test.ts` so they stay safe if SSO/internal-DB hooks ever extend to those paths.

5/5 isolated runs of `middleware.test.ts` clean (35 pass), all CI gates green.

PR #1592, opened 2026-04-19
refactor(web/admin/actions): reason-on-deny + payload rendering + bucket-1 polish

msywulak:
First of three queue-page revamps for #1588 (admin console final-pass coverage). Validates the bucket-1 target shape (button-row filter, inline expand, RelativeTimestamp, reason-on-deny, single-banner errors) before primitives extract on the third adopter (per the #1551 rule).

## Why this one first

`/admin/actions` was already closest to ideal — defaults to `pending`, has bulk select with partial-failure preservation, relative timestamps with absolute tooltip, inline expand. Smallest delta to the target shape. Perfect to validate the pattern before porting it to `/admin/learned-patterns` (PR 2) and `/admin/approval` (PR 3).

### Compliance — reason-on-deny (P0)
- New \`deny-dialog.tsx\` (\`DenyActionDialog\`) handles both single and bulk deny
- Captures an optional reason; dialog shows audit-history hint when blank
- Cmd/Ctrl+Enter submits; textarea autofocused on open
- Persists actual reason text to \`action_log.error\` instead of the legacy hardcoded \"Denied by admin\" / \"Bulk denied by admin\"

### Clarity — type labels
- New \`labels.tsx\` (\`ACTION_TYPE_LABELS\`, \`actionTypeLabel\`, \`actionTypeIcon\`)
- Badges now read \"SQL Write\" / \"API Call\" / \"File Write\" / \"Shell\" instead of \`sql_write\` / \`api_call\` / \`file_write\` / \`shell\`

### Inline expand — payload by type
- New \`PayloadView\` component branches on \`action_type\`:
  - \`sql_write\`/\`sql\` with \`payload.sql\` → highlighted SQL
  - \`api_call\`/\`api\` with \`payload.method\`/\`url\` → method chip + URL + structured body
  - \`file_write\`/\`file\` with \`payload.path\` → path + content blocks
  - Unknown types → JSON fallback (so a new agent tool's payload is never silently hidden)

### Resilience — single error banner
- Funnels approve / deny / rollback / bulk errors into one \`mutationError\` state instead of stacking up to 4 \`ErrorBanner\`s (aligns with #1557)

### Cleanup
- Dropped chevron column — \`cursor-pointer\` row class is sufficient indication
- Removed cargo-cult \`useTransition\` wrapping a synchronous \`setState\`
- Added \`Tooltip\` on truncated Target / Summary cells
- Removed unused \`Database\` / \`Globe\` lucide imports (now in \`labels.tsx\`)

[...]

Issue #1591, opened 2026-04-19
refactor(types): unify ActionStatus and ActionDisplayStatus, drop mapStatus


msywulak:
Surfaced during the bucket-1 critique pass for #1588.

## Problem

Two parallel enums for the same concept:

- \`ActionStatus\` (server, persisted in \`action_log.status\`): includes \`pending\`, lacks \`pending_approval\`
- \`ActionDisplayStatus\` (client wire format): includes \`pending_approval\`, lacks \`pending\`

`packages/web/src/app/admin/actions/page.tsx` carries a thin adapter:

\`\`\`ts
function mapStatus(status: ActionLogEntry[\"status\"]): ActionDisplayStatus {
  return status === \"pending\" ? \"pending_approval\" : status;
}
\`\`\`

Every consumer of an action's status either calls this or accidentally renders the raw server enum. The drift is documented in `@useatlas/types/action.ts`:

\`\`\`ts
/**
 * Distinct from the server-internal \`ActionStatus\` which uses \"pending\"
 * instead of \"pending_approval\".
 */
\`\`\`

## Proposal

Pick one, drop the other. Easiest: rename `pending_approval` → `pending` in `ActionDisplayStatus`. The badge label can still read \"Pending Approval\" — the *enum* doesn't have to.

Steps:
1. Update `ActionDisplayStatus` union, `ALL_STATUSES`, `RESOLVED_STATUSES`, `ActionToolResultShape` discriminated union, and `ActionApprovalResponse` in `packages/types/src/action.ts`
2. Update `STATUS_CONFIG` in `packages/web/src/ui/components/actions/action-status-badge.tsx` (rename key, keep label)
3. Update tests — many in `packages/web/src/ui/__tests__/action-types.test.ts` and `action-approval-helpers.test.ts` reference `\"pending_approval\"` literally
4. Drop `mapStatus` from `packages/web/src/app/admin/actions/page.tsx`
5. Confirm the agent loop's tool-result emitter (in `packages/api/src/lib/tools/actions/`) emits the new value — currently it emits `pending_approval` per the discriminated union

## Why

- One source of truth for action lifecycle status
- Removes a footgun (rendering the wrong enum value)
- Bucket-1 pattern alignment — `/admin/approval` and `/admin/learned-patterns` use single-status-enum patterns

## Out of scope

- Server-side `ACTION_STATUSES` union shape — already the right one (\`pending\`)
- DB migrations — `action_log.status` already stores \`pending\`

[...]

Issue #1590, opened 2026-04-19
feature(api/actions): atomic bulk approve/deny endpoint


msywulak:
Surfaced during the bucket-1 critique pass for #1588.

## Problem

`/admin/actions` bulk approve / bulk deny currently fires N parallel HTTP requests via `Promise.allSettled`:

\`\`\`ts
const results = await Promise.allSettled(
  ids.map((id) =>
    fetch(\`/api/v1/actions/\${id}/\${endpoint}\`, ...)
  ),
);
\`\`\`

Approving 50 selections = 50 HTTP roundtrips, 50 separate transactions, no atomicity, partial-failure semantics by accident. Inconsistent with `/admin/learned-patterns` which has a real \`POST /api/v1/admin/learned-patterns/bulk\` endpoint.

## Proposal

Add \`POST /api/v1/actions/bulk\` (mirroring learned-patterns' shape):

\`\`\`json
// Request
{
  \"ids\": [\"uuid1\", \"uuid2\", ...],
  \"action\": \"approve\" | \"deny\",
  \"reason\": \"optional reason text\"
}

// Response
{
  \"approved\": [\"uuid1\", ...],
  \"failed\": [{ \"id\": \"uuid2\", \"reason\": \"already resolved\" }]
}
\`\`\`

Server-side: single transaction. UI keeps its existing partial-failure preservation (re-narrow selection to failed IDs) but reads the result from the response instead of inferring from \`Promise.allSettled\`.

## Why

- Atomicity: bulk action either applies coherently or surfaces a real per-item failure list
- Cuts N HTTP roundtrips to 1
- Auth/rate-limit checks happen once instead of N times
- Brings actions in line with the established bucket-1 bulk pattern

## Out of scope

- Schema changes for \`action_log\` (use the existing \`approve\`/\`deny\` codepaths internally)
- Bulk rollback (separate concern — different per-row preconditions)

## Related

- #1588 (admin console final-pass coverage tracker)
- Bucket-1 \`/revamp-queue\` template work (will adopt this endpoint when porting `/admin/learned-patterns` and `/admin/approval`)

## Labels
\`feature\`, \`area: api\`

Issue #1588, opened 2026-04-19
tracker: admin console final-pass coverage


msywulak:
Tracks the remaining admin pages that haven't had a design pass. Bucketed by what kind of treatment each page needs so the right tool (or no tool) gets reached for.

Parent/context: `/revamp` covered 16 pages with a shared CompactRow+Shell distill target. Remaining pages don't share that pattern — need different treatments.

---

## Bucket 1 — Queue / moderation

**Treatment**: `/critique` all 4 first. If a shared pattern emerges (filter chips + bulk-selection bar + inline detail), write `/revamp-queue` and batch. Otherwise handle per page.

- [ ] \`/admin/approval\` — approval queue (pending/approved/denied)
- [ ] \`/admin/learned-patterns\` — pattern moderation
- [ ] \`/admin/actions\` — plugin action log with approve/deny/rollback
- [ ] \`/admin/abuse\` — flagged workspaces

---

## Bucket 2 — Data tables

**Treatment**: `/critique` + targeted `/arrange`, `/polish`, `/clarify`. These pages are for operators — polish affordances, not restructure. Per-page, not batched.

- [x] \`/admin/users\` — PR #1665
- [x] \`/admin/sessions\` — PR #1628
- [x] \`/admin/organizations\` — PR #1668
- [x] \`/admin/audit\` — PR #1671
- [x] \`/admin/admin-actions\` — PR #1671
- [x] \`/admin/token-usage\` — PR #1671 (reviewed, no changes needed)
- [x] \`/admin/usage\` — PR #1671

**Bucket 2 complete.**

---

## Bucket 3 — Specialized / complex

**Treatment**: No blanket pass. Run `/critique` only when you've personally hit a rough edge or an operator complained. Stable surfaces.

- [ ] \`/admin/semantic\` — 3-panel editor
- [ ] \`/admin/schema-diff\` — diff viewer
- [ ] \`/admin/roles\` — 8-flag permission matrix (would need a new PermissionGrid primitive)
- [ ] \`/admin/prompts\` — nested collections → items (933 lines; considered for Tier 2, deferred)
- [ ] \`/admin/compliance\` — reports dashboard
- [ ] \`/admin/scheduled-tasks\` — task list + cron create dialog

---

## Bucket 4 — Small / simple

**Treatment**: One-shot polish. `/critique` → quick skills → ship.

- [x] \`/admin/cache\` — stats + controls (PR #1705)

---

## Bucket 5 — Platform subtree

[...]

PR #1594, opened 2026-04-19
refactor(web/admin/learned-patterns): bucket-1 pattern adoption + correctness fixes

msywulak:
Second of three queue-page revamps for #1588 (admin console final-pass coverage). Adopts the bucket-1 target shape validated in #1592 — without trying to cram everything into a single uniform layout. The Sheet stays because this page's detail is genuinely richer than \`/admin/actions\`'s inline-expand can carry.

### Real optimistic revert
\`updatePatternStatus\` and \`bulkUpdateStatus\` now snapshot \`patterns\` / \`detailPattern\` before the optimistic write and restore from snapshot on \`!result.ok\`. The previous code wrote a generic error and *left the optimistic state applied* — only a refresh would show the truth.

### Server error message surfacing
Replaced three \"Failed to update pattern\" / \"Failed to delete pattern\" / \"Failed to bulk update\" placeholder strings with \`result.error\` from the mutation hook. That message already carries the server's \`message\` + \`(Request ID: ...)\` via \`extractFetchError\`, fixing the same Sentry-correlation gap addressed in #1592.

### Initial fetch uses extractFetchError
Inline JSON parsing replaced with \`extractFetchError\` so failed list loads also surface \`requestId\`.

### Filter consolidation (was: two parallel \`<Tabs>\` strips with label-above-control)
- One button-row matching \`/admin/actions\`: \`All / Pending / Approved / Rejected | Any type / Query Patterns / Amendments\` separated by a border divider
- Entity filter remains a Select (cardinality is unbounded)
- Bulk actions appear inline in the same row when items are selected (was: bulk actions in page header next to title, which shifted the title)
- \`Clear\` button still appears on right when any filter active

### Stats compression (was: 4 StatCards eating the fold)
Single inline strip: \`4 pending · 1 approved · 1 rejected · 6 total\`. Removes ~120px of vertical chrome, surfaces the table sooner.

### Time
Sheet's \`Created\` / \`Last Updated\` / \`Reviewed\` fields now use the \`RelativeTimestamp\` pattern from \`/admin/actions\` (with absolute timestamp on tooltip). Table date column kept as absolute — patterns are reviewed in batches over days/weeks, real-time framing doesn't help triage here.

### Sheet kept intentionally
\`/admin/actions\` uses inline row expansion. \`/admin/learned-patterns\` keeps the Sheet because its detail content (full SQL, semantic_amendment diff with green/red/cyan colorization, metadata grid, source queries, review history) is too rich to inline. Per the user's call: \"we don't need to fit everything into the compactrow/shell mechanic if something else makes sense.\"

## Polish

- \"Reps\" column header → \"Seen\" (full label \"Times seen\" for sort menu)
- Page wrapped in \`TooltipProvider\` for the new tooltip timestamps

## Not in this PR (deferred)

- **Stats endpoint refactor** — page still fires 4 parallel \`?limit=1\` HTTP roundtrips for the inline stats. Server-side rollup belongs in a separate API change, not a UI revamp.
- **Status enum unification** — covered by #1591
- **Bulk endpoint atomicity** — already exists for this page (\`/api/v1/admin/learned-patterns/bulk\`), so no change needed here. \`/admin/actions\` parity is #1590.
- **Primitive extraction** (\`RelativeTimestamp\`, button-row filter component) — happens in PR 3 (\`/admin/approval\`) per the \"extract after third adopter\" rule (#1551).

## Verified end-to-end via Playwright + psql

- Stats render as inline strip with correct counts
- Single filter row holds all controls + bulk actions
- Approving a row from the Sheet persists the change (status flips, \`reviewed_by\` recorded)
- \"Times seen\" appears in Sheet metadata grid
- Created / Last Updated render as relative timestamps with absolute on tooltip

\`bun run lint\` clean, \`bun run type\` clean, 51 unit tests pass.

## Related

- #1588 (admin console final-pass coverage tracker)
- #1592 (PR 1 of 3, validated the patterns adopted here)
- #1551 (extract primitives after third adopter — applies on PR 3)
- #1591 (status enum unification — out of scope)

## Test plan

- [ ] \`bun run dev\`, sign in as admin, navigate to \`/admin/learned-patterns\`
- [ ] Verify stats render as a single inline strip (not 4 cards)
- [ ] Verify all filter controls (status, type, entity, clear) sit in one row
- [ ] Click a row → verify Sheet shows \"Times seen\" and relative timestamps for Created / Last Updated
- [ ] Approve from Sheet → verify status updates and reviewed_by persists
- [ ] Trigger an intentional failure (e.g., revoke admin role mid-mutation) → verify the error surfaces the actual server message + Request ID, and the optimistic state reverts
- [ ] Select multiple rows → verify bulk actions appear inline in filter row

### DIFF
--- a/packages/web/src/app/admin/learned-patterns/columns.tsx
+++ b/packages/web/src/app/admin/learned-patterns/columns.tsx
@@ -208,14 +208,14 @@ export function getLearnedPatternColumns(): ColumnDef<LearnedPattern>[] {
       id: "repetitionCount",
       accessorKey: "repetitionCount",
       header: ({ column }) => (
-        <DataTableColumnHeader column={column} label="Reps" />
+        <DataTableColumnHeader column={column} label="Seen" />
       ),
       cell: ({ row }) => (
         <span className="text-xs text-muted-foreground tabular-nums">
           {row.getValue<number>("repetitionCount")}
         </span>
       ),
-      meta: { label: "Repetitions", icon: Hash },
+      meta: { label: "Times seen", icon: Hash },
       size: 72,
     },
     {
--- a/packages/web/src/app/admin/learned-patterns/page.tsx
+++ b/packages/web/src/app/admin/learned-patterns/page.tsx
@@ -216,9 +216,9 @@ export default function LearnedPatternsPage() {
             all: allRes.status, pending: pendingRes.status, approved: approvedRes.status, rejected: rejectedRes.status,
           });
         }
-      } catch {
+      } catch (err) {
         // Stats are non-critical — don't block the page.
-        console.debug("Failed to fetch learned pattern stats");
+        console.debug("Failed to fetch learned pattern stats", err);
       }
     }
 
@@ -236,8 +236,8 @@ export default function LearnedPatternsPage() {
           if (p.sourceEntity) entities.add(p.sourceEntity);
         }
         if (!cancelled) setSourceEntities([...entities].toSorted());
-      } catch {
-        console.debug("Failed to fetch source entities");
+      } catch (err) {
+        console.debug("Failed to fetch source entities", err);
       }
     }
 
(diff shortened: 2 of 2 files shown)


msywulak:
Review-feedback fix commit pushed (6e539144).

## Addressed (all 7 review findings)

### Critical
- **Bulk endpoint partial-failure silently swallowed** → \`bulkUpdateStatus\` now inspects \`result.data.{notFound, errors}\` after a 200 and surfaces a per-reason summary via new \`bulkPartialSummary\` helper. Reverts only rows the server didn't actually update.
- **\`table.resetRowSelection()\` ran unconditionally** → moved to success-only path. Whole-failure and partial-failure both narrow selection to failed IDs so retry hits exactly the unfinished work.
- **Concurrent \`updatePatternStatus\` clobbered snapshots** → captures \`originalRow\` / \`originalDetail\` *inside* the functional setState callback, then reverts *only the affected row* on failure. Concurrent mutations on different rows no longer step on each other.

### High
- **\`bulkUpdateStatus\` didn't sync \`detailPattern\`** → snapshot + optimistic update + revert applied to \`detailPattern\` alongside \`patterns\`. Sheet stays consistent with the table through both optimistic and revert paths.
- **\`setError\` discarded structured FetchError** → minimal local mitigation: \`result.error\` already includes \`(Request ID: ...)\`, so the message survives. Cross-cutting fix filed as #1595 (\`useAdminMutation\` should return structured \`FetchError\` not flattened string — affects ~20 admin pages).

### Low
- **\`} catch {} \` blocks dropped error binding** → both stats/entities catches now bind \`err\` and pass to console.debug.
- **2 over-explanatory comments** → deleted (lines 253, 273). Kept the load-bearing \"Stats are non-critical\" comment.

## Filed as follow-ups (not in this PR)

- #1595 — \`useAdminMutation\` should return structured \`FetchError\` not flattened string
- (existing) #1593 — pure-function tests; will get an additional bullet for \`bulkPartialSummary\` and the optimistic snapshot/restore pattern when the third adopter (PR 3) extracts these helpers

## Verification

- Bulk approve on 2 rows persists, stats update, selection clears, no false error banner
- Lint clean, type clean, 51 unit tests pass

Diff: 96 insertions, 21 deletions.

PR #1628, opened 2026-04-19
polish(web/admin/sessions): confirm revoke + relative timestamps + useAdminFetch

msywulak:
## Summary

Bucket-2 baseline polish pass on `/admin/sessions` per tracker #1588. Three targeted fixes, no structural redesign:

- **Single-row Revoke now gated by `AlertDialog`** (mirrors the bulk flow). Previously, clicking Revoke in the row action bar signed a user out immediately with no confirmation — inconsistent with the bulk pattern and a real accidental-revoke risk in a high-stakes operator surface.
- **`Created` + `Last Active` columns use `RelativeTimestamp`** from `@/ui/components/admin/queue` (the bucket-1 extraction). Relative time reads as "still active" vs "stale" at a glance; the absolute datetime is preserved on hover for precision. Also rolls the page into the `TooltipProvider` the tooltip expects.
- **Fetch state migrates from bespoke `useState` + `useEffect` to `useAdminFetch`** — drops ~40 lines of manual cancellation + error handling, picks up auto-invalidation on successful revoke (no more `fetchKey`), and surfaces structured `FetchError` through the shared admin error shape. Page pagination reads `?page=` directly via `useQueryState` so the fetch deps can key on offset without a circular dependency on the `DataTable` instance.

No backend change. API already supports search over email + IP.

## Scope guard

Per tracker #1588, bucket-2 is polish-only — no `CompactRow`/progressive disclosure on data tables. Not introducing shared primitives until a third adopter emerges.

## Test plan

- [x] `bun run lint`
- [x] `bun run type`
- [x] `bun run test`
- [x] `bun x syncpack lint`
- [x] `SKIP_SYNCPACK=1 bash scripts/check-template-drift.sh`
- [ ] Browser spot-check — deferred; dev stack takes several minutes to boot and this session is remote-driven. Reviewer should verify:
  - `/admin/sessions` renders (smoke)
  - Revoke button on any row opens the new AlertDialog confirm
  - Confirming revokes the session; cancelling leaves the list intact
  - Bulk-revoke still works (select rows → "Revoke N selected")
  - Hovering a timestamp shows the absolute datetime tooltip
  - Search filters the list; clearing the search resets

Refs tracker #1588 bucket-2.

### DIFF
--- a/packages/web/src/app/admin/sessions/columns.tsx
+++ b/packages/web/src/app/admin/sessions/columns.tsx
@@ -46,7 +46,7 @@ function shortUA(ua: string | null): string {
 // ── Columns ───────────────────────────────────────────────────────
 
 export interface SessionActions {
-  onRevoke: (sessionId: string) => void;
+  onRevoke: (sessionId: string) => Promise<void>;
   isRevoking: (id: string) => boolean;
 }
 
--- a/packages/web/src/app/admin/sessions/columns.tsx
+++ b/packages/web/src/app/admin/sessions/columns.tsx
@@ -4,8 +4,19 @@ import type { ColumnDef } from "@tanstack/react-table";
 import { DataTableColumnHeader } from "@/components/data-table/data-table-column-header";
 import { Button } from "@/components/ui/button";
 import { Checkbox } from "@/components/ui/checkbox";
+import {
+  AlertDialog,
+  AlertDialogAction,
+  AlertDialogCancel,
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+  AlertDialogTrigger,
+} from "@/components/ui/alert-dialog";
+import { RelativeTimestamp } from "@/ui/components/admin/queue";
 import { Clock, User, Globe, Monitor, Trash2 } from "lucide-react";
-import { formatShortDateTime } from "@/lib/format";
 
 // ── Types ─────────────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


msywulak:
## 4-agent review outcome

Ran code-reviewer, silent-failure-hunter, comment-analyzer, pr-test-analyzer in parallel. Summary:

### Fixed in this PR (commit 31ef4fae)

1. **\`perPage\` URL drift** — \`useDataTable\` can write \`?perPage=\` but fetch hardcoded \`LIMIT\`. Now reads via \`useQueryState\`; \`offset\`, \`qs.limit\`, fetch deps, and \`initialState.pagination.pageSize\` all derive from it.
2. **Bulk-revoke silent partial failure** — switched to \`Promise.allSettled\` + \`bulkFailureSummary\`/\`failedIdsFrom\` from the shared queue primitives; failed rows stay selected for retry; dedicated \`bulkError\` banner.
3. **\`SessionActions.onRevoke\` type** — \`void\` → \`Promise<void>\` so callers correctly await the mutation.
4. **Misleading \`onRetry={refetch}\` on revoke banner** — dropped (retry was refetching the list, not the failed mutation).

Single-row path now routes through a non-throwing adapter that debug-logs + surfaces via the hook's \`revokeError\`; bulk path uses a throwing \`revokeSessionOrThrow\` for \`Promise.allSettled\` categorization.

### Filed as follow-ups

- #1629 — \`useAdminMutation\` concurrent error state stomping (hook-level architecture — affects every page using bulk fan-out)
- #1630 — Shared \`ADMIN_FETCH_QUERY_KEY\` constant between \`useAdminFetch\` + \`useAdminMutation\`
- #1631 — e2e smoke + pure-function tests for \`shortUA\` + \`SessionsListSchema\` round-trip
- #1632 — \`useAdminFetch\` schema-mismatch error copy is not actionable

### Clean

- Comment analyzer: both new comments justified (KEEP verdict).
- No critical findings; no type-design issues; no lint/type/test regressions.

Gates re-run: lint ✓ type ✓ test ✓ syncpack ✓ template drift ✓.

Issue #1718, opened 2026-04-20
1.2.3 — Security Sweep (tracking)


msywulak:
Tracking issue for the 1.2.3 security sweep milestone. Full audit with in-milestone remediation.

## Shape

Audit-and-fix. P0/P1/P2 findings fixed in-milestone; P3s spun out as a cleanup tail (1.2.4 or punt). Milestone doesn't close until the P2+ board is clean.

Findings captured in `.claude/research/security-audit-1-2-3.md` as we go.

## Phases

- [x] **1. Auth config audit** (#1720) — Better Auth session duration, cookie flags (secure/httpOnly/sameSite), CSRF token generation/validation, password policy, magic-link expiry, refresh token rotation, admin-role assignment path. Verify `isAuthenticated` / `requireAdmin` / `requirePlatformAdmin` middleware coverage on every route.
- [x] **2. Org-scoping audit** (#1721) — Walk every `/api/v1/*` route; confirm `orgId` filtering on reads and writes. Look for: cross-org read leaks, cross-org write leaks, `?orgId=` query-param trust, admin routes that miss the platform-vs-workspace distinction. Same pass for ContentModeRegistry consumers.
- [x] **3. SQL validation edges** (#1722) — Attack the 4-layer validator: AST parser bypasses, CTE table-whitelist escapes, UNION/subquery edges, schema-qualified names, case sensitivity. Fuzz-input property tests. Verify timeout + LIMIT apply under load.
- [x] **4. Audit log coverage** (#1723) — Every write route emits an audit event with actor/org/target/action. Verify: delete flows, role changes, plugin install, connection edits, SSO config, publish events. Retention + append-only integrity.
- [x] **5. Secrets + error surfaces** (#1724) — audit complete; 6 P1/P2 findings (F-41..F-47) filed #1815–#1820; 0 P0. Grep for connection strings / API keys in 500 responses, stack traces, client-side bundles. `.env` / config paths never log values. Plugin credential storage encrypted at rest with a documented rotation path.
[...]

PR #1776, opened 2026-04-23
fix(security): 1.2.3 phase 3 — close F-17/F-18/F-19 validator bypasses

### LINKED ISSUE #1772: security: P1 — MySQL /*!NNNNN ... */ version-gated executable comments bypass validator (F-17, 1.2.3 phase 3)
Phase 3 of the 1.2.3 security sweep (tracking #1718, audit #1722). Full details in `.claude/research/security-audit-1-2-3.md` under "Phase 3 — SQL validator audit + fuzz".

## Summary

MySQL's version-gated executable comment `/*!NNNNN ... */` bypasses Atlas's 4-layer SQL validator. When the MySQL (or MariaDB) server version is ≥ `NNNNN`, the contents are executed as live SQL — but Atlas's `stripSqlComments()` and `node-sql-parser` both treat the block as a plain comment, hiding the executable content from the regex guard, the AST-type check, and the table whitelist.

## Repro

```sql
-- Agent writes (validator sees a benign "SELECT 1"):
SELECT 1 /*!50000 UNION SELECT user, authentication_string FROM mysql.user */

-- MySQL 8 executes (80000 ≥ 50000):
SELECT 1 UNION SELECT user, authentication_string FROM mysql.user
```

Confirmed against `node-sql-parser` 5.4 and live MySQL 8 during the phase-3 audit. Known-bypass assertion pinned in `packages/api/src/lib/__tests__/sql-validator-fuzz.test.ts` under "fuzz: known bypasses — current behavior, follow-up fixes" (F-17).

## Impact

- Validator bypass — any table the DB user has SELECT on is reachable (including `mysql.user`, other schemas, and tables intentionally omitted from the semantic layer)
- No privilege beyond agent tool access required — a prompt-injected agent can exfiltrate via this path
- Runtime `SET SESSION TRANSACTION READ ONLY` does NOT catch this — it's a SELECT/UNION read, not a write

## Fix sketch

Preprocess MySQL-mode queries before `stripSqlComments` / AST parse / whitelist:
- **Option A (preferred):** strip the leading `/*!NNNNN` marker and trailing `*/` so subsequent layers see the executable content as live SQL.
- **Option B (simpler):** reject queries that contain `/*!` in MySQL mode.

Either option is scoped to `packages/api/src/lib/tools/sql.ts`. Acceptance:
- [ ] Fuzz-suite assertion at F-17 flips from `expect(r.valid).toBe(true)` to `expectInvalid(...)`.
- [ ] Positive regression — `SELECT 1 /*!50000 UNION SELECT id FROM companies */` on a mysql connection where `companies` is whitelisted and `mysql.user` is not: validator rejects.
- [ ] PG mode behavior unchanged (PG does not support executable comments).
- [ ] New cases added for `/*!00000 */`, `/*!99999 */`, nested `/*!50000 /*!80000 ... */ */`, and EOF without closing `*/`.

## References

- Phase 3 audit section: `.claude/research/security-audit-1-2-3.md`
- Known-bypass pin: `packages/api/src/lib/__tests__/sql-validator-fuzz.test.ts`
- Tracking: #1718
- Audit issue: #1722

### LINKED ISSUE #1773: security: P2 — PostgreSQL SELECT INTO new_table bypasses validator (F-18, 1.2.3 phase 3)
Phase 3 of the 1.2.3 security sweep (tracking #1718, audit #1722). Full details in `.claude/research/security-audit-1-2-3.md` under "Phase 3 — SQL validator audit + fuzz".

## Summary

PG's `SELECT ... INTO new_table FROM source` construct creates a new table (DDL-equivalent). Atlas's validator accepts it because `node-sql-parser` returns `{ type: "select" }` and `tableList()` surfaces only the source, not the target. The regex guard matches `INTO\s+OUTFILE` but not bare `INTO`. The query reaches the driver and is rejected at runtime only by `SET default_transaction_read_only = on`.

## Repro

```sql
-- Validator: PASS (classified as SELECT, tableList = [companies])
SELECT * INTO new_table FROM companies
```

Known-bypass pinned in `packages/api/src/lib/__tests__/sql-validator-fuzz.test.ts` under F-18.

## Impact

- Defense-in-depth only: runtime `default_transaction_read_only` blocks the write. Audit logs would record the attempt as a legitimate SELECT.
- Gap is structural — a plugin, config misapply, or future refactor that weakens the read-only session would leave this as an exploitable write path.

## Fix sketch

Check the AST's `into` field directly instead of relying on regex. After `parser.astify(...)`, inspect each statement; reject if `stmt.into != null`. Preferred over regex because `\bINTO\s+[A-Za-z_]\w*` has false-positive risk against legitimate column references named "into".

Acceptance:
- [ ] Fuzz-suite assertion at F-18 flips from `expect(r.valid).toBe(true)` to `expectInvalid(...)`.
- [ ] `SELECT * INTO new_table FROM companies` rejected.
- [ ] `SELECT * INTO TEMPORARY new_t FROM companies` rejected (currently fails AST parse, keep assertion either way).
- [ ] Positive regression: `SELECT id, name FROM companies` — unchanged, still passes.

## References

- Phase 3 audit section: `.claude/research/security-audit-1-2-3.md`
- Tracking: #1718
- Audit issue: #1722

### LINKED ISSUE #1774: security: P2 — MySQL INTO DUMPFILE not blocked by regex guard (F-19, 1.2.3 phase 3)
Phase 3 of the 1.2.3 security sweep (tracking #1718, audit #1722). Full details in `.claude/research/security-audit-1-2-3.md` under "Phase 3 — SQL validator audit + fuzz".

## Summary

Atlas's regex guard blocks `INTO OUTFILE` but not `INTO DUMPFILE`. Both write to the MySQL server's filesystem; `DUMPFILE` differs only in that it writes a single blob (common for exfiltrating BLOB column contents) while `OUTFILE` writes formatted rows. The two variants share the same attack vector and the same privilege requirement (`FILE`).

## Repro

```sql
-- Validator: PASS
SELECT * FROM companies INTO DUMPFILE '/tmp/x'

-- Validator: REJECT (INTO OUTFILE matched by regex)
SELECT * FROM companies INTO OUTFILE '/tmp/x'
```

Known-bypass pinned in `packages/api/src/lib/__tests__/sql-validator-fuzz.test.ts` under F-19.

## Impact

- If the MySQL user has `FILE` privilege (should not, but common in dev environments), arbitrary bytes written to disk. Combined with a world-readable data directory, trivial escalation.
- Runtime `SET SESSION TRANSACTION READ ONLY` does NOT block filesystem writes — it restricts table writes only.

## Fix sketch

Change `FORBIDDEN_PATTERNS` entry from `/\bINTO\s+OUTFILE\b/i` to `/\bINTO\s+(?:OUTFILE|DUMPFILE)\b/i`. Single-character class addition, no grammar implications.

Acceptance:
- [ ] Fuzz-suite assertion at F-19 flips from `expect(r.valid).toBe(true)` to `expectInvalid(..., "forbidden")`.
- [ ] `SELECT * FROM companies INTO DUMPFILE '/tmp/x'` rejected.
- [ ] `SELECT * FROM companies INTO OUTFILE '/tmp/x'` still rejected.
- [ ] Column/alias named `dumpfile` unaffected (only matches when preceded by `\bINTO\s+`).

## References

- Phase 3 audit section: `.claude/research/security-audit-1-2-3.md`
- Tracking: #1718
- Audit issue: #1722

msywulak:
## Summary

Close the three open SQL-validator bypasses surfaced in the 1.2.3 phase-3 audit (PR #1775) — F-17 (P1, #1772), F-18 (P2, #1773), F-19 (P2, #1774). All three touch the same validator (`packages/api/src/lib/tools/sql.ts`) and the same fuzz test, so they ship in one PR per milestone guidance.

### F-17 — MySQL version-gated executable comments (`/*!NNNNN ... */`)

- `node-sql-parser` treated `/*!50000 UNION SELECT user FROM mysql.user */` as an ordinary comment, hiding the payload from the AST parser and the table whitelist, while MySQL 8 / MariaDB 10 both execute the inner SQL when the server version ≥ NNNNN.
- **Fix (Option A):** new `unwrapMysqlExecutableComments()` preprocesses MySQL-mode queries before every validation layer. The wrapper peels with a regex that (a) alternates against string literals so `/*!` inside a quoted string is preserved, (b) accepts `\d{0,5}` digits so the MariaDB `/*!` form and boundary `/*!00000` / `/*!99999` forms all unwrap, and (c) loops until stable so stacked `/*!NNNNN /*!MMMMM ... */ */` wrappers peel in both levels.
- Unclosed forms (no `*/`) fall through to the existing regex mutation guard so `DROP`/`INSERT`/etc. inside still reject. PG mode is unaffected (PostgreSQL has no executable-comment syntax).

### F-18 — PG `SELECT * INTO new_table FROM src`

- `node-sql-parser` returns `{ type: "select" }` with `tableList` only surfacing the source. The regex guard matched `INTO OUTFILE` but not bare `INTO`, so the only thing blocking it was runtime `SET default_transaction_read_only = on`.
- **Fix:** AST-layer guard in `validateSQL` rejects when `stmt.into?.type === "into"` and `stmt.into.keyword !== "var"`. Plain `SELECT` uses `{ position: null }` (no `type`) and is unaffected; MySQL `SELECT ... INTO @var` has `keyword === "var"` and stays allowed as session-local variable assignment.

### F-19 — MySQL `INTO DUMPFILE`

- `FORBIDDEN_PATTERNS` enumerated only `INTO OUTFILE`. Same FILE-privilege attack class.
- **Fix:** pattern extended to `/\bINTO\s+(?:OUTFILE|DUMPFILE)\b/i`. A column named `dumpfile` with no leading `INTO` is unaffected (regression pin).

## Fuzz coverage

`packages/api/src/lib/__tests__/sql-validator-fuzz.test.ts` — 354 cases total, all green:

- **F-17.a–F-17.h** cover the variant matrix: bare `/*!50000`, boundary `/*!00000`, high `/*!99999`, no-digit MariaDB form, inside a CTE body, comma-splice column smuggle, nested `/*!NNNNN /*!MMMMM ... */ */`, unclosed `/*!NNNNN DROP` without `*/`. Plus a positive-regression (`/*!...*/` around a whitelisted UNION still accepted) and a string-literal protection case (`'/*!50000 ...'` inside a quote stays intact).
- **F-17 PG-mode regression** asserts the unwrap is MySQL-only — PG sees `/*!...*/` as a plain block comment and does not evaluate the body.
- **F-18** pin plus two regressions (plain SELECT still accepts, MySQL `SELECT ... INTO @my_var` still accepts).
- **F-19** pin plus `INTO OUTFILE` still-rejected + `SELECT dumpfile FROM companies` still-accepted regressions.
- Unused `expectCurrentBypass` helper removed now that all phase-3 pins have flipped to `expectInvalid` / `expectValid`.

## Audit doc + ROADMAP

- `.claude/research/security-audit-1-2-3.md` — F-17/F-18/F-19 entries gain "Fix shipped" paragraphs describing the applied option and regression coverage. Severity-summary table gains a Status column. Phase-3 header and closing summary note the follow-up branch.
- `ROADMAP.md` — Phase-3 checklist item annotated with the three fixes; checkbox stays `[ ]` until the phase-3 bundle (audit + fixes) is marked fully closed on the tracker sweep.

## Test plan

- [x] `bun test packages/api/src/lib/__tests__/sql-validator-fuzz.test.ts` — 354 pass
- [x] `bun test packages/api/src/lib/tools/__tests__/sql.test.ts` — 103 pass (no pre-existing regressions)
- [x] `bun run lint` — clean
- [x] `bun run type` — clean
- [x] `bun run test` — full suite green
- [x] `bun x syncpack lint` — no issues
- [x] `SKIP_SYNCPACK=1 bash scripts/check-template-drift.sh` — clean
- [ ] On merge: close #1772, #1773, #1774 (the `Closes` lines handle this)
- [ ] On merge: flip ROADMAP phase-3 checkbox and update `#1718` tracker if all phase-3 items are now green

Closes #1772
Closes #1773
Closes #1774


msywulak:
Follow-up commit (93fe8998) addresses findings from the 4-agent review pass. No follow-up issues — everything rolled in.

**Critical:**
- `security-audit-1-2-3.md` stale reference to the deleted `expectCurrentBypass` helper removed; replaced with the pre-fix pinning pattern for the next audit author (comment-analyzer).

**Silent-failure telemetry (silent-failure-hunter):**
- `validateSQL` now emits a `log.warn` when a MySQL query collapses to whitespace after `/*!NNNNN ... */` unwrap — attack-shaped probe whose payload lived entirely inside the wrapper. Still returns "Empty query" to the user; this is pure audit-trail coverage.

**New fuzz pins (pr-test-analyzer, rated 5-8):**
- F-17.i `/*! 50000` (whitespace before digits) — locks in `\d{0,5}` tolerance against future `\d{1,5}` regression.
- F-17.j `/*!500000` (six digits, over spec) — parse-layer rejection pin.
- F-17.k nested no-digits `/*! /*! ... */ */` — separate regex branch from F-17.g.
- F-17 PG-mode pin strengthened: payload now references `mysql.user` inside the comment body, so unwrap-in-PG would demonstrably reject. Stays valid → stronger property.
- F-18 multi-var `INTO @a, @b` + F-18 PG TEMP pins.
- F-19 tab/newline evasion (`INTO\tDUMPFILE`, `INTO\nDUMPFILE`) + backtick `` `DUMPFILE` `` parse-layer pins.

**Comment polish (comment-analyzer):**
- F-17 block docstring: "no equivalent syntax" → tied to the actual `dbType === "mysql"` gate (ClickHouse/DuckDB/plugins may differ; gate is the real contract).
- F-18 AST-guard comment: notes F-19 regex already catches OUTFILE/DUMPFILE upstream; this guard is the safety net.
- F-17.f test title typo ("smuggle a column smuggle" → "smuggle a column").

**Test isolation (pr-test-analyzer):**
- Inline `process.env.ATLAS_DATASOURCE_URL = MYSQL_URL` inside a PG-dialect describe moved into its own `describe(... MySQL dialect slice)` with `useDialect(MYSQL_URL)`. Safe today but fragile for future additions.

Fuzz suite: 354 → 372 pins (all green). Combined with existing `sql.test.ts`: 465 cases, 1140 `expect()` calls pass. All five CI gates (lint, type, test, syncpack, template drift) green locally.

PR #1827, opened 2026-04-24
fix: add admin_action_retention_config to schema.ts (Drizzle drift)

msywulak:
## Summary
- Migration `0035_admin_action_retention.sql` (from #1821, F-36 phase 1) added the `admin_action_retention_config` table to SQL but not to the Drizzle schema.
- `scripts/check-schema-drift.sh` fails in CI on `main` with "Table 'admin_action_retention_config' exists in migration SQL but not in schema.ts".
- Without the schema entry, a future `drizzle-kit generate` would emit a `DROP TABLE` migration.

Mirrors `auditRetentionConfig` at line 718 — same column shape + `idx_admin_action_retention_config_org` index.

Together with #1826 this unblocks the two failing checks on `main`.

## Test plan
- [x] `bash scripts/check-schema-drift.sh` → "Schema drift check passed — all 61 migration tables found in schema.ts"
- [ ] CI passes

### DIFF
--- a/packages/api/src/lib/db/schema.ts
+++ b/packages/api/src/lib/db/schema.ts
@@ -1385,3 +1385,27 @@ export const adminActionLog = pgTable(
     check("chk_admin_action_status", sql`status IN ('success', 'failure')`),
   ],
 );
+
+// ---------------------------------------------------------------------------
+// Admin action retention config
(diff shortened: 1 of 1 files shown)


msywulak:
Folded into #1826 as part of a consolidated CI-unblock PR.

PR #1830, opened 2026-04-24
fix(security): F-43 mask secrets in marketplace GET /available (closes #1817)

### LINKED ISSUE #1817: Marketplace GET /available returns raw installedConfig without masking secrets (F-43, phase 5)
Phase-5 finding from the 1.2.3 security sweep (tracking #1718, scope #1724). Full context in `.claude/research/security-audit-1-2-3.md` under **F-43**.

## Repro

`GET /api/v1/admin/plugins/marketplace/available` (`admin-marketplace.ts:682-719`) returns `installedConfig: inst?.config ?? null` for each catalog entry the workspace has installed.

`inst.config` is the raw JSONB blob from `workspace_plugins.config`. For a plugin whose catalog entry declares `secret: true` fields (stored in `plugin_catalog.config_schema`), the secret values land in the response body unchanged.

Compare `admin-plugins.ts` (the platform-plugin surface) which uses `MASKED_PLACEHOLDER` for the same field type on `GET /:id/schema`.

## Impact

Any workspace admin role (which includes every org owner) can read stored credentials by calling the marketplace endpoint. UX precedent from the platform surface says these values are write-only once saved — the marketplace surface contradicts that. A lower-role admin added to the workspace later inherits read access without having typed the credential.

Audit trail: `logAdminAction` for `plugin.config_update` already emits `keysChanged: string[]` only, not values (F-22 precedent). The disclosure path is only the GET endpoint.

## Fix sketch

Look up the catalog entry's `config_schema` in the same handler, walk the JSONB response, and replace any `secret: true` key's value with `MASKED_PLACEHOLDER`. The write path already tolerates the placeholder (`admin-plugins.ts:416-421`). Chokepoint: `admin-marketplace.ts:711` (`installedConfig: inst?.config`). ~20 lines.

## Severity

**P1** — live credential disclosure to any workspace admin; read/write asymmetry vs. the platform surface.

msywulak:
## Summary

Phase 5 of the 1.2.3 Security Sweep. `GET /api/v1/admin/plugins/marketplace/available` returned `installedConfig` as the raw JSONB from `workspace_plugins.config`, exposing every `secret: true` field in each catalog's `config_schema` to any workspace admin — the platform surface (`admin-plugins.ts` GET `/:id/schema`) already masked with the `MASKED_PLACEHOLDER` bullet string; the marketplace surface contradicted the write-once-then-opaque contract.

- **GET /available** — walks each entry's `plugin_catalog.config_schema`, replaces any `secret: true` value with `MASKED_PLACEHOLDER`. Null / empty-string values pass through so the UI can tell "set but hidden" from "never configured".
- **PUT /:id/config** — pre-fetches the stored config + catalog schema and swaps any re-submitted `MASKED_PLACEHOLDER` on a secret field back to its prior persisted value before writing. Without this, the first PUT from any UI rendering the masked GET would overwrite the live credential with the literal bullet string — converting a disclosure bug into a corruption bug.
- Extracted `MASKED_PLACEHOLDER` + `maskSecretFields` + `restoreMaskedSecrets` + `parseConfigSchema` into `packages/api/src/lib/plugins/secrets.ts`. Both admin surfaces now share one constant — drift between them would silently break the write-path restore guard. `admin-plugins.ts` refactored to import the shared constant, dropping its two local definitions.

Closes #1817.

## Test plan

- [x] `bun test src/lib/plugins/__tests__/secrets.test.ts` — 16 new unit tests pin the exact placeholder string, mask/restore semantics, edge cases (null config, empty-string secrets, rotated secrets, non-secret placeholder passthrough, missing-original drop).
- [x] `bun test src/api/__tests__/admin-marketplace.test.ts` — 3 new GET tests (masks secret field, leaves non-secret untouched, null installedConfig when not installed) + 3 new PUT tests (round-trip placeholder to original, rotate with new value, pre-fetch 404 path). 50/50 green. Existing PUT tests updated to set the pre-SELECT row alongside the UPDATE result.
- [x] `bun test src/api/__tests__/admin-plugins.test.ts` — 33/33 green after the shared-constant refactor.
- [x] `cd packages/api && bun run scripts/test-isolated.ts --affected` — 4/4 files pass in 750ms.
- [x] `bun run lint` — 0 errors.
- [x] `bun run type` — 0 errors.
- [x] `bun run test` — full suite, exit 0.
- [x] `bun x syncpack lint` — no issues.
- [x] `SKIP_SYNCPACK=1 bash scripts/check-template-drift.sh` — passed.
- [x] `bash scripts/check-railway-watch.sh` — all COPY sources covered.

## Acceptance (from prompt)

- [x] Every `secret: true` field in `installedConfig` returns as `MASKED_PLACEHOLDER` (exact string match vs `admin-plugins.ts`). ✅ Pinned in `secrets.test.ts` + marketplace masking test.
- [x] Non-secret fields pass through unchanged. ✅ `region` field passthrough asserted in the F-43 test.
- [x] Write-path round-trip: PUT with `MASKED_PLACEHOLDER` for a secret field → stored DB value unchanged. ✅ `round-trips MASKED_PLACEHOLDER` test captures the UPDATE params and asserts the `apiKey` param is `sk-live-12345`, not the placeholder.
- [x] Docs impact check — `apps/docs/content/docs/guides/plugin-marketplace.mdx` does not describe read-after-write masking behavior today; no doc changes needed. (The existing note at line 80 is about internal-DB absence, not secret masking.)
- [ ] ROADMAP Phase 5 line — will append "F-43 → PR #<this>" in a follow-up commit on this branch once the PR number is assigned.


msywulak:
Review pass 2 — addressed all review findings (commit b84bc311).

**Critical (fixed):**
- Fail-open on malformed `config_schema`: `parseConfigSchema` now returns a three-state discriminator (`absent | parsed | corrupt`). A catalog row with a JSONB envelope / wrong shape previously collapsed to `null` → `maskSecretFields` treated it as "no schema, pass through" → every secret leaked. Corrupt state now masks every non-empty string and restores every stored key on write.

**Important (fixed):**
- Strict `secret === true` in `parseConfigSchema` (type-design #5) — a stored `secret: "true"` (string) was silently non-secret.
- Pre-SELECT failure emits its own `plugin.config_update` status=failure audit with `priorLookupFailed: true` so the admin-triggered attempt never goes silent when the DB is down before UPDATE runs.
- Omit-to-preserve for secret fields on PUT (rolled in from the deferred follow-up): dirty-field saves that drop the masked secret key no longer wipe the credential. Explicit clears (`""` / `null`) still pass through as rotations.
- `log.warn` on corrupt schema in both GET /available and PUT /:id/config so operators see catalog drift in pino.

**Polish:**
- `maskSecretFields` JSDoc now documents the defensive `{}` return for non-object configs.
- admin-plugins.ts mask comment no longer names specific sibling surfaces (less rot risk).
- admin-marketplace.test.ts round-trip comment points at the shared `secrets.ts` module.

Diff now +401 / -70. Tests: 23 unit tests (all three schema states, strict-boolean guard, omit-to-preserve, explicit-clear, corrupt mask-all + restore-all) + 6 new marketplace route tests. 79/79 green in the combined `bun test`, full `bun run test` exit 0, all six CI gates pass locally.

PR #2147, opened 2026-05-07
fix(api,wizard,semantic): make demo + wizard workspaces queryable on MCP/executeSQL (#2142)

### LINKED ISSUE #2142: fix(onboarding): demo-data flow doesn't sync semantic entities to per-org DB
## Problem

When a new workspace picks the **demo data** option during onboarding, only the connection (database) gets created. The semantic-layer entities — `customers`, `orders`, `products`, etc. that the demo NovaMart schema needs — are NOT written to the per-org `entities` table in the internal DB.

Result: every subsequent `executeSQL` call against demo tables (whether via the chat agent or via MCP clients like Claude Desktop / Cursor) is silently rejected with `unknown_entity` because the SQL whitelist comes from the empty per-org cache.

The MCP `listEntities` tool is misleading — it reads from a separate disk-based source (`semantic/entities/*.yml` global dir) and returns 13 NovaMart entities. So users see `listEntities` ✅ but `executeSQL` ❌ on the same table names.

## Repro

1. Sign up for a fresh Atlas workspace (SaaS)
2. Pick "demo data" during onboarding (NovaMart)
3. Connect Claude Desktop / Cursor or hit the MCP edge directly
4. Try `tools/call executeSQL` with `SELECT id FROM customers LIMIT 5`
5. Get: `{"code":"unknown_entity","message":"Table \"customers\" is not in the allowed list. Check catalog.yml for available tables."}`

## Severity

**P0 for the demo experience.** Every demo workspace is broken on its primary use case. Found while load-testing — the load-test workspace also picked demo and exhibits the same.

## Expected fix

The demo-data onboarding path should call `bulkUpsertEntities()` (or hit the existing `POST /api/v1/admin/semantic/org/import` endpoint internally) with the bundled NovaMart entities, populating the per-org `entities` table. Same shape as what the wizard's `/save` does.

## Related

- The dashboard has no UI to trigger the import endpoint manually (separate gap, also being filed)
- The catalog.yml error message is misleading on SaaS — the per-org store has no catalog.yml. Separate issue worth fixing for clearer diagnostics.
- PR #2141 fixed a related bug (MCP edge wasn't preloading the per-org cache). After #2141, the cache loads correctly — but it loads empty for workspaces that hit this onboarding gap.

## Verification

Probe: `POST /mcp/{ws}/sse` `tools/call executeSQL` `SELECT id FROM customers LIMIT 5` → expect 200 with rows, not `unknown_entity` envelope.

msywulak:
Closes #2142.

## What was actually broken

The issue title said "demo-data flow doesn't sync semantic entities to per-org DB" but the trace turned up two independent root causes in the same blast radius. Fixing only one would leave the other half-broken.

### Root cause 1 — whitelist lookup mismatch

`/use-demo` (`onboarding.ts:792`) calls `importFromDisk(orgId, { connectionId: "__demo__" })`, which DOES persist demo entities into `semantic_entities` — keyed under `connection_id = "__demo__"`. But `executeSQL` defaults to `connectionId = "default"` when callers omit it (`sql.ts:994`), which is the common case for both the chat agent and MCP clients (Claude Desktop, Cursor, etc.).

The old `getOrgWhitelistedTables` in `whitelist.ts:486-491` strict-matched by key once any non-`"default"` connection existed:

```ts
const hasNonDefault = Array.from(byConnection.keys()).some((k) => k !== "default");
if (!hasNonDefault) {
  tables = byConnection.get("default") ?? new Set();
} else {
  tables = byConnection.get(connectionId) ?? new Set();  // empty for demo workspaces
}
```

For a demo workspace `byConnection = { "__demo__": Set<13 tables> }` → `hasNonDefault = true` → lookup for `"default"` → empty Set → every table rejected with `unknown_entity`.

**Fix**: refactor to a single-connection fallback. When the requested `connectionId` is `"default"` and the org has exactly one connection keyed under a non-`"default"` name, use that single connection's tables. The fallback is intentionally narrow: multi-source orgs still require explicit `connectionId`, so we don't create cross-routing risk.

### Root cause 2 — wizard `/save` never populated `semantic_entities`

While tracing root cause 1, I read `wizard.ts:670-805` and noticed the `/save` handler writes YAMLs to disk but doesn't call `bulkUpsertEntities`. Pre-#2141 this didn't bite because the SQL whitelist read from the file-based store. Post-#2141 the MCP edge consults `loadOrgWhitelist` which reads from `semantic_entities`. So **every wizard-onboarded workspace has the same MCP-side breakage as demo workspaces** — chat works, MCP rejects every table.

**Fix**: add a `bulkUpsertEntities` + `invalidateOrgWhitelist` call right before the existing `_resetWhitelists()`. Failure surfaces as 500 `db_persist_failed` (not silently logged) so the operator notices rather than discovering it via a customer report.

## Test plan

- [x] `bun run scripts/test-isolated.ts --affected` — 10 affected files, all green
- [x] `bun run type` — clean
- [x] `bun run lint` — clean
- [x] 4 new whitelist tests in `semantic-org.test.ts`:
  - Demo / wizard fallback fires for single-connection orgs with non-"default" name
  - Strict isolation preserved for non-"default" lookups
  - Multi-source orgs do NOT get the fallback
  - "default" entries take priority over the fallback path
- [x] 2 new wizard tests in `wizard.test.ts`:
  - `/save` calls `bulkUpsertEntities` with the correct args + invalidates the org whitelist
  - DB-write failure surfaces as 500 `db_persist_failed`
- [ ] Smoke against `mcp.useatlas.dev` after deploy: `tools/call executeSQL` on `SELECT id FROM customers LIMIT 5` against the load-test workspace returns rows, not `unknown_entity`

## Follow-ups

- Re-run `tool-call-mix.js` against the live workspace after this lands; the per-tool error-envelope rates in `mcp-performance.mdx` should drop to ~0%. Update the doc with a follow-up commit.
- The wizard's disk-write half is now somewhat vestigial for the SaaS deploy mode (the DB is the source of truth post-this-PR); consolidating that is its own design discussion, not in scope here.

### DIFF
--- a/apps/docs/openapi.json
+++ b/apps/docs/openapi.json
@@ -10606,6 +10606,31 @@
                       "items": {
                         "type": "string"
                       }
+                    },
+                    "warnings": {
+                      "type": "array",
+                      "items": {
+                        "type": "object",
+                        "properties": {
+                          "kind": {
+                            "type": "string",
+                            "enum": [
+                              "disk_sync_failed"
+                            ]
+                          },
+                          "tableName": {
+                            "type": "string"
+                          },
+                          "reason": {
+                            "type": "string"
+                          }
+                        },
+                        "required": [
+                          "kind",
+                          "tableName",
+                          "reason"
+                        ]
+                      }
                     }
                   },
                   "required": [
--- a/packages/api/src/api/__tests__/admin-wizard-save-audit.test.ts
+++ b/packages/api/src/api/__tests__/admin-wizard-save-audit.test.ts
@@ -130,6 +130,37 @@ mock.module("@atlas/api/lib/semantic/sync", () => ({
   importFromDisk: mock(async () => ({ imported: 0, skipped: 0, total: 0 })),
 }));
 
+// Mock the entity store: this test exercises the wizard /save audit
+// emission, not the DB write. Returning a full success count keeps
+// the route on the 201 path so the audit assertions can run.
+const mockBulkUpsertEntities: Mock<(
+  orgId: string,
+  entities: Array<{ entityType: string; name: string; yamlContent: string; connectionId?: string }>,
+) => Promise<number>> = mock(async (_orgId, entities) => entities.length);
+
+mock.module("@atlas/api/lib/semantic/entities", () => ({
+  DEMO_CONNECTION_ID: "__demo__",
+  SEMANTIC_ENTITY_STATUSES: ["published", "draft", "draft_delete", "archived"] as const,
+  bulkUpsertEntities: mockBulkUpsertEntities,
+  upsertEntity: async () => {},
+  upsertDraftEntity: async () => {},
+  upsertTombstone: async () => {},
+  deleteDraftEntity: async () => false,
+  listEntities: async () => [],
+  listEntitiesWithOverlay: async () => [],
+  getEntity: async () => null,
+  deleteEntity: async () => false,
+  countEntities: async () => 0,
+  createVersion: async () => "",
+  listVersions: async () => [],
+  getVersion: async () => null,
+  generateChangeSummary: async () => "",
+  applyTombstones: async () => 0,
+  promoteDraftEntities: async () => 0,
+  archiveSingleConnection: async () => ({ ok: true as const, archived: 0 }),
+  restoreSingleConnection: async () => ({ ok: true as const, restored: 0 }),
+}));
+
 // Profiler module — wizard.ts imports the full set (OBJECT_TYPES,
 // outputDirForDatasource, generate* helpers). Only a handful are used in
 // /save; the rest are stubbed to satisfy the import graph. Keep list of
--- a/packages/api/src/api/__tests__/wizard.test.ts
+++ b/packages/api/src/api/__tests__/wizard.test.ts
@@ -692,6 +692,7 @@ describe("POST /api/v1/wizard/save", () => {
     // failure so the operator notices instead of discovering it via a
     // customer report.
     mockBulkUpsertEntities.mockClear();
+    mockWriteFileSync.mockClear();
     mockBulkUpsertEntities.mockImplementationOnce(() => Promise.reject(new Error("internal DB unreachable")));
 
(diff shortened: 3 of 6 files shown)


msywulak:
**Review fix-up pushed (`f292621`).** All 4 review aspects ran (`pr-review-toolkit:review-pr`); this commit addresses every finding worth landing here:

**silent-failure-hunter**
- ✅ Reorder `/save`: DB upsert runs before disk writes. A `db_persist_failed` 500 no longer leaves orphan YAMLs on disk.
- ✅ Partial-success: `bulkUpsertEntities` returns a success count and swallows per-entity failures. The route now returns `500 db_partial_persist` (with `attempted` / `succeeded` fields) instead of 201ing a half-persisted state — that 201 was silently recreating the exact #2142 failure mode for unaffected rows.
- ✅ `syncEntityToDisk` failures now surface in a `warnings: [{ kind, tableName, reason }]` array on the 201 response instead of being `.catch(log.warn)`'d into the void.
- ✅ Added `log.debug` inside the whitelist single-connection fallback so future debuggers can tell direct hits from heuristic matches.

**comment-analyzer**
- ✅ Both new comment blocks tightened per CLAUDE.md ("one short line max"); dropped caller-naming prose that would rot when callers change.

**pr-test-analyzer**
- ✅ Mode-keyed cache: new test asserts the fallback fires identically for `published` and `developer` mode caches.
- ✅ Plugin-table merge: new test pins the contract that plugin-registered tables merge into the fallback path.
- ✅ New wizard tests: `db_partial_persist` 500, `disk_sync_failed` warnings surfaced, warnings field omitted on clean save, no-disk-writes-on-`db_persist_failed`.

**Deferred** (small, separate PR): broader scope work on `bulkUpsertEntities` to return per-entity failure metadata (rather than just a count), and a follow-up audit of other fire-and-forget `.catch(log.warn)` patterns in the codebase.

`bun run scripts/test-isolated.ts --affected` → 10/10 green. Lint + type clean.

Issue #2274, opened 2026-05-11
feat(api): #2174d — promote BYOT discovery cache to Postgres + periodic refresh


msywulak:
Sub-issue of #2174. Lifts the in-memory per-(org_id, provider) catalog cache (introduced in #2174a) into Postgres once 2-3 direct providers exist, so refresh is centralized across pods and survives restarts.

### Scope

1. **Schema** — `workspace_model_catalog` (org_id, provider, payload JSONB, fetched_at, refresh_status, last_error). Mirror in `schema.ts`. Index on (org_id, provider). Mode-system opt-out: this is an operational cache, not user-surfaced content — bypass via comment in the migration.
2. **Helpers** — `getCachedCatalog(orgId, provider)` reads + checks freshness against TTL (default 6h, configurable via `ATLAS_BYOT_CATALOG_TTL_SECONDS`). `refreshCatalog(orgId, provider)` upserts using the provider-specific fetcher.
3. **Scheduler hook** — register a daily refresh job in the scheduler service that walks active BYOT workspaces and refreshes each `(org_id, provider)` row whose `fetched_at` is older than TTL. Skips workspaces with `decrypt_failed` keys (those need user intervention).
4. **Endpoint integration** — `GET /admin/model-config/catalog?provider=…` reads from cache, triggers a fresh fetch only on miss or `?refresh=1`.

### Why this isn't in #2174a

In-memory works for the first slice — single-pod self-hosted is fine, and our SaaS deployment is small enough that the cache-miss penalty per pod (~1 fetch per provider per 6h) is acceptable. But once 2 direct providers exist, cross-pod refresh storms become real, and the daily-refresh story needs a durable home.

### Depends on

- #2174a (cache shape) — must land first since the in-memory implementation defines the payload contract this migration must mirror.

### Non-goals

- Redis-shaped cache (tracked in #2055 as part of the broader Redis backlog).

msywulak:
Shipped via #2280 (rolled-up squash-merge `c9bc4132`).

msywulak:
Reminder: this is an `architecture`-labeled issue closed via #2280. Consider adding an entry to `.claude/research/architecture-wins.md` capturing the L1→L2 cache-promotion + Postgres-backed catalog deepening if it represents a module-deepening refactor worth tracking.

PR #2278, opened 2026-05-11
feat(api,web,admin): #2273 — Bedrock BYOT + ListFoundationModels discovery

msywulak:
## Summary

Third direct-provider slice of #2174. Stacked on #2277 (OpenAI) → #2276 (Anthropic) — review/merge in order.

Bedrock is the most involved direct-provider because it diverges on cred shape (IAM access key + secret + optional STS session token, JSON-bundled into \`api_key_encrypted\`) and adds a region column that's part of the catalog identity.

- \`@useatlas/types\` 0.0.22 → 0.0.23 — adds \`"bedrock"\` to \`MODEL_CONFIG_PROVIDERS\`, \`BEDROCK_REGIONS\` tuple, \`BedrockRegion\` type, \`BedrockCredentialBundle\` shape, \`bedrockRegion\` fields. **Consumer refs intentionally stay pinned at 0.0.22 in this PR** — bump in a follow-up after the publish workflow tags \`types-v0.0.23\` (per the npm-publish playbook).
- Migration **0057** — adds bedrock to \`chk_model_provider\`, \`bedrock_region\` column, \`chk_model_provider_region\` CHECK (NOT NULL when provider='bedrock'; NULL otherwise). Mirrored in \`schema.ts\`.
- \`packages/api/src/lib/bedrock-catalog.ts\` — per-(orgId, region) cache; discovery via \`@aws-sdk/client-bedrock\` \`ListFoundationModels\`. Filters non-text-gen + LEGACY models. Tagged failure classes (AWS status + exception name) for unauthorized / rate-limited / unavailable.
- EE — \`setWorkspaceModelConfig\` stores the JSON cred bundle encrypted (existing helper round-trips strings), persists region, invalidates the bedrock catalog cache on upsert. \`getWorkspaceModelConfigRaw\` returns the region. \`testModelConfig\` adds a \`bedrock\` case that verifies the bundle via \`ListFoundationModels\` (no inference tokens burned). \`rowToConfig\` masks the accessKeyId tail — the secret half NEVER appears on the wire, not even masked.
- AI Layer — \`getModelFromWorkspaceConfig\` \`case "bedrock"\` parses the bundle and instantiates \`createAmazonBedrock\`.
- Route — extends the \`/catalog?provider=\` enum and the byot-adapter dispatch; bedrock takes a separate inline flow because the fetcher signature includes a region. Extracted a \`finalizeByotCatalog\` helper so the audit + envelope tail stays DRY across providers.
[...]

PR #2277, opened 2026-05-11
feat(api,web,admin): #2272 — OpenAI BYOT + /v1/models discovery picker

msywulak:
## Summary

Second direct-provider slice of #2174 BYOT. Stacked on #2276 (Anthropic) — review/merge that one first; this PR's diff base re-targets to `main` once it lands.

- `packages/api/src/lib/openai-catalog.ts` — fetches OpenAI \`/v1/models\` with the workspace's BYOT key. Filters to chat-capable IDs (drops embeddings / whisper / tts / dall-e / moderation / instruct / realtime / audio / image variants). Same per-orgId cache shape and tagged failure classes as anthropic-catalog; shares the \`ATLAS_BYOT_CATALOG_TTL_MS\` TTL knob.
- Route \`/catalog?provider=openai\` now flows through a small \`byotAdapter()\` dispatch table that holds the per-provider knobs (fetcher, error classes, friendly name). This cuts what would have been ~60 lines of duplicate Anthropic-vs-OpenAI handler logic, and gives Bedrock (#2273) a single-entry plug-in point. The existing \`?provider=anthropic\` flow now flows through the same dispatch — same envelopes, same audit shape.
- \`setWorkspaceModelConfig\` calls \`invalidateOpenAICatalog(orgId)\` on a successful openai upsert.
- \`/admin/model-config\` swaps the free-text input for the picker when the saved provider is openai with a healthy key.

Closes #2272.

## Test plan
- [x] \`bun test src/lib/__tests__/openai-catalog.test.ts\` — 12 unit tests covering filtering, every error class, per-orgId scoping, refresh-bypass, invalidation, inflight dedup, recommended set.
- [x] \`bun test src/api/__tests__/admin-model-config.test.ts\` — 30/30 (anthropic + openai + gateway + audit + decrypt paths).
- [x] \`bun run type\` + \`bun run lint\` — both clean.
- [x] \`bun run scripts/test-isolated.ts --affected\` — 57/57.
- [x] Schema + template drift checks pass.
- [x] OpenAPI spec regenerated.
- [ ] Manual: save an OpenAI key on a SaaS workspace, confirm the picker renders the filtered chat-capable list, "Refresh now" forces a fresh fetch.

PR #2276, opened 2026-05-11
feat(api,web,admin): #2271 — Anthropic BYOT + /v1/models discovery picker

msywulak:
## Summary

- First direct-provider slice of #2174 BYOT (after the gateway BYOT shipped in #2173). Establishes the per-workspace discovery pattern that the OpenAI / Bedrock / cache-promotion / unknown-model-handling slices (#2272–#2275) will reuse.
- `packages/api/src/lib/anthropic-catalog.ts` — fetches Anthropic `/v1/models` with the workspace's BYOT key, in-memory cache keyed by `orgId` with 6h TTL (`ATLAS_BYOT_CATALOG_TTL_MS`). Tagged failure classes for 401 / 429 (with `Retry-After`) / 5xx so the route maps cleanly to envelopes.
- `GET /api/v1/admin/model-config/catalog?provider=anthropic[&refresh=1]` — reads the workspace's stored Anthropic key via `getWorkspaceModelConfigRaw` and returns the catalog. Decrypt drift surfaces as **422** with a re-enter-key message; missing key → **400 `missing_byot_key`**. Default (no `?provider`) still returns the gateway catalog (backward-compat for the existing UI fetch).
- `setWorkspaceModelConfig` calls `invalidateAnthropicCatalog(orgId)` on upsert so a key rotation doesn't serve a stale catalog from before the rotation.
- New audit action `model_config.catalog_refresh` records the fetch outcome (`{ provider, modelCount, source, error? }`) — apiKey is never logged, matching the `model_config.test` threat model.
- `/admin/model-config` swaps the free-text input for `<GatewayModelPicker>` when the saved provider is anthropic with a healthy key. Adds a "Last refreshed N ago / Refresh now" footer.

Closes #2271. Parent #2174 stays open until #2272–#2275 land.

## Test plan
- [x] `bun test src/lib/__tests__/anthropic-catalog.test.ts` — 15 unit tests cover normalization, 401/403/429 (with and without Retry-After) / 5xx / network / malformed-data paths, per-orgId scoping, refresh bypasses cache, invalidation forces re-fetch, and the inflight-promise dedup.
- [x] `bun test src/api/__tests__/admin-model-config.test.ts` — 24/24 tests pass; new block covers the `?provider=anthropic` flow end-to-end (success / 400 missing key / 400 wrong provider / 422 decrypt / 401 / 429 with Retry-After / 503 / audit emissions / backward-compat default).
- [x] `bun run type` (root) — tsgo + web typecheck both green.
[...]

PR #2279, opened 2026-05-11
feat(api): #2274 — Postgres L2 cache for BYOT discovery catalogs

msywulak:
## Summary

Promotes the per-pod in-memory BYOT catalog caches (anthropic/openai/bedrock) to a DB-backed L2 layer. Each provider module now reads L1 (in-mem) → L2 (DB) → upstream, and writes back through both on fresh fetch. The L1 cache stays as the inflight-dedup substrate; L2 gives cross-pod consistency + survival across pod restarts.

Stacked on #2278 (Bedrock) → #2277 (OpenAI) → #2276 (Anthropic).

- Migration **0058** — \`workspace_model_catalog (org_id, provider, region, payload JSONB, fetched_at)\` with UNIQUE (org_id, provider, region). Region is empty string for non-region providers; bedrock writes the actual region. Provider whitelist matches \`workspace_model_config\` minus \`gateway\` (anonymous + globally cached). Schema.ts mirror added.
- \`packages/api/src/lib/byot-catalog-store.ts\` — \`loadFromDB\`, \`storeToDB\`, \`deleteFromDB\`, \`isFresh\` helpers. **All DB paths are best-effort** — a DB outage falls through to the in-memory + upstream path. L2 is a performance optimization, not correctness.
- Each per-provider catalog module gets:
  - L2 read between the L1 hit check and the upstream fetch.
  - L2 write inside the inflight-dedup block on fresh fetch (so concurrent callers don't duplicate the persist).
  - Invalidator flushes both layers; the L2 delete is fire-and-forget so Effect-scoped callers stay synchronous.
- The L2 cache is OPERATIONAL — it intentionally bypasses the content-mode system. Documented in the migration header.

**Scheduler-driven periodic refresh** is split out as a follow-up — this PR ships the durable layer + L1↔L2 wiring; the daily walk job lands separately. Adding a scheduler hook here would expand the review surface without changing the observable BYOT behavior.

Closes #2274.

## Test plan
- [x] \`bun test src/lib/__tests__/byot-catalog-store.test.ts\` — 15 unit tests (load hit/miss, region handling, malformed payload, best-effort error swallowing, freshness, delete).
- [x] Existing 37 admin-model-config integration tests still pass — L2 wiring is transparent to the route layer.
- [x] \`bun run type\` + \`bun run lint\` clean.
- [x] \`bun run scripts/test-isolated.ts --affected\` — 92/92 affected isolated tests pass.
- [x] Schema + template drift checks pass.
- [ ] Manual: with Postgres running, save anthropic key → refresh catalog → restart API → confirm catalog is served from \`workspace_model_catalog\` row without a fresh anthropic /v1/models call.

### DIFF
--- a/packages/api/src/lib/__tests__/byot-catalog-store.test.ts
+++ b/packages/api/src/lib/__tests__/byot-catalog-store.test.ts
@@ -1,0 +1,182 @@
+/**
+ * L2 BYOT catalog cache tests (#2274).
+ *
+ * Mocks `db/internal` to assert wire shape without standing up Postgres.
+ * Per CLAUDE.md, every named export is mirrored in the mock so the
+ * partial-mock SyntaxError trap doesn't fire elsewhere.
+ */
+
+import { afterEach, beforeEach, describe, expect, test, mock } from "bun:test";
+
+interface RecordedCall {
+  sql: string;
+  params: unknown[];
+}
+
+const recorded: RecordedCall[] = [];
+let nextRows: unknown[] = [];
+let nextError: Error | null = null;
+let hasInternalDBValue = true;
+
+const mockInternalQuery = mock(async (sql: string, params: unknown[]) => {
+  recorded.push({ sql, params });
+  if (nextError) throw nextError;
+  return nextRows;
+});
+
+mock.module("@atlas/api/lib/db/internal", () => ({
+  hasInternalDB: () => hasInternalDBValue,
+  internalQuery: mockInternalQuery,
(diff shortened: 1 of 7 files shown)


msywulak:
Rolled into #2280 — closing without merge. The stacked-PR chain hit a CI drift pattern that needed fixing on every branch (per-PR `openapi.json` divergence → matching MDX regen on each branch), and the practical cost outweighed the per-PR review trail benefit. The full diff + every review-pass commit is in #2280. Source branch stays intact for reference until cleanup.

PR #2348, opened 2026-05-12
fix(web): hide native OS scrollbars on shadcn CommandList + SidebarContent

msywulak:
## Summary

Mirrors the docs-only CSS on `ui.shadcn.com` (`.cn-command-list`, `.cn-sidebar-content`). The shadcn registry command/sidebar primitives ship with raw `overflow-y-auto`, which renders an OS gutter on Windows/Linux. The shadcn docs site adds these scrollbar-hiding rules to its own demo previews — we were missing them.

- Hides scrollbars on `[data-slot="command-list"]` → covers the command palette, every combobox built on the primitive (faceted filters, sort lists, view options, gateway-model-picker).
- Hides scrollbars on `[data-slot="sidebar-content"]` → covers `AdminSidebar`.

cmdk and Sidebar handle their own keyboard/scroll behavior, so the OS gutter only adds visual noise. macOS overlay bars are already invisible — no-op there.

`DropdownMenuContent` and `SelectContent` are also native-scrollable but intentionally left alone (matches upstream shadcn's choice; they rarely overflow and a visible bar is the only "more below" affordance).

## Test plan

- [ ] Open the command palette on Windows/Linux — no scrollbar gutter when items overflow
- [ ] Same palette on macOS — overlay bar still works on scroll
- [ ] Admin sidebar with many nav items — no gutter
- [ ] Data-table filter combobox (long list) — no gutter
- [ ] Verify dropdowns and selects still show their bar when overflowing (intentional)

### DIFF
--- a/packages/web/src/app/globals.css
+++ b/packages/web/src/app/globals.css
@@ -137,6 +137,22 @@
   --animate-typing-dot: typing-dot 1.2s ease-in-out infinite;
 }
 
+/* Hide native OS scrollbars on shadcn primitives that ship with raw
+ * `overflow-y-auto`. Mirrors the docs-only CSS on ui.shadcn.com
+ * (`.cn-command-list`, `.cn-sidebar-content`) — cmdk + sidebar handle their
+ * own keyboard/scroll behavior, so the OS gutter just adds visual noise on
+ * Windows/Linux where bars are always-visible. macOS overlay bars are
+ * already invisible; this rule is a no-op there. */
+[data-slot="command-list"],
+[data-slot="sidebar-content"] {
+  -ms-overflow-style: none;
+  scrollbar-width: none;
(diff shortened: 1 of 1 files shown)


msywulak:
Code-review caught a real one: `SidebarContent` is wrapped in Radix `ScrollArea` (customized in #1559), so the `[data-slot="sidebar-content"]` selector matched the Radix Root (which is `overflow-hidden`), not a scrolling element — the rule was dead code and the comment was factually wrong.

Pushed b15ca74b to fix: narrowed to `[data-slot="command-list"]` only and rewrote the comment to call out the sidebar exclusion so a future reader doesn't re-add the dead selector.

Silent-failure-hunter and pr-test-analyzer both came back clean before this.

PR #2361, opened 2026-05-12
refactor(api): rename encryptUrl/decryptUrl → encryptSecret/decryptSecret (#2285)

### LINKED ISSUE #2285: refactor(api): rename encryptUrl/decryptUrl to encryptSecret/decryptSecret
Deferred from the BYOT review (code-reviewer I6). \`encryptUrl\` / \`decryptUrl\` in \`packages/api/src/lib/db/internal.ts\` are AES-GCM helpers that round-trip any string — the \`Url\` suffix is a relic of their first caller (connection URLs in \`organization\` and \`internal_connection\`).

#1832 / F-42 already widened the usage to non-URL secrets (integration credential blobs, BYOT model-config keys). #2278 piles on bedrock IAM cred JSON. CLAUDE.md already documents an \`encryptSecret\` name for non-URL secrets, but the code still uses \`encryptUrl\`.

The naming mismatch is harmless today, but it's a footgun: if a future security review tightens the helper to actually validate URL shape (e.g. parse the input and assert protocol), every bedrock + integration row breaks silently.

### Proposal

1. Rename the helper: \`encryptUrl → encryptSecret\`, \`decryptUrl → decryptSecret\`.
2. Update every call site (about 30 across \`packages/api\` + \`ee/src\`).
3. Add a deprecation re-export for one minor version so external SDK consumers (if any) don't break:
   \`\`\`ts
   /** @deprecated Use \`encryptSecret\`. */
   export const encryptUrl = encryptSecret;
   \`\`\`
4. Add a short comment near the def: \"Handles arbitrary string secrets — URL, API key, JSON cred bundle. The historic \`Url\` name is preserved as a re-export.\"

### Out of scope

- Changing the encryption algorithm or key derivation (handled by F-47 keyset rotation).
- Splitting \`encryptSecretFields\` (JSONB selective-field encryption) — different helper.

### Acceptance

- [ ] \`encryptSecret\` / \`decryptSecret\` are the canonical names.
- [ ] \`encryptUrl\` / \`decryptUrl\` exist as deprecated re-exports for one minor version.
- [ ] CLAUDE.md's encryption rule references the new names.

msywulak:
Closes #2285. Milestone: 1.4.3.

## Summary
- `packages/api/src/lib/db/internal.ts` helpers renamed `encryptUrl` → `encryptSecret`, `decryptUrl` → `decryptSecret`. The `Url` suffix was a relic — the helpers already round-trip non-URL payloads (bedrock JSON cred bundles, BYOT API keys, OIDC client secrets).
- `encryptUrl` / `decryptUrl` kept as `@deprecated` re-exports for one minor version so external SDK consumers don't break.
- 165 internal call sites across `packages/api` + `ee/src` migrated to the new names (no call site left on the deprecated alias inside our own code).
- `scripts/rotate-encryption-key.ts` now aliases the internal pair as `encryptUrlSecret` / `decryptUrlSecret` locally — both `db/internal.ts` and `db/secret-encryption.ts` export `encryptSecret`/`decryptSecret` after this rename, and the rotation script legitimately consumes both pairs.
- Three partial-mock test files (`byot-catalog-store`, `lookups`, `list-entities-data-source`) updated so the mock objects declare each export distinctly (no duplicate keys).
- New focused test in `packages/api/src/lib/db/__tests__/internal.test.ts` verifies `encryptUrl === encryptSecret` (reference equality) and that the alias round-trips correctly.

## CLAUDE.md / docs
- "Encrypted at rest" rule tightened to disambiguate the two `encryptSecret`s (the URL-aware one in `db/internal.ts` vs the prefix-only one in `db/secret-encryption.ts`) and call out the deprecated aliases as legacy.
- `apps/docs/content/docs/platform-ops/encryption-key-rotation.mdx` references the new names where appropriate.
- One stray frontend comment (`model-provider-section.tsx`) and one type JSDoc (`packages/types/src/model-config.ts`) updated.

## Out of scope
- Encryption algorithm / key derivation (F-47 keyset rotation).
- `encryptSecretFields` / `decryptSecretFields` JSONB selective-field helpers — separate name pattern, untouched.
- SQL migrations (`0036`, `0038`, `0057`) keep their historical references — migrations are immutable history.

## Test plan
- [x] `bun run type` — clean (across all workspace packages).
- [x] `bun run lint` — clean.
- [x] `bun run deps:lint` (syncpack) — clean.
- [x] `bash scripts/check-schema-drift.sh` — passed (68 tables, 1 dropped excluded).
- [x] `bash scripts/check-template-drift.sh` — passed (558 files verified after regeneration; `create-atlas` templates regenerated to mirror the new names).
- [x] `bash scripts/check-railway-watch.sh` — passed.
- [x] `cd packages/api && bun run scripts/test-isolated.ts --affected` — 186 files passed, 0 failed.
- [x] `bun run test:others` — all non-api workspace suites green.
- [ ] CI green on this PR.


msywulak:
## Review pass — applied (commit 4f7ba05)

Ran all 5 `/pr-review-toolkit` specialists in parallel. Strong convergence on findings; applied everything that was actionable in this PR. One Important finding (branded-string types to disambiguate the two `encryptSecret` modules) was filed as follow-up #2370 — too large to bundle here without ballooning the diff past the rename's scope.

### Applied

**Critical (sed-induced comment defects)**
- `packages/api/src/__mocks__/api-test-mocks.ts:346-348` — comment was `encryptSecret / encryptSecret` (sed collapsed two distinct symbols). Restored the contrast.
- `packages/api/src/lib/db/__tests__/secret-encryption.test.ts:60-62` — Telegram-token rationale was self-contradictory (same fn appeared to have both bug and fix). Disambiguated with module path.

**Important — converged across multiple agents**
- `internal.ts::decryptSecret` error/log strings broadened (`"connection URL"` → `"stored secret"`, ~9 sites). Test assertions in `internal.test.ts` updated to match. Operators triaging an SSO or model-config decrypt failure no longer get pointed at the connections table.
- Shared mock factories (`api-test-mocks.ts` + `ee/src/__mocks__/internal.ts`) now declare `encryptUrl`/`decryptUrl` stubs as insurance during the deprecation window.
- CLAUDE.md "Encrypted at rest" rule rewritten — describes the helper split by what payloads currently use which module instead of the misleading "everything else" framing.
- Deprecated alias JSDoc names a removal milestone (1.5.0) and ties the trigger to the SDK consumer surface.
- `secret-encryption.ts` module header upgraded from description to **WARNING** with explicit picking guide.

**Type design**
- `rotate-encryption-key.ts::rotateValue` now uses `switch` with `satisfies never` exhaustiveness — adding a third `kind` is a compile error instead of silent fall-through.
- Same file: warning above the import-aliasing block to prevent a future "fix" that would collapse the two same-name helpers into a single import.

**Polish**
- New alias test in `internal.test.ts` simplified to single referential-identity assertion (Bun's import cache isn't actually reset by dynamic `await import()`, so it added complexity for no isolation).
- `internal.ts` `encryptSecret` JSDoc dropped redundant phrasing, added companion-module pointer.
- `encryption-keys.ts` cross-reference reworded.

### Filed as follow-up
- **#2370** — brand `URLSecret` / `OpaqueSecret` return types so the type system enforces the picking guide. Touches every persistence call site; deserves its own PR.

### Deliberately deferred
- Frontend comment in `model-provider-section.tsx:264` referencing a server function name (the comment was already there pre-PR; rename made it slightly more visible but it's not net-new debt).
- Removing the `encryptUrl` / `decryptUrl` re-exports — that's the milestone-1.5.0 trigger documented on the JSDoc.

### Local CI re-run
- `bun run type` — clean
- `bun run lint` — clean
- `bash scripts/check-template-drift.sh` — passed (templates regenerated)
- `bun run scripts/test-isolated.ts --affected` — 186/186 passed

Issue #2340, opened 2026-05-12
feat(semantic): group-scoped entities + atomic publish extension


msywulak:
## Parent PRD

#2336

## What to build

Migrate `semantic_entities` to be group-scoped. Add `connection_group_id` to the table; backfill from existing `connection_id` → its auto-created group (1:1 from #2339). Update the three partial unique indexes to key on `(org_id, entity_type, name, COALESCE(connection_group_id, '__default__'))` per status. Extend the `/api/v1/admin/publish` atomic transaction to promote drafts using the new key. Update `/admin/semantic` so multi-member groups show one entity row, not N. The legacy `connection_id` column stays for transitional dual-write — it goes away in #SLICE_9.

Reference the 0028 post-mortem in #2336's "Implementation Decisions → 0028 prevention." The new partial indexes must include `entity_type` in the key, and the real-Postgres migration smoke test must cover this change.

## Acceptance criteria

- [ ] Migration adds `semantic_entities.connection_group_id` and backfills from existing `connection_id`.
- [ ] Three partial unique indexes recreated with `connection_group_id` in the key, including `entity_type`.
- [ ] Real-Postgres migration smoke test extended (`migrate-pg.test.ts`); the test must catch the 0028-class missing-column-in-unique-index bug.
- [ ] `withGroupScope` from #2338 used by every read site.
- [ ] Atomic publish endpoint (`/api/v1/admin/publish`) promotes drafts using the group-scoped key in a single transaction.
- [ ] Phase 4 archive cascade extends to group-archive (archiving a single group member does NOT archive content; archiving the group does).
- [ ] Content-mode `tables.ts` registry updated; `readFilter` / `resolveStatusClause` key on `connection_group_id`.
- [ ] Draft counts in `/api/v1/mode` report once per group, not once per connection (pending-changes banner copy: "N draft changes" — no per-region multiplier).
- [ ] `/admin/semantic` collapses N→1 for multi-member groups; entity-list badge shows environment name, not connection name.
- [ ] SQL whitelist resolves entities through group ownership (entity queryable on connection C iff C's group owns it).
- [ ] `@useatlas/types` adds `connectionGroupId` field to `SemanticEntity` (additive; `connectionId` retained as deprecated).
- [ ] Browser e2e (`@llm`) covers multi-region admin happy path mirroring `published-vs-draft.test.ts`.

## Blocked by

- Blocked by #2339 (connection groups table must exist with backfill).

## User stories addressed

[...]

PR #2391, opened 2026-05-13
[codex] Scope semantic entities by connection group

msywulak:
## Summary

- Add `semantic_entities.connection_group_id` and migrate uniqueness from per-connection to per-connection-group scope.
- Preserve legacy `connection_id` compatibility while resolving reads, writes, publish overlays, diffing, and whitelist expansion through connection groups.
- Surface group/source context in admin semantic APIs and the semantic file tree.

## Validation

- `bun test packages/api/src/lib/db/__tests__/with-group-scope.test.ts packages/api/src/lib/semantic/__tests__/admin-source.test.ts packages/api/src/lib/semantic/__tests__/list-entities-data-source.test.ts packages/api/src/lib/semantic/__tests__/overlay-queries.test.ts`
- `bun test packages/api/src/lib/semantic/__tests__/overlay-queries-integration.test.ts packages/api/src/lib/db/__tests__/migrate-pg.test.ts packages/api/src/lib/content-mode/__tests__/registry.test.ts packages/api/src/lib/content-mode/adapters/__tests__/semantic-entities.test.ts packages/types/src/__tests__/migration.test.ts`
- `bun run type`
- `git diff --check`

Note: real Postgres migration smoke tests skipped locally because `TEST_DATABASE_URL` was unset.


msywulak:
Closing in favor of #2392.

We had two parallel implementations of #2340 (this one and #2392). After comparing them, #2392 was the stronger base — correct archive cascade via a `member_count` CTE (this PR's `AND connection_group_id IS NULL` would have silently disabled the cascade post-migration, since 0063 backfills the column for every legacy row); complete acceptance-criteria coverage (browser e2e, types version bump, docs site update, architecture-wins entry); and safer use of the with-group-scope helper (passes `column: "connection_group_id"` explicitly at every call site rather than flipping the helper's default, which would have been a footgun for the sibling slices in milestone 1.4.4 that don't yet have a `connection_group_id` column).

The three concrete wins this PR had over #2392 — the `entity_type` clause in the draft/published joins, the migration's dedup CTE for the pre-merged-connections case, and threading `connection_group_id` through `lib/residency/export.ts` — are now cherry-picked into #2392 in commit b9bfc073 with an explicit smoke test for the dedup behavior. Thanks for the parallel implementation; it caught real latent bugs in the merged version.

Issue #2347, opened 2026-05-12
chore(db): drop legacy connection_id columns


msywulak:
## Parent PRD

#2336

## What to build

Schema cleanup: drop the legacy `connection_id` columns from the six content tables (`semantic_entities`, `dashboard_cards`, `conversations`, `scheduled_tasks`, `approval_queue`, `pii_column_classifications`). Remove the transitional dual-write code. Drop the `connections.group_id` nullability — every connection must belong to a group at this point.

Note: `conversations.connection_id` is **retained** (it's the execution target, not deprecated content scope) — only the content-table columns are dropped.

The 1:1 backfill from #2339 guarantees every row has a `connection_group_id` set, so the column drops are safe. Real-Postgres migration smoke test must cover this.

## Acceptance criteria

- [ ] All transitional dual-write code paths removed; `connection_group_id` is the only source of truth in code.
- [ ] Migration drops `connection_id` from: `semantic_entities`, `dashboard_cards`, `scheduled_tasks`, `approval_queue`, `pii_column_classifications` (NOT `conversations`).
- [ ] Migration tightens `connections.group_id` to NOT NULL.
- [ ] Drizzle `schema.ts` mirrors the migration; `bash scripts/check-schema-drift.sh` passes.
- [ ] Real-Postgres migration smoke test (`migrate-pg.test.ts`) extended.
- [ ] `withGroupScope` helper from #2338 simplified — the COALESCE-with-sentinel branch for the legacy `connection_id` case can be removed.
- [ ] No test or code references the dropped columns.
- [ ] Architecture-wins entry updated to note the cleanup landed.

## Blocked by

- Blocked by #2346 (wire format must have already dropped legacy fields).

## User stories addressed

Closeout / cleanup — no direct PRD user story. Removes the technical debt incurred during the transitional period.

msywulak:
## Preflight before implementation

This is the final schema cleanup tail. It should remain blocked until #2346 completes and consumers no longer depend on transitional connectionId / connection_id content-scope fields.

Clarifications before coding:

- Do not drop conversations.connection_id; it remains the execution target.
- Only drop legacy content-scope columns after dual-write/read compatibility is removed everywhere.
- Tightening connections.group_id to NOT NULL should include a migration assertion/backfill repair path for any ungrouped connections created during the transition.
- If #2340/#2344 never introduce group archive/status, avoid cleanup language that assumes such a lifecycle exists.

[...]

Issue #2343, opened 2026-05-12
feat(scheduled-tasks): group-scoped scheduled tasks


msywulak:
## Parent PRD

#2336

## What to build

Migrate `scheduled_tasks` to be group-scoped. Tasks reference `connection_group_id`. At fire time, the scheduler engine picks the group's **primary member** to execute against. Tasks fire once per cron tick, not N times across members. Federation (run on every member, merge results) is out of scope.

**HITL — confirm before implementation:** PRD default is one-member execution. Alternative (rejected for v1): task fires on every group member, results aggregated downstream. Confirm the default is correct.

## Acceptance criteria

- [ ] Stakeholder confirms one-member execution (not federation).
- [ ] Migration adds `scheduled_tasks.connection_group_id`, backfills from the foundation slice.
- [ ] Scheduler engine picks group's primary member at fire time.
- [ ] Task form UI scopes by group; dropdown shows environments instead of raw connections.
- [ ] `withGroupScope` used by read sites.
- [ ] `@useatlas/types`: `ScheduledTask.connectionGroupId` added (additive); `connectionId` retained as deprecated.
- [ ] Real-Postgres migration smoke test extended.
- [ ] Unit tests cover: primary-member resolution, missing primary (fall back to first member), group-reassignment doesn't break in-flight tasks.

## Blocked by

- Blocked by #2339 (groups table + backfill).

## User stories addressed

- User story 6 (scheduled task fires once on chosen member)

msywulak:
## HITL + schema prerequisite before implementation

Please confirm scheduled-task execution semantics: a group-scoped task fires once per schedule against the group primary member, not once per member and not federated/aggregated.

This depends on the same primary-member decision as #2342. #2339 did not ship connection_groups.primary_connection_id, so this issue should either depend on #2342 adding it or a dedicated primary-member foundation issue.

Acceptance criteria should include behavior when the primary member is removed/ungrouped: choose a deterministic fallback or fail visibly with an actionable admin error.

msywulak:
Shipped in PR #2398 (commit 32dafb7b). Migration 0068 adds `scheduled_tasks.connection_group_id`, scheduler resolves the group's primary member at fire time (fall back to first by `(created_at, id)`), admin form scopes by environment, types add `connectionGroupId` while retaining `connectionId` for the #2346 deprecation window. 1.4.4 is now content-complete — only #2346/#2347 deprecation tail remains, gated on the stabilization window.

Issue #2341, opened 2026-05-12
feat(pii): group-scoped PII classifications


msywulak:
## Parent PRD

#2336

## What to build

Migrate `pii_column_classifications` to be group-scoped. Replace `connection_id` (NOT NULL DEFAULT 'default') with `connection_group_id`; backfill from #2339. The `UNIQUE(org_id, table_name, column_name, connection_id)` index becomes `UNIQUE(org_id, table_name, column_name, connection_group_id)`. PII scanning runs once per group; classifications are group-wide.

**⚠️ HITL — confirm before implementation:** PRD assumes same column = same PII across all group members (likely true for replicas of one schema; verify before locking in the unique-index change). If the assumption breaks, the unique key needs `connection_id` retained, and group-wide rollup becomes a different design.

## Acceptance criteria

- [ ] Stakeholder confirms "same column = same PII across group members" assumption (or redirects to per-member retention).
- [ ] Migration adds `pii_column_classifications.connection_group_id`, backfills, drops legacy default.
- [ ] `UNIQUE(org_id, table_name, column_name, connection_group_id)` index in place.
- [ ] `withGroupScope` used by every read site.
- [ ] PII scan tooling runs once per group; classifications surface group-wide in admin UI.
- [ ] Real-Postgres migration smoke test extended.
- [ ] `@useatlas/types`: PII classification shape adds `connectionGroupId` (additive).
- [ ] Unit tests cover the unique-constraint behavior under group reassignment (move connection between groups → classifications stay with the originating group).

## Blocked by

- Blocked by #2339.

## User stories addressed

- User story 4 (PII classifications group-wide; scan once)

msywulak:
## HITL required before implementation

Please confirm the PRD assumption: same table_name + column_name means the same PII classification across all members of a connection group.

If yes, this issue can proceed with UNIQUE(org_id, table_name, column_name, connection_group_id) and one scan/classification per environment.

If no, retain per-member classification identity and redesign this as a group rollup/view rather than a column-level uniqueness migration.

Also note: #2338 did not create a generic table-wide scope helper; use existing helpers only where they actually fit this table.

PR #2396, opened 2026-05-13
feat(dashboards): group-scoped cards (#2342)

### LINKED ISSUE #2342: feat(dashboards): group-scoped dashboard cards
## Parent PRD

#2336

## What to build

Migrate `dashboard_cards` to be group-scoped. Cards reference `connection_group_id` instead of `connection_id`. At view time, a card executes against the group's **primary member** (the first connection assigned to the group, or an explicit `primary_connection_id` set on the group). Multi-member side-by-side rendering is a separate follow-on; not in this slice.

**⚠️ HITL — confirm before implementation:** PRD default is primary-member execution. Alternative (rejected for v1): card shows N panels side-by-side, one per group member. Confirm the default is correct, or pick the side-by-side variant.

## Acceptance criteria

- [ ] Stakeholder confirms primary-member execution vs. side-by-side rendering.
- [ ] Migration adds `dashboard_cards.connection_group_id`, backfills from #2339.
- [ ] `connection_groups` table gets `primary_connection_id` column (if needed for the chosen UX).
- [ ] `withGroupScope` used by read sites.
- [ ] Card execution picks primary group member; card-create form scopes by group.
- [ ] `@useatlas/types`: `DashboardCard.connectionGroupId` added (additive); `connectionId` retained as deprecated.
- [ ] Real-Postgres migration smoke test extended.
- [ ] Browser e2e covers card-render against a group with three members.

## Blocked by

- Blocked by #2339.

## User stories addressed

- User story 5 (one dashboard card renders against chosen group member)

msywulak:
## Summary

Closes #2342. Third 1.4.4 multi-env slice: dashboard cards collapse onto the connection-group axis the same way semantic entities did in #2340.

- **Migration 0066** — adds `connection_groups.primary_connection_id` (composite FK to `connections (id, org_id)` so the primary stays org-isolated; `ON DELETE SET NULL` so removing a member silently demotes) + `dashboard_cards.connection_group_id` (additive, transitional). Backfills cards through 0062's 1:1 `g_<connId>` mapping.
- **Resolver** (`lib/dashboards-group-resolve.ts`) — pure `selectGroupMember(snapshot)` picks primary if set + still a member, else first by `(created_at, id)`. `NoGroupMembersError` surfaces zero-member groups as a typed 500 + `requestId` on interactive routes; scheduler tick treats it as a per-card skip.
- **Card create form** — `Add to Dashboard` dialog now carries an **Environment** picker driven by `/api/v1/admin/connection-groups`. Hidden for workspaces with ≤1 group. "Executes against `<primary>`" hint underneath; empty groups get an upfront warning.
- **@useatlas/types** — `DashboardCard.connectionGroupId` (additive); `connectionId` JSDoc-deprecated for read-side dual-write per the deprecation slice in #2347.

## Resolution rules (matches PRD default)

1. `primaryConnectionId` set AND still a member → primary.
2. `primaryConnectionId` null or stale → first member ordered by `(created_at ASC, id ASC)`.
3. Zero members → `NoGroupMembersError`. **Never** fall back to the workspace default.

## Why no FK on `dashboard_cards.connection_group_id`

`dashboard_cards` doesn't carry its own `org_id` — the parent dashboard provides scope. A composite FK to `connection_groups (id, org_id)` would either require denormalising `org_id` onto every card or relax to a single-column ref that silently leaks cross-org. Same trade-off 0063 took for `semantic_entities.connection_group_id`; org scope stays in the route handler, the real-Postgres smoke pins the column shape.

## Test plan

- [x] `bun run scripts/test-isolated.ts --affected` — 7/7 passing
- [x] Full `bun run test` — 373 api + 22 web + smaller suites, 0 failures
- [x] `bun run type` — clean
- [x] `bun run lint` — clean (pre-existing warning only)
- [x] `bash scripts/check-schema-drift.sh` — 69 tables, 1 dropped excluded
- [x] `bash scripts/check-template-drift.sh` — 563 files verified
- [x] `bun run deps:lint` (syncpack) — no issues
- [x] Unit: `selectGroupMember` 6/6 (primary set, primary null, primary stale, tie-break, empty, empty-with-primary)
- [x] `migrate-pg.test.ts` (real-PG smoke): column shapes + composite-FK cross-org rejection + `ON DELETE SET NULL` + backfill — extended with five 0066-specific cases
- [x] `migrate.test.ts`: pinned count bumped 64→65; `0066_dashboards_group_scoped.sql` added to the applied-list fixture in both `db/__tests__/migrate.test.ts` and `auth/__tests__/migrate.test.ts`
- [x] E2E `@llm` (Playwright) — `multi-env-admin.spec.ts` extended with three-member retarget (us-int → eu via primary PATCH, then null primary → us-int via `(created_at, id)` fallback) and an empty-group hint case
- [ ] Browser e2e against a live dev server (run by CI)

## Notes

- This branch owns migration 0066. 0064/0065 are reserved for the PII (#2341) and approvals (#2343) sibling slices; the idempotency contract on `__atlas_migrations` keeps the set consistent if they land out of order.
- Existing `connectionId` field on the wire stays for read-side dual-write until #2347.
- Drizzle's `onDelete` API doesn't expose the column-list `SET NULL (primary_connection_id)` form, so the FK is declared only in the migration. The smoke test pins the constraint shape so a future drift surfaces explicitly.

### DIFF
--- a/apps/docs/content/docs/guides/dashboards.mdx
+++ b/apps/docs/content/docs/guides/dashboards.mdx
@@ -41,6 +41,12 @@ Each dashboard card shows:
 - **Data table** — the underlying rows and columns
 - **Last refreshed** — timestamp of the most recent data
 
+### Environment (1.4.4+)
+
+If your workspace has more than one **connection group** (an "environment") configured under **Admin → Connections → Environments**, the **Add to Dashboard** dialog includes an **Environment** picker. Pick the group your card should run against — Atlas resolves it to the group's primary member at view time, falling back to the first member by creation order when no primary is pinned. Moving the primary to a different member (e.g. from `us-int` to `eu`) retargets every card scoped to that environment without editing the cards themselves.
+
+Single-environment workspaces don't see the picker — cards run against the workspace default, same as pre-1.4.4. A card pointing at an environment with zero members will refuse to refresh until an admin adds one, surfacing as a 500 with the offending group name; the dialog warns up-front so you don't save a card that can't run.
+
 ### Card Actions
 
 | Action | How |
--- a/apps/docs/openapi.json
+++ b/apps/docs/openapi.json
@@ -4887,6 +4887,12 @@
                       "null"
                     ]
                   },
+                  "connectionGroupId": {
+                    "type": [
+                      "string",
+                      "null"
+                    ]
+                  },
                   "layout": {
                     "type": [
                       "object",
@@ -49637,6 +49643,18 @@
                             "type": "integer",
                             "minimum": 0
                           },
+                          "primaryConnectionId": {
+                            "type": [
+                              "string",
+                              "null"
+                            ]
+                          },
+                          "resolvedConnectionId": {
+                            "type": [
+                              "string",
+                              "null"
+                            ]
+                          },
                           "createdAt": {
                             "type": "string"
                           },
@@ -49648,6 +49666,8 @@
                           "id",
                           "name",
                           "memberCount",
+                          "primaryConnectionId",
+                          "resolvedConnectionId",
                           "createdAt",
                           "updatedAt"
                         ]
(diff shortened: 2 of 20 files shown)


msywulak:
@codex review this PR

Issue #2407, opened 2026-05-15
1.4.4 closeout audit — findings tracker


msywulak:
## Context

10-agent post-merge audit of milestone 1.4.4 (multi-environment semantic layer, PRD #2336) surfaced 16+ findings across env picker UX, semantic keystone correctness, scheduler edge cases, test theater, and incomplete deprecation tail. Milestone was marked content-complete (8/11) on 2026-05-12; this tracker reopens the closeout.

## What was found

**Two prior "schema-only" worries were false alarms** — PII (#2341) and approvals (#2344) are fully wired in `ee/src/compliance/masking.ts` and `ee/src/governance/approval.ts`. The initial audit grepped only `packages/api/src/lib` and missed the `ee/` integrations. Lesson logged.

**What's actually broken** (organized by tier):

### Tier 1 — ship blockers
- env picker hidden by broken predicate on single/multi-singleton workspaces — the user's primary complaint
- admin Environments page leaks `g_*` synthetic names + no merge-into-group UX
- env-delete still 23503s on `url=''` tombstones after #2405/#2406
- 4 migration backfill tests are tautologies (0063, 0064, 0065, 0066)
- semantic entities collapse ACROSS groups; `getEntity` / `deleteEntity` are unscoped
- Phase 4 group-archive cascade not implemented (no archive route on `/admin/connection-groups`)

### Tier 2 — correctness gaps
- per-turn env override unverified through `executeSQL` streaming (the central correctness claim of #2345)
- `resolveGroupForConnection` broken for self-hosted null-orgId
- Codex scheduler crosses tenant→`__global__` org boundary
- Codex migration 0068 leaks synthetic names to tenant dropdowns
- Codex scheduled-task form + PATCH have two zero-group footguns
- dashboard card-create silently bypasses picker for single-group orgs
- `multi-env-admin.spec.ts` is route-mock theater + misuses `@llm` tag

### Tier 3 — planning hygiene
- `/admin/connections` cards must show group membership (PRD user story #20)
- `me-connection-groups` returns silent empty on no-active-org
- `ExportedSemanticEntity.connectionGroupId` should be optional in the wire type
- conversation/dashboard writes lack cross-org FK validation on `connection_group_id`

## Status

Sub-issues filed below. Milestone re-opens until Tier 1 + Tier 2 close. Tier 3 may bleed into 1.4.5 closeout if scope grows.

## Parent PRD

#2336

## Audit method

[...]

PR #2427, opened 2026-05-15
test(api): replace tautological backfill tests with honest insert-then-backfill (#2411)

### LINKED ISSUE #2411: test(api): four 1.4.4 migration backfill tests are tautologies (0063/0064/0065/0066)
## Tracker

#2407

## Problem

`migrate-pg.test.ts` "backfill" tests for migrations 0063 (semantic), 0064 (PII), 0065 (approvals), 0066 (dashboards) are tautologies: each test inserts a row with `connection_group_id` **already set to the target group**, then asserts the column equals that value. The migration's actual backfill UPDATE never runs.

Locations:
- `packages/api/src/lib/db/__tests__/migrate-pg.test.ts:495-506` (semantic 0063)
- same file `:840-851` (PII 0064)
- `:1085-1096` (approval 0065)
- `:1376-1386` (dashboards 0066)

Only **0067 conversations** (`:658-685`) does it honestly: inserts with `NULL`, re-runs the backfill SQL, asserts post-state.

## Why it matters

Four backfills are unverified end-to-end. A typo in any of the four backfill blocks would ship unnoticed. CI is green for the wrong reason. Given that the user has flagged 1.4.4 work as suspect, this is exactly the test-theater pattern they're worried about.

## Fix

Mirror the 0067 pattern for the four broken tests:
1. Insert the row with `connection_group_id IS NULL` (or whatever the pre-migration legacy shape was).
2. Execute the migration's actual backfill UPDATE block.
3. Assert post-migration state.

This is cheap — the test infra already exists; ~15 lines of test diff per migration.

## Files

- `packages/api/src/lib/db/__tests__/migrate-pg.test.ts:495-506,840-851,1085-1096,1376-1386`
- `packages/api/src/lib/db/__tests__/migrate-pg.test.ts:658-685` (the honest pattern to copy)

## Origin

Surfaced by pr-test-analyzer specialist agent during 1.4.4 audit.

msywulak:
## Summary
- The 0063/0064/0065/0066 backfill smoke tests inserted rows with `connection_group_id` already set to the target group, then asserted the column equaled that value — the migration's actual backfill `UPDATE` never executed. A typo in any of those four backfill blocks would have shipped unnoticed.
- Rewrite each test to mirror the honest pattern already used by the 0067 conversations backfill (`migrate-pg.test.ts:637`): seed connection + 1:1 group, insert a row in the pre-migration shape (`connection_id` set, `connection_group_id IS NULL`), run the migration's verbatim `UPDATE` block, then assert the post-state.
- Because 0069 drops `connection_id` from all four tables and the smoke suite runs every migration end-to-end first, each test re-adds the column for the duration of the assertion and drops it in a `finally` so the rest of the suite keeps the post-0069 production shape.

Closes #2411. Refs tracker #2407.

## Test plan
- [x] `bun run db:up && export TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5432/atlas_migrate_smoke && cd packages/api && bun test src/lib/db/__tests__/migrate-pg.test.ts` — **46 pass, 0 fail** against real Postgres (fresh per-test schema; legacy ALTER COLUMN no-ops on pre-migration columns)
- [x] `bun run test` — full workspace suite, all packages green
- [x] `bun run lint` / `bun run type` / `bun x syncpack lint`
- [x] Template drift / security-header drift / Railway-watch drift / schema drift / OAuth-helper drift — all pass
- [x] Confirm tests now would catch a regression: each test inserts with `connection_group_id = NULL` and the migration's verbatim `UPDATE` has predicate `WHERE connection_group_id IS NULL` plus the join. A typo that breaks the join skips the update, the column stays NULL, the `toBe(groupId)` assertion fails. No tautology remains.

### DIFF
--- a/packages/api/src/lib/db/__tests__/migrate-pg.test.ts
+++ b/packages/api/src/lib/db/__tests__/migrate-pg.test.ts
@@ -517,7 +517,16 @@ describeIfPg("migrate-pg (real Postgres)", () => {
       );
       expect(rows[0]?.connection_group_id).toBe(groupId);
     } finally {
-      await pool.query(`ALTER TABLE semantic_entities DROP COLUMN connection_id`);
+      // Cleanup must not shadow an in-`try` assertion error. If the DROP
+      // itself trips (pool closed mid-test, etc.), log and let the
+      // original failure propagate.
+      try {
+        await pool.query(`ALTER TABLE semantic_entities DROP COLUMN connection_id`);
+      } catch (err) {
+        console.warn(
+          `cleanup semantic_entities.connection_id DROP failed: ${err instanceof Error ? err.message : String(err)}`,
+        );
+      }
     }
   }, PG_TEST_TIMEOUT_MS);
 
@@ -879,7 +888,16 @@ describeIfPg("migrate-pg (real Postgres)", () => {
       );
       expect(rows[0]?.connection_group_id).toBe(groupId);
     } finally {
-      await pool.query(`ALTER TABLE pii_column_classifications DROP COLUMN connection_id`);
+      // Cleanup must not shadow an in-`try` assertion error. If the DROP
+      // itself trips (pool closed mid-test, etc.), log and let the
+      // original failure propagate.
+      try {
+        await pool.query(`ALTER TABLE pii_column_classifications DROP COLUMN connection_id`);
+      } catch (err) {
+        console.warn(
+          `cleanup pii_column_classifications.connection_id DROP failed: ${err instanceof Error ? err.message : String(err)}`,
+        );
+      }
     }
   }, PG_TEST_TIMEOUT_MS);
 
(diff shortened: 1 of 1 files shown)


msywulak:
PR review (5 specialist agents) summary:

**Critical issues:** none.

**Important issue (1):**
- silent-failure-hunter — `finally { DROP COLUMN }` could shadow the original assertion error if the cleanup itself throws (pool closed mid-test, etc.). Addressed in dd949d06: each cleanup DROP is now wrapped in its own try/catch with `console.warn`, so the original test failure propagates intact while the cleanup error stays visible. Per CLAUDE.md "never silently swallow errors" — the warn is required, not optional.

**Test-coverage gaps (rating 6–7):** filed as follow-up #2429 — out of scope for #2411's headline tautology fix, but the same typo-detection logic. Covers NULL-stays-NULL, `__global__` branch in 0063/0064/0066, multi-row backfill, and cross-org isolation.

**Comment-rot risk:** comment-analyzer flagged the "verbatim from 0063" claim as nothing-enforced. Considered adding `KEEP IN SYNC WITH 0063_*.sql:53-59` markers but line numbers themselves rot. Keeping current phrasing — the migration file path is already mentioned in each comment and that's the durable hook.

**Code-reviewer + simplifier:** ship-it. Verbatim SQL copy verified byte-for-byte against all four migration files. Duplication is load-bearing (a "helper" would defeat the test by hiding the verbatim claim).

Local re-verification after the silent-failure fix: migrate-pg.test.ts 46/46 pass against real Postgres.

Issue #2300, opened 2026-05-11
Proactive chat: one-time activation announcement + install-flow consent language


msywulak:
## Parent PRD

#2291

## What to build

When proactive mode is first enabled in a workspace, post a single message to an admin-selected announcement channel (default suggestion: `#general` if it exists). Idempotent — disabling and re-enabling doesn't re-announce.

Also adds disclosure language ('Atlas reads messages in channels you opt in to proactive mode') to the Slack/Teams install/OAuth flow so admins can present the consent story to their org cleanly.

`AnnouncementCoordinator` module handles the one-shot post.

See PRD §Disclosure and §Modules (7).

## Acceptance criteria

- [ ] `AnnouncementCoordinator` — `announceIfFirstTime(workspaceId, channelId)` with idempotency via a flag in `workspace_proactive_config`
- [ ] Admin console: 'Announcement channel' picker (autocomplete from chat-platform channel list, default `#general` if present)
- [ ] Slack/Teams install/OAuth flow updated with new disclosure copy
- [ ] Re-enabling proactive mode after a disable does NOT trigger a second announcement
- [ ] Unit tests for the coordinator (idempotency + channel selection)
- [ ] Demoable: fresh-enable proactive mode in a test workspace → single message in the chosen channel; disable + re-enable → no second message

## Blocked by

- Blocked by #2294

## User stories addressed

- User story 7
- User story 22

msywulak:
shipped via PR #2539

Issue #2294, opened 2026-05-11
Proactive chat: admin opt-in surface (workspace + per-channel config)


msywulak:
## Parent PRD

#2291

## What to build

Replace the env-var channel allowlist from #2292 with a full admin-console surface. Admin can:

- Enable proactive mode workspace-wide (master toggle, off by default)
- Add/remove channels to a deny-list
- Set per-channel interjection style (reaction-first / reply-first / DM-first / disabled)

Backed by two new tables (`workspace_proactive_config`, `channel_proactive_config`). Uses existing `useAdminFetch` / `useAdminMutation` hooks and the `FeatureName` registry for upsell/gating banner consistency.

See PRD §Schema changes and §Modules (9).

## Acceptance criteria

- [ ] Migrations for `workspace_proactive_config` + `channel_proactive_config` (mirrored to `schema.ts` in the same PR per repo policy)
- [ ] Admin API endpoints under `/api/v1/admin/proactive/...` for read + write config
- [ ] Admin console 'Proactive Chat' page: master toggle, deny-list editor, per-channel override table
- [ ] `MessageClassifier` + `InterjectionPolicy` (from #2292) now read config from DB, not the env var
- [ ] `FeatureName` registry entry added (`"Proactive Chat"`) for upsell/gating banner consistency
- [ ] Unit tests for the new DB-backed config reader
- [ ] Real-PG smoke for the new migrations via `migrate-pg.test.ts` harness
- [ ] Demoable: admin toggles a channel on/off in the console → Atlas behavior changes without restart

## Blocked by

- Blocked by #2292

## User stories addressed

- User story 1
- User story 2
- User story 3
- User story 4

msywulak:
Shipped via PR #2528 (commit `3247c90f`).

msywulak:
shipped via PR #2528

Issue #2291, opened 2026-05-11
PRD: Proactive chat-platform participation (Slack/Teams/Discord/etc. — Atlas answers without being summoned)


msywulak:
## Problem Statement

Atlas currently lives at the end of an explicit summon — users have to remember `@atlas` or `/atlas` to get an answer. That works for the people who already know Atlas exists and have an account, but it leaves Atlas invisible to the wider org. Data questions get asked constantly in Slack/Teams ("hey what was MRR last month?", "how many signups yesterday?") and most of them never reach Atlas — they get answered slowly by a human analyst, get answered wrong from memory, or never get answered at all. The net effect is that Atlas serves a small audience of explicit-summoners rather than acting like a teammate who pipes up when they can help.

## Solution

Add a proactive mode to the existing `@useatlas/chat` plugin that lets Atlas listen to channel messages in opted-in channels, detect questions it can confidently answer (against the workspace's semantic layer), and offer to help — without being mentioned or slash-commanded. Reactive plumbing (slash commands, mentions, threaded follow-ups, OAuth, state, error scrubbing) already ships in `@useatlas/chat`; this PRD covers only the additive proactive layer.

The proactive layer ships behind an `/ee`-gated flag, Slack-first, with Teams/Discord/Google Chat/Telegram following. It's positioned as a paid product tier — Atlas can answer questions on behalf of a whole workspace, not just the seated Atlas users, expanding TAM beyond the data team.

## User Stories

1. As a workspace admin, I want to enable proactive mode for my workspace from the Atlas admin console, so that I can opt in without editing config files.
2. As a workspace admin, I want a "master on/off" toggle for proactive mode, so that I can disable it across all channels in one click during an incident.
3. As a workspace admin, I want to maintain a deny-list of channels (e.g., `#random`, `#exec-only`) where Atlas should never proactively interject, so that I keep proactive mode off in noisy or sensitive places.
4. As a workspace admin, I want a per-channel interjection-style override (reaction-first / reply-first / DM-first / disabled), so that `#ask-data` can be aggressive while `#general` stays subtle.
5. As a workspace admin, I want a sensitivity preset (Cautious / Balanced / Eager), so that I can dial the confidence threshold without understanding the underlying classifier.
6. As a workspace admin, I want to curate a "public dataset" — a named subset of the semantic layer (e.g., MRR, headcount, top-line KPIs) — so that Atlas only answers non-linked askers from data I've explicitly approved for company-wide visibility.
[...]

PR #2539, opened 2026-05-17
feat(api,chat): #2300 — proactive-chat activation announcement + install consent

msywulak:
## Summary

Slice #2300 — when a workspace admin first flips `proactive.enabled = true`, Atlas posts a one-time announcement to the configured channel so end-users see the consent story at the point of contact. Disabling + re-enabling never produces a second announcement.

- `AnnouncementCoordinator` (`announceActivation`) with port/adapter shape (`ChatAnnouncer`) for test isolation; idempotent via new `workspace_proactive_config.announcement_posted_at` stamp.
- Admin PUT detects false→true `enabled` transition + announcement-channel set, fires the coordinator best-effort (failures don't fail the API).
- `ChatBridge.postChannelAnnouncement` adds channel-level posting via Chat SDK's `postChannelMessage`.
- Install-flow consent language added to the chat plugin docs and the Slack guide so admins can present "Atlas reads messages in channels you opt in to" before granting OAuth.

Foundation #2294 is merged on main; parent PRD is #2291.

## Test plan

- [x] `bun test src/lib/proactive/__tests__/announcement-coordinator.test.ts` (11 pass) — pure renderer + idempotency contract
- [x] `bun test src/api/__tests__/admin-proactive.test.ts` (26 pass) — route trigger + best-effort failure handling
- [x] `bun run scripts/test-isolated.ts --affected` (4 files pass)
- [x] `bun run lint` clean
- [x] `bun run type` clean
- [x] `bun x syncpack lint` clean
- [x] `bash scripts/check-schema-drift.sh` passes (mirror added for new column)
- [x] `bash scripts/check-template-drift.sh` passes
- [x] `bash scripts/check-railway-watch.sh` passes

Refs #2291.


msywulak:
Pushed openapi regen + migration count fix; rerunning CI.

Issue #2017, opened 2026-05-03
Invert core→ee dependency: replace dynamic @atlas/ee imports with Context.Tag services


msywulak:
## Problem

`CLAUDE.md` claims:

> **Self-hosted is always free** — Core AGPL functionality must never depend on `/ee`.

In practice, **23 non-test files in `packages/api/src/` import from `@atlas/ee`**, including the agent's hot path. The functional gate works (every import is dynamic + `isEnterpriseEnabled()`-guarded, so self-hosted users never load these modules), but the architectural direction is wrong: `core → ee` instead of `ee → core`.

### Hot-path evidence

| File | Import | What it does |
|---|---|---|
| `packages/api/src/lib/agent.ts:538` | `await import("@atlas/ee/platform/model-routing")` | Per-workspace model config |
| `packages/api/src/lib/db/connection.ts:1423-1425` | `await import("@atlas/ee/platform/residency")` | Region-based DB routing |
| `packages/api/src/lib/tools/sql.ts:774, 820` | `await import("@atlas/ee/sla/index")` | Query metrics recording |
| `packages/api/src/lib/tools/sql.ts:878, 1139` | `await import("@atlas/ee/compliance/masking")` | PII masking |
| `packages/api/src/lib/tools/sql.ts:1029, 1055` | `await import("@atlas/ee/governance/approval")` | Approval workflows |
| `packages/api/src/lib/auth/server.ts:26` | `import { isEnterpriseEnabled } from "@atlas/ee/index"` | **Static** import (worst offender) |

Plus ~17 other route/lib files. Full list:

```
gh -R AtlasDevHQ/atlas grep "@atlas/ee\|isEnterpriseEnabled\|requireEnterprise" packages/api/src/ --exclude="*.test.ts"
```

## Why this matters

1. **Architectural direction is inverted.** `core → ee` means the AGPL core mentions commercial-licensed module paths by string literal. EE should depend on core, never the reverse.
2. **Repo won't compile without `ee/`.** Even though dynamic imports don't execute at runtime when the flag is off, TypeScript still has to resolve the strings. You can't ship core as a standalone package today.
3. **Silent breakage risk.** Dynamic-import call sites aren't fully type-checked against the imported module's signature. If `@atlas/ee/sla/index` renames `recordQueryMetric`, core's call site breaks at runtime, not compile time.
4. **Docs don't match code.** `CLAUDE.md` says "core never depends on ee" — false. Either the docs change or the code does.

[...]

PR #2598, opened 2026-05-18
feat(api): #2593 — compile-time exhaustiveness gate for route domainErrors

### LINKED ISSUE #2593: feat(api): compile-time exhaustiveness gate for route domainErrors
## Background

EE-domain errors (`RetentionError`, `RoleError`, `ApprovalError`, `ResidencyError`, `BrandingError`, `DomainError`, `ComplianceError`, `ReportError`, `ModelConfigError`, `ModelConfigDecryptError`) are NOT in the `AtlasError` union or `ATLAS_ERROR_TAG_LIST` in `packages/api/src/lib/effect/errors.ts`. Instead they flow through a per-route `domainErrors: [...]` opt-in.

`domainError<TCode>()` already enforces compile-time exhaustiveness over the error's `code` field — adding a code without updating the statusMap is a `tsgo` error.

**But** there is no equivalent gate that says "if your route yields `RolesPolicy` you must also include `roleDomainError` in `domainErrors`." A new admin route that yields a Tag whose error isn't listed will surface as a generic 500 in production, not the expected 4xx envelope.

## Sketch

Option A: Tag the Effect's failure channel with a marker type that the route handler can introspect. Compile-time error if a known domain error is in `E` but missing from `domainErrors`.

Option B: A lint rule (eslint custom rule or a build-time script) that walks each `runEffect` / `runHandler` call site, infers the program's `E` channel, and checks coverage.

Option C: Just add every domain error class to `AtlasError` union + `ATLAS_ERROR_TAG_LIST`, with a default 500 mapping that's explicit ("uncategorized internal error"). Lose per-route status nuance but get one exhaustive place to look.

## Why this matters

The 1.5.1 closeout review (#2587) found this is a real risk: 10 promoted error classes, ~20 route files using them, easy to forget.

## Acceptance criteria

- [ ] Pick a design, prototype it
- [ ] Audit current routes for any missing `domainErrors:` wiring (manually first)
- [ ] Lock in via the chosen mechanism so future routes can't regress

## Dependencies

None.

msywulak:
Closes #2593.

## Summary

Approach **C** from the issue body — extend the `AtlasError` union and `ATLAS_ERROR_TAG_LIST` in `packages/api/src/lib/effect/errors.ts` with the 10 EE-domain error classes promoted to core during the #2017 slice trail:

- `RetentionError`, `RoleError`, `ApprovalError`, `ResidencyError`, `BrandingError`, `DomainError`, `ComplianceError`, `ReportError`, `ModelConfigError`, `ModelConfigDecryptError`

The exhaustive `mapTaggedError()` switch in `hono.ts` gains a matching case for each tag. Status assignments mirror the canonical per-route maps in `shared-{retention,residency,domains}.ts` and the inline maps in `admin-{roles,approval,branding,compliance,model-config}.ts` (e.g. `RoleError(builtin_protected) -> 403`, `ResidencyError(no_internal_db) -> 503`, `DomainError(railway_error) -> 502`, `ModelConfigDecryptError -> 422`).

## Why

Before this PR, a route that yields `RolesPolicy` / `ApprovalGate` / etc. without listing the corresponding `domainError(...)` in its `domainErrors: [...]` opt-in would surface the typed failure as a generic 500 in production — the issue body called this out as "10 promoted error classes, ~20 route files using them, easy to forget."

After this PR: per-route `domainErrors:` mappings still win on `classifyError` precedence (the `instanceof` check at step 3 runs before the tagged-error fallback at step 4), so existing routes keep their bespoke per-code status maps unchanged. **But** if a route forgets to wire `domainErrors:`, the typed failure now falls through to the tagged-error path and surfaces as a proper 4xx envelope instead of a 500.

## Audit findings

Manually walked every route under `packages/api/src/api/routes/` that yields one of the affected Tags. All currently wire `domainErrors:` correctly:

- `admin-router.ts` yields `RolesPolicy` via `runEnterprise` (different runner) with explicit try/catch fail-closed 503 — safe.
- `public-branding.ts` yields `Branding` but only calls `getWorkspaceBrandingPublic`, whose Effect channel is `never` — safe.
- `onboarding.ts` first handler yields `ResidencyResolver` for configuration reads (no error channel); the second handler (which calls `assignWorkspaceRegion`) already wires `residencyDomainError`.

The union-level mapping is the safety net for future drift, not a fix for an existing leak.

## Exhaustiveness mechanism

- **Outer switch**: TS's existing exhaustiveness over `error._tag` (the `mapTaggedError` switch) — adding a new tag to `AtlasError` without a case fails `tsgo`. Same pattern as before, now covering 10 more tags.
- **Inner switches**: each domain error's `switch (error.code)` ends with `return assertNever(error);`. Adding a new code variant to e.g. `RoleErrorCode` without a case fails `tsgo` until the case is added.
- **Tag list**: `ATLAS_ERROR_TAG_LIST` already uses `as const satisfies readonly AtlasErrorTag[]` plus `_AssertComplete` — both extended.

## Test plan

- [x] `bun x tsgo --noEmit -p packages/api/tsconfig.json` — clean.
- [x] `cd packages/api && bun run scripts/test-isolated.ts --affected` — 42 files, 0 failures.
- [x] `cd packages/api && bun run scripts/test-isolated.ts` — full api suite, 413 files, 0 failures.
- [x] 13 new `mapTaggedError` unit tests in `hono.test.ts` (75 tests total in the file, all pass).
- [x] `bun run lint` — only pre-existing warnings, no errors introduced.
- [x] `bash scripts/check-ee-imports.sh` — passes. The new imports are all from core (`@atlas/api/lib/...`), never `@atlas/ee`.
- [x] `bash scripts/check-schema-drift.sh` — passes (no schema changes).
- [x] `bash scripts/check-template-drift.sh` — passes (no template changes).

## Acceptance criteria (from #2593)

- [x] Pick a design, prototype it.
- [x] Audit current routes for any missing `domainErrors:` wiring (manually first).
- [x] Lock in via the chosen mechanism so future routes can't regress.

### DIFF
--- a/packages/api/src/lib/effect/__tests__/hono.test.ts
+++ b/packages/api/src/lib/effect/__tests__/hono.test.ts
@@ -45,6 +45,17 @@ const {
   SchedulerExecutionError,
   DeliveryError,
   UnsafeRegionMigrationResetError,
+  // ── EE-domain errors (safety net — see #2593) ─────────────────────
+  RetentionError,
+  RoleError,
+  ApprovalError,
+  ResidencyError,
+  BrandingError,
+  DomainError,
+  ComplianceError,
+  ReportError,
+  ModelConfigError,
+  ModelConfigDecryptError,
 } = await import("../errors");
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


msywulak:
Review came back clean — no blocking issues. Applied the comment-analyzer suggestion (#1): promoted the wire-code drift note to the block-level comment so future readers don't have to scan the inline `ApprovalError(expired)` case to find it.

The two other suggestions (`assertNever` payload-leak hardening, type-level exhaustiveness test) deferred — both noted as low priority by the reviewer; happy to spin as follow-ups if anyone wants them.

Issue #2624, opened 2026-05-19
ProactiveUserResolver should carry workspace context for multi-tenant linking


msywulak:
## Context

Surfaced during slice 2 of #2607 (PR for #2610 — wiring proactive listener callbacks in SaaS deploy).

The chat plugin's `ProactiveUserResolver` shape — defined at
`plugins/chat/src/proactive/answerer.ts:42` — is currently:

```ts
type ProactiveUserResolver = (asker: ProactiveAsker) => Promise<ResolvedAsker>;
type ProactiveAsker = { platform: string; externalUserId: string; userName?: string };
type ResolvedAsker = { atlasUserId?: string };
```

The asker carries no tenant context (Slack `team_id`, Teams `tenantId`, etc.). On SaaS multi-tenant the same Slack `externalUserId` can exist across multiple workspaces and resolve to different Atlas users; with the current shape the host can only do a global lookup, not a per-workspace lookup.

## Today's safe-fallback wiring

`deploy/api/atlas.config.ts` (slice 2 of #2607) stubs the resolver to always return `{ atlasUserId: undefined }` (every asker is treated as unlinked). The host-side `createProactiveAnswerAdapter` then takes the unlinked path, and because the adapter's `getPublicDataset` option is intentionally omitted, every unlinked-asker answer attempt is refused with `PROACTIVE_USER_SAFE_ERROR_MESSAGE`. The 🤖 reaction still fires; only the reaction-back / answer flow refuses.

Safe but limiting — linked-account answering (with the asker's full workspace RLS + toolset) won't light up until this is closed.

## Resolution options

1. **Plugin-side change** — thread the resolved `workspaceId` into `ProactiveAsker` so the resolver receives `{ platform, externalUserId, userName?, workspaceId }`. The listener already resolves `workspaceId` per event via `ResolveWorkspaceIdFn` (#2620), so the plumbing is local to `plugins/chat/src/proactive/listener.ts` + `answerer.ts`.
2. **Alternative shape** — keep `ProactiveAsker` minimal and pass workspaceId as a second argument: `(asker, ctx: { workspaceId }) => Promise<ResolvedAsker>`. Same plumbing, cleaner separation of concerns (the asker is the chat-platform identity; the workspace is per-event context).

Option 2 is the cleaner shape but option 1 is the smaller wire-diff. Both unblock linked-account answering on SaaS.

## Acceptance criteria

[...]

Issue #2623, opened 2026-05-19
Proactive listener type-design polish (P2 follow-ups from #2620)


msywulak:
## Context

Spun out of PR #2621 review (closes #2620). The multi-tenant refactor landed clean, but six type-design / safety items were deferred as P2 follow-ups. Bundling them here so they don't get lost.

### 1. Discriminated unions for coupled `ProactiveListenerConfig` callbacks

`ProactiveListenerConfig` has 25+ optional fields with several documented-only couplings:
- `getPublicDataset` + `executeQueryProactive` together unlock the public-dataset branch
- `userResolver` + `executeQueryProactive` together unlock the linked-asker path
- `isPaused` + `onPauseRequest` are paired in practice
- `feedbackCollector` is consulted only when `slashCommandName` is set

A host wiring half a feature group silently falls back to a stub. Discriminated unions would make the illegal combinations unrepresentable. Sketch:

```ts
type PublicDataset =
  | { mode: "off" }
  | { mode: "on"; getPublicDataset: GetPublicDatasetFn; executeQueryProactive: ProactiveExecuteQuery };
```

Breaking; only worth it if it lands alongside a 1.5.x bundle of changes.

### 2. `ResolverEventLite` type for synthetic-Message coupling

`listener.ts:884-895, 957-965` and `bridge.ts:984-992` synthesize Message shapes (`{ id: event.messageId, raw: event.raw } as unknown as Message`) when calling resolvers from contexts that don't carry a full Message (action / modal / slash). The resolver-side contract is "only reads `adapter.name` + `message.raw`", but it's convention-enforced. A typed `ResolverEventLite` that omits unread fields would lock the contract at compile time.

### 3. Slash-command resolver-throw degrades silently to question flow

`bridge.ts:994-1004`: when `resolveWorkspaceId` throws on a `/atlas feedback` invocation, the helper logs at warn and continues to the standard question flow — meaning `/atlas feedback the chart was wrong` runs as a SQL question. Either (a) post an ephemeral error when the subcommand is `feedback` and resolution failed, or (b) document the trade-off explicitly.

### 4. `enabled-gate` registration-retry path

[...]

Issue #2620, opened 2026-05-19
Refactor proactive listener for per-event workspaceId (multi-tenant prerequisite)


msywulak:
## Parent

- Prerequisite for #2610 (slice 2 wiring), #2611 (slice 3 @mention migration), #2612 (slice 4 dogfood)
- Surfaces under #2607

## Why

The chat plugin's proactive listener is **per-process, single-tenant by construction**:

- `bridge.ts:1583` calls `registerProactiveListener(chat, log, { workspaceId, ... })` exactly once
- `workspaceId` is a string baked at registration; stamped on every meter event (`listener.ts:309-310`), pause check (`:412-416`), quota call (`:460-464`)
- `chat.onNewMessage(/.+/, ...)` subscribes globally but all callbacks close over a single `workspaceId`
- `enabled-gate.ts:81-85` documents the constraint: *"Per-process today — if the EE roadmap adds runtime license activation…"*

SaaS routes Slack events from N tenants through the same Chat instance (slice 3 of #2607 migrates `@mention` + thread follow-ups off `slack.ts`). The moment slice 3 lands, every tenant's events arrive at the chat plugin but the listener stamps them all with one baked `workspaceId` — cross-tenant meter pollution, wrong workspace's quota, wrong pause-registry rows. Correctness/security bug.

`@chat-adapter/slack@4.23.0` **already** handles per-tenant bot tokens via `setInstallation`/`getInstallation` + request-context (`dist/index.d.ts:194`: *"Checks request context (multi-workspace) → default token (single-workspace) → throws"*). The Slack adapter side is multi-tenant ready. The remaining gap is the proactive listener.

### Plugin side (`plugins/chat/`)

**Replace single-`workspaceId` registration with per-event resolution.**

`plugins/chat/src/proactive/types.ts`:

```typescript
// Per-event workspace resolution. Returns null when the event doesn't
// belong to a known tenant (e.g., unrecognized team_id). Listener skips
// silently — no classify, no meter row.
export type ResolveWorkspaceIdFn = (event: {
  adapter: Adapter;
  thread: Thread;
  message: Message;
}) => Promise<string | null>;

// CHANGED: gate now takes workspaceId per call (was () => Promise<boolean>).
export type ProactiveGateFn = (workspaceId: string) => boolean | Promise<boolean>;

// NEW: replace static `workspace` config object with a fetcher.
export type GetWorkspaceConfigFn = (
  workspaceId: string,
[...]

Issue #2607, opened 2026-05-18
feat(api): wire @useatlas/chat plugin to enable proactive listener in production


msywulak:
## Context

The proactive admin UI at `/admin/proactive-chat` is currently a **no-op in production** even when `workspace_proactive_config.enabled = true`. The listener that classifies messages and reacts 🤖 is never instantiated.

Verified on prod (`satisfied-creation/us-int-postgres`) during 2026-05-18 dogfood:
- `workspace_proactive_config` — Atlas org has `enabled=t, sensitivity=balanced, classifier_mode=regex-prefilter` (toggled 21:50 UTC).
- `channel_proactive_config` — 0 rows.
- `proactive_meter_events` — 0 rows. Classifier has never run.
- `admin_action_log` — zero `proactive.classify` / `.react` / `.answer` events for the org.
- `slack_threads` — 0 rows.

## Root cause

The proactive logic lives in [`plugins/chat/src/proactive/`](https://github.com/AtlasDevHQ/atlas/tree/main/plugins/chat/src/proactive) (`@useatlas/chat`). The plugin's `bridge.ts:1578-1594` auto-registers the listener whenever `config.proactive` is supplied.

But `deploy/api/atlas.config.ts` has **no `plugins:` key**, so `@useatlas/chat` is never loaded in the SaaS deploy. The api server only has the bespoke `packages/api/src/api/routes/slack.ts` (slash command + thread follow-ups + `app_mention`), which doesn't route to the proactive listener.

## Decision: wire the plugin (path a)

Earlier drafts of this issue hedged between (a) wire the plugin or (b) port the classifier modules into core. After reading the code, (a) is the correct path for both immediate and long-term work:

- **Pure modules are small, listener is large.** `classifier.ts` (208) + `policy.ts` (146) + `pause.ts` (159) = ~500 lines of importable pure functions. But `listener.ts` is **1407 lines** and owns the reaction-post flow, ephemeral offer/answer cards, `PendingAnswers` cache, `runAnswerFlow` (lines 866-1097), feedback collection, public-refusal posts, unlinked-asker handling, kill-switch gating, and meter+audit emission. Porting (b) means rewriting ~900 lines of orchestration against `slack.ts`'s `postMessage`/`postEphemeral` helpers and then maintaining two copies.
[...]

PR #2625, opened 2026-05-19
feat(deploy): #2610 — wire proactive listener callbacks in SaaS deploy

### LINKED ISSUE #2610: feat(api): wire proactive listener callbacks (host bridge for @useatlas/chat)
## Parent

- #2607

## What to build

Add a `proactive:` config block under the plugin registration that supplies the callbacks `bridge.ts:1583-1594` expects. Each callback maps to existing host code — this slice is wiring, not new business logic:

- `isEnabled` — `isEnterpriseEnabled() && workspaceFlag` (read `workspace_proactive_config.enabled` for the requesting workspace)
- `classify` — pass the LLM through `AtlasAiModel` so the plugin's `classifier.ts` can call it
- `workspace` — read `workspace_proactive_config` (sensitivity, classifier_mode, monthly cap)
- `channelConfigs` — read `channel_proactive_config` (per-channel allow + overrides)
- `userResolver` — map Slack `externalUserId` → Atlas user via existing Slack installation tables
- `executeQueryProactive` — run the agent loop with the asker's identity bound (mirrors `executeAgentQuery` from `slack.ts` but tagged for the proactive path)
- `onMeterEvent` — write to `proactive_meter_events` via `lib/proactive/answer-meter.ts` and emit the matching `admin_action_log` row using `ADMIN_ACTIONS.proactive.{classify,react,answer}` from `lib/audit/actions.ts:668-670`
- `feedbackCollector` — wire `/atlas feedback <text>` outcomes
- `linkUrl` — the prod web URL for unlinked-asker prompts

Wire the three-layer kill switch: workspace toggle (read at every classify), `proactive_pauses` row via `lib/proactive/pause-registry.ts`, and DM `unsubscribe` (the plugin handles detection — host just provides `onPauseRequest`).

This is the load-bearing slice. After this lands, flipping `workspace_proactive_config.enabled=true` on a workspace with at least one `channel_proactive_config` row produces 🤖 reactions in real time. The admin UI at `/admin/proactive-chat` stops being a no-op.

## Acceptance criteria

- [ ] `deploy/api/atlas.config.ts` supplies all 9 `proactive.*` callbacks listed above
- [ ] With `workspace_proactive_config.enabled=true` and a row in `channel_proactive_config`, posting a question in the subscribed channel produces a 🤖 reaction within a few seconds
- [ ] `proactive_meter_events` accumulates `classify` rows under the workspace monthly cap
- [ ] `admin_action_log` shows `proactive.classify` / `.react` events with correct `org_id` + `target_id`
- [ ] Three-layer kill switch verified: workspace toggle off → no classify; `proactive_pauses` row → short-circuits before classify; DM `unsubscribe` → registers pause and short-circuits
- [ ] `isEnabled` short-circuits when enterprise is not loaded (no calls into the listener)
- [ ] Unit tests cover each callback's host-side behavior (meter write, audit emit, pause read)

## Blocked by

- #2608

msywulak:
## Summary

- Adds the `proactive:` block to the existing `chatPlugin({...})` call in `deploy/api/atlas.config.ts`. Slice 2 of #2607 — pure wiring, no new business logic.
- All required callbacks wired to host helpers under `packages/api/src/lib/proactive/`:
  - `resolveWorkspaceId` → Slack `team_id` → `slack_installations.org_id` lookup (post-#2620)
  - `isEnabled` → `createProactiveEnabledGate(getEnterpriseRuntime())` (two-tier: EE cached, workspace re-read)
  - `getWorkspaceConfig` / `getChannelConfigs` → fetchers from `workspace-config-loader.ts`
  - `classify` → `createProactiveClassifier(getProactiveAiRuntime())` (#2617)
  - `executeQueryProactive` → `createProactiveAnswerAdapter(getProactiveAiRuntime())` (#2618)
  - `isPaused` / `onPauseRequest` → `pause-registry.ts`
  - `onMeterEvent` → `recordMeterEvent` from `answer-meter.ts`
  - `getQuotaStatus` → `getWorkspaceQuotaStatus` (shape-adapted from `{ workspaceId }` → bare arg)
  - `getPublicDataset` → `getAllowlist` (shape-adapted similarly)
- Lazy `getProactiveAiRuntime()` (process-lifetime cached) defers `AtlasAiModelLive` materialisation until first call — keeps boot fast and lets the layer's 5s TTL handle admin model swaps.
- All imports are relative paths (`./packages/api/src/lib/...`) per the slice-1 container constraint (`atlas.config.ts` lives at `/app/` in SaaS, outside the workspace tree).

## Known gap (deferred — #2624)

- `userResolver` stubbed to always-unlinked (`{ atlasUserId: undefined }`)
- `getPublicDataset` adapter option intentionally omitted

Reason: the plugin's `ProactiveUserResolver` and `executeQueryProactive` shapes don't carry workspace context. On SaaS the same Slack `externalUserId` can exist across tenants. With the safe stubs in place, the 🤖 reaction + classify + meter / kill switch / quota all light up, but the reaction-back answer flow refuses safely until #2624 closes the plugin-side contract gap. Slice 4 dogfood will need #2624 landed (or worked around) to fully demonstrate the answer flow.

`feedbackCollector` intentionally deferred — non-trivial host helper.

## Test plan

- [x] CI gates green: lint / type / test / syncpack / template-drift / security-headers / railway-watch / schema-drift / oauth-helper-drift
- [ ] Boot Smoke green on Railway (proves the new config block parses + the lazy runtime construction doesn't trip boot)
- [ ] No regression on `executeQuery` stub (slice 1's friendly-upgrade throw still wired — slice 3 retires it)
- [ ] /pr-review-toolkit:review-pr addresses any P0/P1

## Follow-ups

- #2624 — `ProactiveUserResolver` + `executeQueryProactive` need workspace context (unblocks slice 4 dogfood)
- #2623 — Type-design polish (P2 items from #2620 review)

Closes #2610

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -289,6 +289,12 @@ ATLAS_ADMIN_EMAIL=admin@useatlas.dev
 # Default: http://localhost:3001 (matches `bun run dev` fixed port)
 # ATLAS_API_URL=http://localhost:3001
 
+# === Public Web URL (optional) ===
+# Public-facing web URL used in user-visible link prompts (e.g. the
+# proactive listener's "link your Atlas account" DM CTA). Defaults to
+# the SaaS host; set this on self-hosted deploys with a custom domain.
+# ATLAS_PUBLIC_WEB_URL=https://app.useatlas.dev
+
 # === Frontend API URL (optional) ===
 # When set, the frontend calls the API directly (cross-origin).
 # When unset, the frontend uses same-origin relative URLs
--- a/deploy/api/atlas.config.ts
+++ b/deploy/api/atlas.config.ts
@@ -12,6 +12,7 @@
  *   - ATLAS_REGION_APAC_DB_URL
  */
 
+import { ManagedRuntime } from "effect";
 import { defineConfig } from "./packages/api/src/lib/config";
 // Relative import: atlas.config.ts lives at /app/ in the SaaS container,
 // outside any workspace's node_modules resolution tree. The workspace
@@ -20,6 +21,51 @@ import { defineConfig } from "./packages/api/src/lib/config";
 // The `defineConfig` import above uses the same relative-path pattern
 // for the same reason. Resolved at boot via bun's TS loader.
 import { chatPlugin } from "./plugins/chat/src/index";
+import {
+  AtlasAiModel,
+  AtlasAiModelLive,
+} from "./packages/api/src/lib/effect/ai";
+import { getEnterpriseRuntime } from "./packages/api/src/lib/effect/enterprise-layer";
+import { createSlackWorkspaceIdResolver } from "./packages/api/src/lib/proactive/workspace-id-resolver";
+import { createProactiveEnabledGate } from "./packages/api/src/lib/proactive/enabled-gate";
+import {
+  getChannelProactiveConfigs,
+  getWorkspaceProactiveConfig,
+} from "./packages/api/src/lib/proactive/workspace-config-loader";
+import { createProactiveClassifier } from "./packages/api/src/lib/proactive/classifier-adapter";
+import { createProactiveAnswerAdapter } from "./packages/api/src/lib/proactive/answer-adapter";
+import { recordMeterEvent } from "./packages/api/src/lib/proactive/answer-meter";
+import {
+  handlePluginPauseRequest,
+  isPaused as isPausedHelper,
+} from "./packages/api/src/lib/proactive/pause-registry";
+import { getWorkspaceQuotaStatus } from "./packages/api/src/lib/proactive/quota";
+import { getAllowlist } from "./packages/api/src/lib/proactive/public-dataset";
+
+// ── Lazy ManagedRuntime for AtlasAiModel ────────────────────────────
+//
+// The classifier + answer adapters need a ManagedRuntime that provides
+// AtlasAiModel (the configured Vercel AI SDK LanguageModel). The
+// server's own runtime in `packages/api/src/api/server.ts` is built
+// locally and isn't exported, so we materialise a small dedicated
+// runtime here. Lazy construction defers the `AtlasAiModelLive`
+// resolution (which reads ATLAS_PROVIDER / ATLAS_MODEL via the settings
+// + providers modules) until first call — by then settings are
+// populated and the env is stable. Process-lifetime cached: the layer's
+// 5s TTL inside SaaS mode handles admin-driven model swaps without a
+// restart, so we don't need to rebuild the runtime ourselves.
+let _aiRuntime:
+  | ManagedRuntime.ManagedRuntime<AtlasAiModel, never>
+  | null = null;
+function getProactiveAiRuntime(): ManagedRuntime.ManagedRuntime<
+  AtlasAiModel,
+  never
+> {
+  if (_aiRuntime === null) {
+    _aiRuntime = ManagedRuntime.make(AtlasAiModelLive);
+  }
+  return _aiRuntime;
+}
 
 export default defineConfig({
   // ── Datasource ──────────────────────────────────────────────────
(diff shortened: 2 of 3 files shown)


msywulak:
## Addressed in `954c995` (PR-review fix-pass)

Five review aspects ran in parallel. Most agents reported well-structured wiring; main issues were comment accuracy + over-engineered laziness.

### Fixed

- **Lazy `getProactiveAiRuntime()` was a no-op** (simplifier P1): the wrapper deferred `ManagedRuntime.make` until first call, but `defineConfig({...})` evaluates eagerly, so the deferral never kicks in. `ManagedRuntime.make` is itself cheap — layer resolution happens lazily inside `runPromise`. Replaced with a single module-scope `const proactiveAiRuntime = ManagedRuntime.make(AtlasAiModelLive);` and rewrote the comment to describe what's actually happening (~17 LOC dropped).
- **`isPaused as isPausedHelper` rename was unnecessary** (simplifier P1): the local binding and the object-literal property key live in separate namespaces. Dropped the alias; use shorthand `isPaused,` in the proactive object.
- **`getPublicDataset` adapter-vs-listener ambiguity** (comment-analyzer P1 + errors P2): the "we omit `getPublicDataset`" comment on `executeQueryProactive` referred to the adapter-side option, but eight lines later we wire a separate plugin-level `getPublicDataset` on the listener. Rewrote both comments to disambiguate adapter-side vs listener-side and document the defense-in-depth purpose of the listener-side post-filter.
- **`userResolver` TODO wording** (comment-analyzer P1): said the function "doesn't carry team_id" but the function signature does take an asker. Reworded to anchor the gap in `ProactiveAsker`'s shape (`{ platform, externalUserId, userName? }` — no team_id) which is what a maintainer would actually inspect.
- **`ATLAS_PUBLIC_WEB_URL` undocumented** (code-reviewer P1): added an entry to `.env.example` under "Public Web URL (optional)" with one-line explanation of where it's consumed.

### Verified (no change needed)

- All other comment claims (5s TTL, structural runtime widening, structural-subset for `isPaused`, shape-adapter rationale, `recordMeterEvent` swallow-and-log) are accurate per the actual implementations.
- Type check still passes — Effect's `ManagedRuntime` variance accepts the wider `EnterpriseSubsystem` runtime where `ProactiveGate` is required, AND accepts `Layer<R, Error>` where the factory wants `<R, never>` (both verified by `bun run type`).

### Deferred (P2 / non-blocking)

- **AI runtime probe at boot** (errors P1): proposed adding an eager probe so a misconfigured AI provider is surfaced at boot rather than at first proactive event. Skipped — the existing classifier-adapter fail-closed + `log.warn` is the alert signal; layering a synthetic boot probe duplicates that. If ops finds the log spam insufficient for diagnosis later, add a dedicated check then.
- **Shared AppRuntime instead of separate AI runtime** (code-reviewer P2): the separate `proactiveAiRuntime` doesn't share `Telemetry` / `InternalDB` state with the server's main runtime. Today no Tag the proactive adapters need is duplicated, so no correctness bug. Worth a follow-up if we add a Tag that needs sharing.
- **Sentinel test for `EnterpriseSubsystem` → `ProactiveGate` structural assignability** (pr-test-analyzer P1): `bun run type` already covers this. Cheap to add but redundant with the existing gate.

### Verification

- Lint: 0 errors (10 pre-existing warnings unrelated)
- Type: clean
- Diff: +51 / -75 (net –24 LOC after fix-pass)

msywulak:
## Boot Smoke regression fixed in \`cc4db7ad\`

The first attempt failed Boot Smoke with:
\`\`\`
Failed to load config file "atlas.config.ts": ResolveMessage: Cannot find package 'effect' from '/app/atlas.config.ts'
\`\`\`

Same class of bug slice 1 (#2613) needed 4 Dockerfile fixes for: \`/app/atlas.config.ts\` sits OUTSIDE the workspace node_modules tree, so bare-package imports (\`import { ManagedRuntime } from "effect"\`) don't resolve. This wasn't caught by \`bun run type\` because the \`deploy/\` directory isn't in the workspace's tsgo project root.

Fix: moved runtime construction into \`packages/api/src/lib/effect/ai.ts:getProactiveAiRuntime()\` (mirrors the existing \`getEnterpriseRuntime()\` pattern). \`atlas.config.ts\` now imports the built runtime via relative path — no bare \`effect\` import. Used \`Layer.orDie(AtlasAiModelLive)\` to convert the Layer's \`Error\` channel to a defect — matches the \`<R, never>\` shape the proactive factories require (this is the same effective behavior the previous version had, but now type-checked properly because the new code lives in a checked file).

Side benefit: the new test-isolation runs caught a type error on the runtime annotation that was previously hidden by `deploy/` being excluded from `tsgo`. Type now flows correctly through the boundary.

PR #2629, opened 2026-05-19
refactor(plugins/chat): drop legacy channelAllowlist + ATLAS_PROACTIVE_CHANNELS

msywulak:
## Summary

Follow-up to #2628 (channelAllowed hotfix). #2620 made `getChannelConfigs` required on every consumer of the proactive listener, so the legacy `channelAllowlist: string[]` config field + `ATLAS_PROACTIVE_CHANNELS` env var resolution is dead code. Pre-customers — clean delete, no shims.

## What's removed

- `resolveChannelAllowlist()` export from `listener.ts`
- `channelAllowlist?: string[]` field from `ProactiveListenerConfig` (`listener.ts`) and `ProactiveConfig` (`config.ts`)
- `channelAllowlist` from the Zod schema (`config.ts`)
- Registration-time `const allowlist = resolveChannelAllowlist(config.channelAllowlist)` + `allowlistSize` log metric
- Two `channelAllowlist: ...` pass-throughs in `bridge.ts`
- Re-exports from `proactive/index.ts` and the top-level `index.ts`
- 3 tests from the deleted `resolveChannelAllowlist` describe block + 1 redundant "reacts when allow=true" test (subsumed by the migrated baseline)

`channel_proactive_config` rows (read per-event via `getChannelConfigs`) is now the SOLE source of truth for which channels are opted in. No row = not opted in.

## What changes for consumers

- `chatPlugin({ proactive: { ..., channelAllowlist: [...] } })` — no longer accepted (TS error + Zod fail-fast)
- `ATLAS_PROACTIVE_CHANNELS` env var — ignored. The chat plugin no longer reads it.
- SaaS already uses the DB rows; self-hosted dev setups that relied on the env var need to wire `getChannelConfigs` (which is required anyway post-#2620, so this is structurally already the case).

## Test plan

- [x] `bun run lint` — 0 errors
- [x] `bun run type` — clean
- [x] `cd plugins/chat && bun run test` — 438/438 pass
- [x] `grep -rn "channelAllowlist\|ATLAS_PROACTIVE_CHANNELS\|resolveChannelAllowlist" plugins/chat/src/` returns ZERO hits
- [ ] Boot Smoke green on Railway


msywulak:
@codex addressed in `8d4d7ee1`: added `.strict()` to `ProactiveConfigSchema` so any stale `proactive.channelAllowlist` (or other removed keys) fails Zod validation at boot with a clear error, instead of silently no-op'ing.

PR #2630, opened 2026-05-19
fix(plugins/chat,deploy): make slack botToken optional + drop SaaS placeholder (slice 4 dogfood blocker)

msywulak:
## The bug

`@chat-adapter/slack` line 1124:
\`\`\`js
if (!this.defaultBotToken && payload.type === \"event_callback\") {
  // multi-workspace: resolve token per-team from state adapter
} else {
  // single-workspace: use defaultBotToken as bearer
}
\`\`\`

SaaS set \`botToken: process.env.SLACK_BOT_TOKEN ?? \"saas-multi-tenant-unused\"\` — the placeholder made \`defaultBotToken\` truthy, so the adapter used it as the bearer instead of looking up the real per-tenant token from \`chat_cache\`. Slack rejected the bogus bearer; the proactive listener's \`addReaction\` silently failed.

Confirmed in dogfood: classifier ran fine (meter rows with \`action: \"react\"\`), reaction never landed, no \`react\` meter row emitted.

## Fix

1. **Plugin** — \`SlackAdapterConfig.botToken\` is now optional in both the TS interface and Zod schema. Multi-workspace deploys omit it; single-workspace deploys still need it.
2. **SaaS** — drop the placeholder; only spread \`SLACK_BOT_TOKEN\` into the config when actually set. Adapter operates in multi-workspace mode and resolves per-event tokens via \`chat_cache:slack:installation:<teamId>\`.

## Test plan

- [x] \`bun run lint\` — 0 errors
- [x] \`bun run type\` — clean
- [x] Chat plugin: 438/438 tests pass
- [ ] Boot Smoke green on Railway
- [ ] Slice 4 dogfood retry produces the 🤖

### DIFF
--- a/deploy/api/atlas.config.ts
+++ b/deploy/api/atlas.config.ts
@@ -82,15 +82,20 @@ export default defineConfig({
   // handle slash commands, block actions, modals, and OAuth.
   //
   // SaaS is multi-tenant: real Slack bot tokens live in the internal
-  // DB (slack_installations) keyed by team_id. The static `botToken`
-  // below is required by Zod (min(1)) but is never used for outbound
-  // calls in SaaS — `createChatPluginExecuteQuery()` resolves the
-  // per-tenant token via `getBotToken(teamId)` inside the agent loop.
+  // DB (slack_installations) keyed by team_id and are backfilled into
+  // the chat plugin's installation store (chat_cache:slack:installation:<teamId>).
+  // We intentionally OMIT `botToken` so `@chat-adapter/slack` operates in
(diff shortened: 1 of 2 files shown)


msywulak:
Superseded by #2632 — that PR absorbs this diff verbatim (botToken optional + drop placeholder) and adds Zod regex hardening (`^xox[a-z]` for botToken, 32-char hex for signingSecret) that would have failed CI on the original `saas-multi-tenant-unused` placeholder. Closing.

Issue #2622, opened 2026-05-19
feat(api,web): proactive classifier decisions drill-down + misfire labelling


msywulak:
Follow-up to the proactive listener wiring trail (parent #2607). The 1.5.0 release shipped per-event tracking — every classifier call writes one row to `proactive_meter_events` with confidence, metadata (`isQuestion` / `llmInvoked` / `candidate` / `classifierErrored` / `action` / `reason`), tokens, cost, and `message_id`. But the admin UI at `/admin/proactive-chat` only renders aggregate counters — no way to scroll through individual decisions, no way to mark a decision as "this was a misfire."

Result: the PRD's <5% misfire / ≥70% acceptance bar is computable in SQL but not visible to the admin, and there's no labelling loop to train against. Once design-partner dogfood data starts flowing post-#2607, we'll want to spot-check classifier behaviour and capture false-positive labels for eval.

## Key files

- `packages/api/src/api/routes/admin-proactive-analytics.ts` — existing aggregate endpoint; add a sibling `/events` endpoint that returns paginated meter rows
- `packages/api/src/lib/proactive/answer-meter.ts` — `SUMMARY_SQL` lives here; new `listEvents(workspaceId, { sinceMs, eventType?, cursor? })` adds beside it
- `packages/api/src/lib/db/migrations/` — new migration for `proactive_classification_review` (`message_id` PK, `verdict` enum: `misfire` / `correct` / `unsure`, `reviewer_user_id`, `note`, `created_at`)
- `packages/api/src/lib/db/schema.ts` — mirror the new table per CLAUDE.md drift rule
- `packages/web/src/app/admin/proactive-chat/page.tsx` — add an "Events" tab or `/admin/proactive-chat/events` sub-page

## Acceptance criteria

- [ ] `GET /api/v1/admin/proactive/events?since=30d&eventType=classify&cursor=…` returns paginated meter rows joined with the review verdict (if any). Enterprise-gated like analytics. `admin:audit` permission.
- [ ] New `proactive_classification_review` table (migration + `pgTable` + real-PG smoke). Composite key on `(workspace_id, message_id)`. CHECK constraint on `verdict`.
- [ ] `POST /api/v1/admin/proactive/events/:messageId/review` upserts a verdict. Writes a `proactive.review` `admin_action_log` row (add to `audit/actions.ts`).
[...]

PR #2632, opened 2026-05-19
fix(api,plugins/chat,deploy): #2607 aggregate review fixes — F-55 proactive, await, schema hardening, audit dual-write

### LINKED ISSUE #2631: AC drift: proactive onMeterEvent should also emit to admin_action_log
## Context

Surfaced during slice 4 dogfood of #2607. The proactive listener calls `onMeterEvent` for every `classify` + `react` event. The SaaS wiring (#2625) maps this to `recordMeterEvent` in `packages/api/src/lib/proactive/answer-meter.ts` — which writes to `proactive_meter_events` ✓ but does NOT emit to `admin_action_log`.

The original #2610 AC said:

> `onMeterEvent` — write to `proactive_meter_events` via `lib/proactive/answer-meter.ts` **and emit the matching `admin_action_log` row using `ADMIN_ACTIONS.proactive.{classify,react,answer}`** from `lib/audit/actions.ts:668-670`

Verified missing post-dogfood:

```sql
SELECT count(*) FROM admin_action_log
 WHERE org_id = 'VgazIn9VLcR1wkqouZxZ759cBEuvg0Nq'
   AND action_type LIKE 'proactive.%'
   AND timestamp > NOW() - INTERVAL '1 hour';
-- → 0
```

Meanwhile `proactive_meter_events` has the corresponding rows ✓.

## Fix

In `recordMeterEvent` (or its caller in the SaaS wiring), after the meter row is written, emit `ADMIN_ACTIONS.proactive.{classify,react,answer}` to `admin_action_log` with:
- `org_id`: from the meter event's `workspace_id`
- `target_id`: `channel_id` or `message_id` (matches the existing audit pattern for surface-scoped events)
- `actor_id`: from the meter event's `actor_user_id` (or a synthetic system actor for unattended classifies)
- `metadata`: the relevant subset of the meter event's metadata (decision reason, confidence)

## Acceptance criteria

- [ ] `recordMeterEvent` (or a wrapper at the SaaS wiring site) emits to `admin_action_log` on every meter event
- [ ] `proactive.classify` / `proactive.react` rows land with correct `org_id` + `target_id` + `actor_id`
- [ ] Unit test asserts the audit emit happens alongside the meter write
- [ ] `/admin/proactive-chat` analytics surfaces audit rows where it expects them

msywulak:
## Summary

Aggregate fixes from the multi-agent review of the proactive listener wiring trail (parent #2607). Subsumes #2630 (botToken optional + drop placeholder) and closes #2631 (admin_action_log dual-write AC drift). Adds #2623 item 3 (slash-command resolver-throw refuse).

**P0 bug fixes**
- `executeQuery.ts:256` — `await createConversation` so DB errors propagate to the surrounding try/catch and successive thread events can't race the conversation row insert.
- `answer-adapter.ts:338-365` — F-55 fail-closed. `resolveOrgForUser` throwing was logged-and-continued with `orgId=null`, short-circuiting `checkApprovalRequired` on the proactive path. Match the interactive path by letting the throw propagate to the adapter's outer catch. Same module: refuse rather than silently run the agent with `actor=null` when `resolveActor` returns null.
- `slack.ts:480-492` — verify Slack signature BEFORE parsing JSON on the legacy events route. Matches the rest of the file's defense-in-depth ordering.

**P1 / Hardening**
- `plugins/chat/src/config.ts` — `.strict()` on `ChatConfigSchema` top-level and every per-adapter schema. `SlackAdapterSchema`: `botToken` optional + regex `^xox[a-z]`; `signingSecret` regex `^[0-9a-f]{32}$`. Both reject the SaaS `"saas-multi-tenant-unused"` placeholder that put `@chat-adapter/slack` into single-workspace mode and broke #2630 in production.
- `deploy/api/atlas.config.ts` — drop the `botToken` placeholder (subsumes #2630) and the `signingSecret` placeholder. `signingSecret` resolves from `SLACK_SIGNING_SECRET` only; empty string → schema rejects → boot fails.
- `plugins/chat/src/proactive/listener.ts:611` — `workspaceId` on the quota error log so multi-tenant outages are correlatable.

**#2631 — admin_action_log dual-write**
- `answer-meter.ts:recordMeterEvent` — every meter event whose type maps to an `ADMIN_ACTIONS.proactive.*` constant (`classify`, `react`, `accept→answer`, `feedback`) now emits a `logAdminAction` sibling row alongside the `proactive_meter_events` insert. Audit emit fires BEFORE the DB-skip return so `/admin/proactive-chat` analytics that join on `admin_action_log` stay in lockstep with the meter. Closes the #2610 AC drift surfaced during dogfood at 16:57:50Z 2026-05-19. `offer` / `public_refused` have no sibling constant and stay meter-only.

**#2623 item 3 — slash command resolver-throw refuse**
- `plugins/chat/src/bridge.ts:onSlashCommand` — when the user explicitly typed `/atlas feedback <text>` and `resolveWorkspaceId` throws, refuse with an ephemeral instead of falling through to the standard question flow. Previously \`/atlas feedback the chart was wrong\` would run the text as a SQL query on resolver failure.

**Tests**
- New: `answer-meter-audit.test.ts` (7 cases) pinning the dual-write, including the offer/public_refused meter-only branch.
- Updated: `answer-adapter.test.ts` — replaced "degrades to anonymous identity" pin with two refusal cases (deleted account + F-55 regression for `resolveOrgForUser` throw). Both assert `observedRunAgentCalls` is empty.
- Updated: `bridge.test.ts` / `reactions.test.ts` fixtures — `signingSecret` now uses a 32-char hex string instead of `"test-secret"` so the new Zod regex accepts them.

All checks green locally: type 0, lint 0, api tests 67/67, plugin tests 438/438.

## Closes / Supersedes

- Closes #2631 (admin_action_log dual-write)
- Supersedes #2630 (botToken optional + drop placeholder) — its diff is absorbed verbatim plus the regex hardening
- Picks up #2623 item 3 of 6 (slash-command resolver-throw refuse)

## Deferred to follow-up PRs

- **#2624** — ProactiveUserResolver needs workspace context (plugin contract change, breaking; design follows #2620's resolveWorkspaceId pattern). Needs its own RFC + plugin bump.
- **#2623 items 1, 2, 4, 5, 6** — discriminated unions, ResolverEventLite, enabled-gate retry, schema CHECK on empty workspace_id, sentinel cache test. Polish-PR worthy.
- **#2622** — classifier-decisions drill-down + misfire labelling (new endpoint + new table + UI). Forward-looking feature, scoped post-#2607.
- **Integration test category** — \`packages/api/src/__tests__/integration/proactive-slack-e2e.test.ts\` (~250 LOC, 5 cases, uses TEST_DATABASE_URL). Both #2628 + #2630 passed all unit tests + Boot Smoke and were caught only at dogfood; this is the systemic gap.
- **chat-adapter/slack installation store vs Atlas slack_installations** — two separate stores bridged today via \`internal/backfill-chat-installations.ts\`. Long-term consolidation.
- **internal/ operational scripts** — gitignored, production-touching. Worth promoting setup-dogfood.ts to a CLI subcommand; rest can live with a tracked README.

## Test plan

- [ ] CI green (lint, type, api-tests 1-4, drift, deploy validation, symlink stub build)
- [ ] No regressions on existing answer-meter consumers (audit rows now appear for proactive.classify/react/answer/feedback; consumers that grep for these strings will start seeing them)
- [ ] Verify on dogfood after merge: a proactive.classify row in admin_action_log lands alongside the proactive_meter_events row
- [ ] Verify on dogfood after merge: \`/atlas feedback ...\` no longer runs as a SQL question when tenant resolution fails

### DIFF
--- a/deploy/api/atlas.config.ts
+++ b/deploy/api/atlas.config.ts
@@ -81,15 +81,10 @@ export default defineConfig({
   // routes in packages/api/src/api/routes/slack.ts stay put — they
   // handle slash commands, block actions, modals, and OAuth.
   //
-  // SaaS is multi-tenant: real Slack bot tokens live in the internal
-  // DB (slack_installations) keyed by team_id and are backfilled into
-  // the chat plugin's installation store (chat_cache:slack:installation:<teamId>).
-  // We intentionally OMIT `botToken` here so `@chat-adapter/slack`
-  // operates in multi-workspace mode and resolves the per-event token
-  // via `resolveTokenForTeam()`. Passing a placeholder string puts the
-  // adapter in single-workspace mode and that string ends up as the
-  // Slack API bearer (rejected with `invalid_auth`). The schema now
-  // refuses any non-`xox*` botToken so misconfiguration fails fast in CI.
+  // Multi-tenant SaaS: real bot tokens live in `slack_installations`
+  // and are resolved per-event by `@chat-adapter/slack` from its
+  // installation store (`chat_cache:slack:installation:<teamId>`).
+  // OMIT `botToken` so the adapter operates in multi-workspace mode.
   plugins: [
     chatPlugin({
       adapters: {
--- a/deploy/api/atlas.config.ts
+++ b/deploy/api/atlas.config.ts
@@ -84,11 +84,12 @@ export default defineConfig({
   // SaaS is multi-tenant: real Slack bot tokens live in the internal
   // DB (slack_installations) keyed by team_id and are backfilled into
   // the chat plugin's installation store (chat_cache:slack:installation:<teamId>).
-  // We intentionally OMIT `botToken` so `@chat-adapter/slack` operates in
-  // multi-workspace mode and resolves the per-event token via
-  // `resolveTokenForTeam()`. Passing a placeholder string would put the
-  // adapter in single-workspace mode and the placeholder would be sent as
-  // the Slack API bearer token (rejected with `invalid_auth`).
+  // We intentionally OMIT `botToken` here so `@chat-adapter/slack`
+  // operates in multi-workspace mode and resolves the per-event token
+  // via `resolveTokenForTeam()`. Passing a placeholder string puts the
+  // adapter in single-workspace mode and that string ends up as the
+  // Slack API bearer (rejected with `invalid_auth`). The schema now
+  // refuses any non-`xox*` botToken so misconfiguration fails fast in CI.
   plugins: [
     chatPlugin({
       adapters: {
@@ -96,8 +97,10 @@ export default defineConfig({
           ...(process.env.SLACK_BOT_TOKEN
             ? { botToken: process.env.SLACK_BOT_TOKEN }
             : {}),
-          signingSecret:
-            process.env.SLACK_SIGNING_SECRET ?? "saas-multi-tenant-unused",
+          // Single shared signing secret for the Atlas Slack app — Zod
+          // now rejects placeholder strings so an unset env var fails
+          // boot rather than silently passing webhook verification.
+          signingSecret: process.env.SLACK_SIGNING_SECRET ?? "",
           ...(process.env.SLACK_CLIENT_ID
             ? { clientId: process.env.SLACK_CLIENT_ID }
             : {}),
--- a/packages/api/src/api/__tests__/slack.test.ts
+++ b/packages/api/src/api/__tests__/slack.test.ts
@@ -469,6 +469,27 @@ describe("/api/v1/slack", () => {
       expect(json.error).toBe("invalid_signature");
     });
 
+    it("rejects invalid signature with 401 even when body is malformed JSON (verify before parse)", async () => {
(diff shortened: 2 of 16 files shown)


msywulak:
Rebased onto main (#2630 was already merged while this PR was being prepared). The conflicts were on the same files where #2630 made a softer version of what this PR contains:

- `deploy/api/atlas.config.ts` — kept this PR's version (`signingSecret: process.env.SLACK_SIGNING_SECRET ?? ""`) since the schema now rejects empty/non-hex strings, so a missing env var fails boot. #2630 left the "saas-multi-tenant-unused" fallback in place; this PR removes it.
- `plugins/chat/src/config.ts` — kept this PR's regex'd version of `botToken` (`^xox[a-z]`) and `signingSecret` (32-char hex) on top of #2630's optional-botToken change. Without the regex hardening, a future placeholder slip-up would still pass Zod.

All checks still green locally post-rebase: type 0, lint 0, api 67/67, plugin 438/438.

Issue #2677, opened 2026-05-20
Audit: chat-plugin migration ported happy path but keeps missing Atlas-extension contracts (3rd instance)


msywulak:
## Context

The 1.5.2 chat-plugin migration (#2607 trail: #2613 → #2617 → #2618 → #2619 → #2621 → #2625 → #2626 → #2629) replaced the legacy `packages/api/src/api/routes/slack.ts` paths with `plugins/chat/src/...` routes. The architectural intent is clean — but in three back-to-back incidents the migration has shipped the receive-path / happy-path correctly while silently dropping an Atlas-side state contract that the legacy code-path owned:

| # | Hotfix PR | What was dropped | Symptom |
|---|---|---|---|
| 1 | #2628 | `channelAllowed` read from `channel_proactive_config` — new code-path read a legacy env-var instead | Proactive 🤖 didn't fire even when the channel row had `allow=true` |
| 2 | #2630 | `botToken` placeholder forced single-workspace mode; multi-tenant chat-adapter couldn't look up real tokens | Token lookups in multi-tenant Slack flow degraded silently |
| 3 | #2676 | `orgId` field in `chat_cache` — new OAuth callback delegates to `@chat-adapter/slack`'s `setInstallation` which is Atlas-agnostic | Today's outage. `resolveWorkspaceId` returned null, every event resolved as "unknown tenant", silent skip |

Each fix was correct in isolation. The recurring root cause is the **absence of a single audit** of "what fields, behaviours, or invariants did the legacy `slack.ts` (or pre-1.5.0 proactive code) own that the new chat-plugin path is expected to also own?"

## Problem

The chat plugin and `@chat-adapter/*` packages are intentionally Atlas-agnostic — they implement a generic chat protocol, and Atlas extensions (`orgId`, `workspaceName`, proactive config schema, content-mode RLS) ride alongside via dedicated columns / extension fields in shared tables. Every code path that crosses the Atlas/chat-plugin boundary needs to consciously thread the Atlas extension forward.

When a code path is rewritten without that audit, the chat-plugin side gets the migration but the Atlas-extension contract gets dropped. It compiles, tests pass (the extension isn't in the chat-plugin's test surface), CI is green, and the bug only surfaces on a real OAuth re-install or a config edge case — which dogfood eventually hits.

[...]

Issue #2676, opened 2026-05-20
Chat-plugin OAuth callback drops Atlas `orgId` extension field on (re)install


msywulak:
## Context

Today's proactive-chat outage in `#sandbox-atlas` had two stacked root causes. Both are now patched in prod:

1. `SLACK_ENCRYPTION_KEY` was missing → chat plugin's Slack adapter never instantiated. See #2672 / #2673 for the boot-guard + observability fixes.
2. **`chat_cache.value.orgId` was missing on the `T0AQLT0EBMJ` row** → even after the adapter came back, `resolveWorkspaceId` (`packages/api/src/lib/proactive/workspace-id-resolver.ts:103`) returned `null`, every event resolved as "unknown tenant" and the proactive listener silently skipped. Fixed in prod by an out-of-band `UPDATE chat_cache SET value = jsonb_set(value, '{orgId}', '"<orgId>"') WHERE key = 'slack:installation:T0AQLT0EBMJ'`.

This issue is the prevention fix for (2).

## Problem

The chat plugin's `/oauth/slack/callback` route (`plugins/chat/src/index.ts:387`) delegates persistence to the chat-adapter directly:

```ts
const result = await slackAdapterInstance.handleOAuthCallback(c.req.raw);
```

`@chat-adapter/slack@4.23.0`'s `handleOAuthCallback` internally calls its own `setInstallation`, which writes the `chat_cache` row with **only** the fields the chat-adapter knows about: `botToken`, `botUserId`, `teamId`, `teamName`. The Atlas-extension field `orgId` — required by `resolveWorkspaceId`, `getInstallationByOrg`, and the `idx_chat_cache_slack_org_id` partial expression index — is never written by this path.

`packages/api/src/lib/slack/store.ts:saveInstallation` (the helper that *does* preserve `orgId` via `chat_cache.value || EXCLUDED.value` JSONB merge with hijack protection) is invoked from the admin-side BYOT install path (`admin-integrations.ts:705`) and the legacy `slack.ts` callback, but is **not** called by the new chat-plugin OAuth callback. So every re-install through the chat-plugin path overwrites the row and silently drops `orgId`.

[...]

Issue #2673, opened 2026-05-20
AdapterRegistry: log at error (not info) when a SaaS-eligible catalog entry can't be instantiated


msywulak:
## Context

`plugins/chat/src/adapter-registry.ts:244` logs `AdapterRegistry: required env vars missing — adapter not instantiated` at **info** level when a catalog entry's builder returns `null`. Same with `Chat interaction plugin initialized (no chat adapters activated — see AdapterRegistry warns, state: pg)` at `plugins/chat/src/index.ts`. Both render as `[INFO]` in Railway logs and blend into routine boot noise.

This bit us in prod on 2026-05-20: a single dropped `SLACK_ENCRYPTION_KEY` left the chat plugin running with `adapters=[]` — no Slack events received for ~22 hours — and the `[INFO]` lines were missed in Railway's log stream because they look like routine startup info. See companion issue #2672 for the boot-guard fix.

## Problem

For a SaaS deploy, a catalog entry where `saas_eligible: true` and `enabled: true` fails to instantiate (missing env vars, builder absent, etc.) is **always** a critical operational issue. There is no legitimate reason for a SaaS deploy to opt a Platform into the catalog and then run without it. Today the registry treats this case the same as a self-hosted dev box that hasn't bothered to set the envs — both get `info`-level "skipped".

## Proposed fix

Differentiate by intent rather than by deploy mode (which the registry doesn't know about today):

1. **`missingCredSlugs.length > 0` should log at `error`, not `warn`**, when the catalog entry was explicitly enabled. The current `warn` for "required env vars missing" + `info` for "no chat adapters activated" is too quiet.
2. The "no chat adapters activated" line in `plugins/chat/src/index.ts` should escalate to `log.error` when `adapters=[]` AND the catalog had at least one `enabled: true` entry — `info` is wrong because that combination is never intentional.
3. Optional: take a `deployMode` hint as a builder arg so the registry can hard-throw in SaaS mode (defense-in-depth alongside the core boot guard in #2672).

(1) and (2) are 5-line changes. (3) is the more interesting design question — defer until #2672 lands and see if it's still needed.

## Key files

- `plugins/chat/src/adapter-registry.ts` — escalate the `info`/`warn` calls when the entry was explicitly enabled
- `plugins/chat/src/index.ts` — escalate the "no chat adapters activated" log when adapters is empty AND catalog had an enabled chat entry
- `plugins/chat/src/__tests__/adapter-registry.test.ts` — pin the new log levels

[...]

Issue #2672, opened 2026-05-20
SaaS boot guard: fail when chat catalog enables Slack but SLACK_ENCRYPTION_KEY is unset


msywulak:
## Context

Today proactive chat in `#sandbox-atlas` silently stopped working between 2026-05-19 17:59 UTC (last meter row) and 2026-05-20 16:39 UTC (most recent api boot before the fix). Diagnosis: `SLACK_ENCRYPTION_KEY` was unset on all three Railway api services in production (`api`, `api-eu`, `api-apac`). The chat plugin's `AdapterRegistry.SLACK_BUILDER.build()` requires all four envs (`SLACK_CLIENT_ID`, `_SECRET`, `_SIGNING_SECRET`, `_ENCRYPTION_KEY`); with one missing it returns `null` and the adapter is silently dropped. The api still boots, the proactive listener still registers, and the admin analytics endpoints keep returning 200s — every signal looks green except actual events.

Today's boot trail (`5d4363c6fc3c`, 16:39 UTC):

```
[INFO] AdapterRegistry: required env vars missing — adapter not instantiated
       requiredEnv=["SLACK_CLIENT_ID","SLACK_CLIENT_SECRET","SLACK_SIGNING_SECRET","SLACK_ENCRYPTION_KEY"]
[INFO] Proactive listener registered  killSwitch=true answerFlow="both" feedback=false
[INFO] Chat interaction plugin initialized (no chat adapters activated — see AdapterRegistry warns, state: pg)
       adapters=[] stateBackend="pg"
```

The fix to restore service was to generate a 32-byte hex key and set `SLACK_ENCRYPTION_KEY` on all three api services. Post-fix boot (`cce3483aa3f5`, 17:53 UTC) logs `AdapterRegistry: chat adapter registered slug="slack"` and the existing plaintext bot token in `chat_cache` for `T0AQLT0EBMJ` is still readable (read path tolerates both shapes per `installation-encryption.ts:57`).

## Problem

[...]

PR #2680, opened 2026-05-20
fix: proactive reaction-back never fired in prod — pending keyed on bare channelId

msywulak:
## Summary

- **The bug**: `plugins/chat/src/proactive/listener.ts:814` recorded pending answers with `pending.record(thread.channelId, ...)` (= bare `"slack:CHANNEL"`), but the reaction-back handler looks them up with `pending.peek(event.threadId, ...)` (= encoded `"slack:CHANNEL:THREAD_TS"`). Keys never matched. Every reaction-back since the post-#2607 proactive listener wiring has fallen through to the `unknown-message` skip branch in `shouldAnswerOnReaction` (debug-level log, invisible at default info, no meter row, no answer flow).
- **The fix**: `pending.record(thread.id, ...)` (the encoded form, matching `event.threadId`).
- **The hardening**: `ThreadId` + `ChannelId` brands added to `@useatlas/types/proactive` (type-only); `assertThreadId` + `assertChannelId` runtime helpers in `plugins/chat/src/proactive/identity.ts` (matching the #2641 pattern). `PendingAnswers.record/peek/consume` signatures now require `ThreadId`, so `pending.record(thread.channelId, ...)` is a compile error.
- **Test fixture rebuild**: `makeThread` now mirrors Slack's encoded shape (`id: "slack:CHANNEL:THREAD_TS"`, `channelId: "slack:CHANNEL"`). Reaction-event fixtures pass `threadId: thread.id` rather than `thread.channelId`. The pre-fix mock collapsed both into `"C-allowed"` and silently masked the prod bug — CI green, prod broken for the entire dogfood window.

## Context

This was the 4th back-to-back "1.5.2 chat-plugin migration ported the happy path but dropped a Slack-specific contract" bug surfaced today, joining the trail in #2677:

| # | Issue / PR | Dropped contract |
|---|---|---|
| 1 | #2628 | `channelAllowed` read from a legacy env-var, not `channel_proactive_config` |
| 2 | #2630 | Slack `botToken` placeholder forced single-workspace mode |
| 3 | #2676 / #2678 | `chat_cache.value.orgId` clobbered on OAuth re-install (state-adapter overwrite) |
| 4 | **This PR** | `pending.record(thread.channelId, ...)` vs `pending.peek(event.threadId, ...)` — semantically-distinct strings collapsed into `string` |

Discovered during the dogfood arc that produced #2672 / #2673 / #2676 / #2677. Synthetic `reaction_added` probe reached the chat-adapter (`[chat-sdk:slack] Failed to resolve parent thread for reaction, using message ts`) — proving the receive path works — but `pending.peek` returned null because the recorded key shape diverged from the lookup key shape.

## Why brands at the `@useatlas/types` boundary, not in the chat SDK

`chat@4.23.0` is upstream (Vercel-maintained) and treated as gospel — branding `Thread.id`/`Thread.channelId` upstream would require coordination with the SDK maintainer and a release. The Atlas-side fix is to brand at our own `@useatlas/types` seam and assert at the listener boundary (`assertThreadId(thread.id)`). This matches the existing pattern shipped in #2641 for `WorkspaceId` / `AtlasUserId` / `ExternalUserId` and keeps the chat SDK untouched.

## @useatlas/types version bump

`@useatlas/types` is bumped from 0.1.4 → 0.1.5. New exports are type-only (`ThreadId`, `ChannelId`) — safe per the scaffold-CI gotcha (value exports break scaffold CI; type-only exports don't). Per the version-bump-ordering memory rule:

1. Land this PR (types version bump + new exports + plugin consumers using `workspace:*` locally).
2. After merge, tag `types-v0.1.5` and let OIDC publish.
3. Follow-up PR bumps the `">=0.1.3"` ref in `plugins/chat`, `packages/sdk`, `packages/react`, and template `package.json` files to `">=0.1.5"`.

## Files

- `packages/types/src/proactive.ts` — adds `ThreadId` + `ChannelId` brands next to `WorkspaceId`/`AtlasUserId`/`ExternalUserId`
- `packages/types/package.json` — 0.1.4 → 0.1.5
- `plugins/chat/src/proactive/identity.ts` — `assertThreadId` + `assertChannelId` runtime helpers, `ProactiveIdentityField` union extended
- `plugins/chat/src/proactive/answerer.ts` — `PendingAnswers.record/peek/consume` now require `ThreadId`
- `plugins/chat/src/proactive/listener.ts` — the one-line bug fix (line 814) + brand assertions at every `pending.record/peek/consume` call site
- `plugins/chat/src/proactive/index.ts` + `plugins/chat/src/index.ts` — re-export the new types + assert helpers
- `plugins/chat/src/proactive/__tests__/listener.test.ts` — `makeThread` fixture mirrors Slack encoding; 14 reaction-event call sites updated from `thread.channelId` to `thread.id`
- `plugins/chat/src/proactive/__tests__/answerer.test.ts` — `T1` / `T2` constants branded via `assertThreadId`

## Test plan

- [x] `bun run type` — clean (only pre-existing plugin-sdk noise from unrelated peer-dep gaps)
- [x] `bun run lint` — 0 errors
- [x] `bun test` in `plugins/chat` — 412 pass / 0 fail (includes 14 listener tests, 10 answerer tests)
- [ ] CI green on full suite + Deploy Validation
- [ ] After merge + types publish: bump `>=0.1.3` refs to `>=0.1.5` in plugins/chat, packages/sdk, packages/react, create-atlas templates

## Follow-ups

- Adds a fourth entry to the #2677 pattern catalogue. The structural prevention (test fixtures mirroring real platform encoding + brands at the Atlas/chat-SDK seam for every Atlas-extension contract) is now sufficiently exemplified across 4 instances — #2677 should consume this PR's pattern in its audit doc.
- The reaction-back path was never dogfood-verified in #sandbox-atlas before today. Once this PR deploys, reacting back on a 🤖 should produce a `public_refused` meter event (refusal copy because the public-dataset allowlist for the Atlas workspace is empty — unblocking that is separately tracked).


msywulak:
Thanks @chatgpt-codex-connector — good P1 catch. Fixed in `cfe3546c`.

`plugins/chat/package.json`'s `@useatlas/types` peer range bumped from `>=0.1.3` to `>=0.1.5`. The earlier reasoning ("keep the ref at old version until publish") was confused with the `0.0.x` semver rule (where `^0.0.2` pins exactly to 0.0.2); for `0.1.x`, `>=0.1.5` is the right expression of what the code actually requires.

Other refs left untouched:
- `packages/sdk`, `packages/react`, `create-atlas/templates/nextjs-standalone` all use `^0.1.0`, which already auto-accepts 0.1.5 once published — no code in those packages imports `ThreadId`, so no constraint to express
- monorepo CI continues to resolve via `workspace:*`, so the local build is unaffected

Safe to merge before `types-v0.1.5` is published — chat plugin publishes are tag-triggered, not merge-triggered, and the publish-order memory dictates `types-v0.1.5` is tagged + published before any `chat-vNEXT` tag.

Issue #2716, opened 2026-05-23
flaky: packages/cli fatal-error-propagation.test.ts DuckDB profile times out under full-suite parallel load


msywulak:
## Symptom

`@atlas/cli` test `bin/__tests__/fatal-error-propagation.test.ts > DuckDB profiler — error propagation behavior > valid database profiles successfully with all columns` times out at 5000ms when running the full `bun run test` suite in parallel. Passes in 885ms when re-run in isolation.

## Reproduction

```bash
# Fails (under parallel full-suite load — pre-PR /ci on 2026-05-23):
cd /home/matt/code/atlas && bun run test
# ✗ this test timed out after 5000ms

# Passes in isolation in ~885ms:
cd packages/cli && bun test bin/__tests__/fatal-error-propagation.test.ts
# 30 pass / 0 fail
```

## Hypothesis

Same shape as #2710 and #2694 — DuckDB native module + CSV load + profile takes longer to bootstrap when the host is under parallel-suite contention. 5s timeout in the test is tight for the cold-start path.

## Likely fixes

- Bump test timeout for the DuckDB path (cleanest)
- Mock the DuckDB profile call (loses real-DuckDB coverage)
- Drop full-suite parallelism for `@atlas/cli` (slows CI)

The timeout bump is simplest and preserves coverage. The test isn't trying to assert performance — it's checking error-propagation behavior.

## Key files

- `packages/cli/bin/__tests__/fatal-error-propagation.test.ts`

## Acceptance criteria

- [ ] Test passes reliably under full `bun run test` (run 3× in a row, all green)
- [ ] No coverage regression for the DuckDB error-propagation path

msywulak:
Fixed in PR #2831 by bumping the "valid database profiles successfully" timeout to 30s. Cold-start native module + CSV ingest + profile is ~850ms in isolation but exceeds 5s under parallel-suite contention. Test asserts behavior, not performance — bump preserves coverage. 5x consecutive isolated runs at ~440ms each.

PR #2717, opened 2026-05-23
fix(plugins/chat): #2673 — escalate AdapterRegistry log severity for SaaS misconfigs

### LINKED ISSUE #2673: AdapterRegistry: log at error (not info) when a SaaS-eligible catalog entry can't be instantiated

msywulak:
## Summary

- `AdapterRegistry: required env vars missing — adapter not instantiated` now logs at `error` (was `warn`). The entry-disabled branch above already short-circuits, so this branch is only reachable when the operator opted the Platform in — that's always a real misconfig.
- Post-init `Chat interaction plugin initialized (no chat adapters activated …)` log escalates to `error` when `adapters=[]` AND the registry's diagnostics flag a real instantiation failure (missing creds or unknown slug). Stays at `info` for intentionally-empty / all-disabled / non-OAuth catalogs.
- New exported predicate `hasInstantiationFailure(diagnostics)` gates the post-init severity — pinned by unit tests so a future refactor can't silently flip the gate.
- `RegistryLogger` interface gains `error` to match `PluginLogger`.

Motivation: 2026-05-20 outage where a dropped `SLACK_ENCRYPTION_KEY` left the chat plugin running with `adapters=[]` for ~22 hours. The warn/info lines blended into routine Railway boot noise. #2672 shipped the load-bearing SaaS boot guard; this is the observability defense-in-depth.

## Note on AC interpretation

The issue body's AC #2 says "no chat adapters activated AND the catalog has at least one `enabled: true` chat entry" → error. I narrowed this to "instantiation failure" (`missingCredSlugs` or `unrecognizedSlugs` non-empty). The broader literal would over-fire when a catalog contains only `static-bot enabled: true` entries (intentionally not wired in 1.5.2 — Telegram/Discord/gchat/WhatsApp are catalog placeholders pending 1.5.3). The diagnostics-based predicate exactly matches the production incident motivation.

## Test plan

- [x] `bun test src/adapter-registry.test.ts` (plugins/chat) — 20/20 pass including the four new `hasInstantiationFailure` cases and the disabled-entry no-error pin
- [x] Full `bun test` in `plugins/chat` — 437/437 pass
- [x] `bun run type` — clean across the workspace
- [x] `bun run lint` — clean
- [x] `/ci` gates all green except one pre-existing flake in `@atlas/cli` (DuckDB profiler, unrelated to this change) — filed as #2716

Closes #2673

### DIFF
--- a/plugins/chat/src/adapter-registry.test.ts
+++ b/plugins/chat/src/adapter-registry.test.ts
@@ -17,6 +17,7 @@
 import { describe, it, expect } from "bun:test";
 import {
   buildChatAdapterRegistry,
+  hasInstantiationFailure,
   type ChatCatalogEntry,
   type RegistryLogger,
 } from "./adapter-registry";
@@ -34,18 +35,22 @@ function makeLogger(): {
   logger: RegistryLogger;
   infos: Array<{ payload: Record<string, unknown>; msg: string }>;
   warns: Array<{ payload: Record<string, unknown>; msg: string }>;
+  errors: Array<{ payload: Record<string, unknown>; msg: string }>;
   debugs: Array<{ payload: Record<string, unknown>; msg: string }>;
 } {
   const infos: Array<{ payload: Record<string, unknown>; msg: string }> = [];
   const warns: Array<{ payload: Record<string, unknown>; msg: string }> = [];
+  const errors: Array<{ payload: Record<string, unknown>; msg: string }> = [];
   const debugs: Array<{ payload: Record<string, unknown>; msg: string }> = [];
   return {
     infos,
     warns,
+    errors,
     debugs,
     logger: {
       info: (payload, msg) => infos.push({ payload, msg }),
       warn: (payload, msg) => warns.push({ payload, msg }),
+      error: (payload, msg) => errors.push({ payload, msg }),
       debug: (payload, msg) => debugs.push({ payload, msg }),
     },
   };
@@ -91,8 +96,13 @@ describe("buildChatAdapterRegistry — missing env vars", () => {
   ] as const;
 
   for (const missing of required) {
-    it(`skips Slack + logs warn when ${missing} is missing`, () => {
-      const { logger, warns } = makeLogger();
+    it(`skips Slack + logs error when ${missing} is missing (entry enabled)`, () => {
+      // #2673 — `enabled: true` + missing creds is a silent-degradation
+      // bug on SaaS; the registry's control flow guarantees we only
+      // reach the missing-creds branch when the entry was opted in, so
+      // bump the log level from `warn` (lost in routine boot noise) to
+      // `error` (catches operator log-stream alerts).
+      const { logger, errors, warns } = makeLogger();
       const env: NodeJS.ProcessEnv = { ...SLACK_FULL_ENV };
       delete env[missing];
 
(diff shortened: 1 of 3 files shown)


msywulak:
@codex review

PR #2721, opened 2026-05-23
test(plugins/chat): #2693 — backfill bridge.onAction handler dispatch tests

### LINKED ISSUE #2693: test: backfill handler-level tests for @useatlas/chat onAction + Slack OAuth routes
## Context

While reviewing #2687 (retirement of the pre-chat-plugin Slack integration) two test gaps surfaced in `@useatlas/chat` itself. They predate #2683 — not regressions from the retirement PR — but the deleted `slack.test.ts` was the only thing exercising similar shapes, so it's worth backfilling now that the chat plugin owns the Slack surface end-to-end.

## Gap 1 — `bridge.onAction` handler not tested

`plugins/chat/src/bridge.ts:1402-1480` registers four block-action handlers via `chat.onAction([...], ...)`:

- `atlas_action_approve` / `atlas_action_deny` → invokes `config.actions.get/approve/deny`
- `atlas_run_again` / `atlas_export_csv` → other side-effect paths

`plugins/chat/src/bridge.test.ts` only asserts that the card builder *emits* buttons with the right `id`s (lines 215-221, 383-386). No test exercises:

- The handler invoking `config.actions.get(actionId)` and then `approve(actionId)` / `deny(actionId)` with the right `slack:<userId>` actor format.
- The missing-action edit path (`bridge.ts:1424-1435`) — when `actions.get` returns null/expired.
- Rate-limit / signature-replay handling at the handler layer (Bolt owns signature, but the dispatch table is ours).

## Gap 2 — `/oauth/slack/{install,callback}` routes not tested

`plugins/chat/src/index.ts:262-450` (approx) mounts:

- `app.get("/oauth/slack/install", ...)` — multi-workspace install redirect
- `app.get("/oauth/slack/callback", ...)` — bot-token exchange + persistence to `chat_cache:slack:installation:<teamId>`

`grep -rn 'oauth/slack/install\|oauth/slack/callback' plugins/chat/` returns zero matches in any `*.test.ts` file. The deleted `plugins/slack/__tests__/plugin.test.ts:950-1045` covered the analogous routes (401-when-unconfigured, 400 invalid-state, 400 missing-code) on the pre-chat-plugin path; none of that moved to the new path.

## Key files

- `plugins/chat/src/bridge.ts:1402-1480` — onAction handler dispatch table
- `plugins/chat/src/index.ts:262-450` — oauth/slack route mounts
- `plugins/chat/src/bridge.test.ts` — add handler-level cases here
- `plugins/chat/__tests__/` or `plugins/chat/src/__tests__/` — add an `oauth.test.ts` or extend the existing test file for OAuth route coverage

## Acceptance criteria

- [ ] Bridge `onAction` for all four registered actions has a unit test that confirms:
  - `config.actions.get(actionId)` is called with the expected ID
  - On success path, `approve`/`deny`/`run-again`/`export-csv` is invoked with `actor = 'slack:<userId>'`
  - On missing-action path, the message is edited with the expected fallback
- [ ] `/oauth/slack/install` returns 501 (or equivalent) when `clientId` is not configured; returns the expected redirect when configured
- [ ] `/oauth/slack/callback` returns 400 on missing `code`, 400 on bad `state`, and persists `{ botToken, botUserId, teamName, orgId }` to `chat_cache:slack:installation:<teamId>` on success (with `SLACK_ENCRYPTION_KEY` set, write a round-trip encrypt+decrypt assertion via the existing `installation-encryption` helper)
- [ ] `bun run test` green for `@useatlas/chat`

## Dependencies

None — independent of #2683 / #2687 / #2688.

## Notes

These tests would have caught a class of silent regressions that the pre-chat-plugin `slack.test.ts` covered. Bolt owns signature verification + payload parsing, so we don't need to re-test those — focus is on our dispatch + persistence logic.

msywulak:
## Summary

The chat-sdk's `Chat` class isn't trivially mockable without reproducing its full type surface. Extract the approve/deny handler body from the `chat.onAction(...)` lambda into a pure helper `dispatchApproveDenyAction` so tests can pass plain structurally-compatible event objects and assert dispatch behaviour directly.

- New exported helper `dispatchApproveDenyAction(event, actions, log)` accepts a structural `ApproveDenyActionEvent` subset (`actionId`, `value?`, `user.userId`, `threadId`, `messageId`, `adapter.editMessage`) — chat-sdk's `ActionEvent` is compatible at the call site, so runtime behaviour is identical.
- Inline `chat.onAction(["atlas_action_approve", "atlas_action_deny"], …)` lambda becomes a 1-line pass-through.
- New test file `bridge-actions.test.ts` with 9 cases covering: approve/deny happy paths (asserting the `chat-sdk:<userId>` actor format), missing `event.value`, `actions.get` null fallback, already-resolved fallback for both approve and deny, thrown-error fallback, and survives-editMessage-failure recovery.

## Note on issue AC scope

- **Gap 2 (OAuth route tests)** is moot — #2689 retired the chat-plugin's `/oauth/slack/{install,callback}` routes in favor of the canonical `/api/v1/integrations/slack/{install,callback}` surface, which already has coverage at `slack-oauth-handler.test.ts`.
- **Actor format**: the issue body referenced `slack:<userId>` but production uses `chat-sdk:<userId>` (settled during slice 5/6). Tests pin the production value.
- **Run Again / Export CSV handlers**: out of scope for this PR. The same extract-helper pattern applies; can be filed as a follow-up if there's demand.

## Test plan

- [x] `bun test src/bridge-actions.test.ts` (plugins/chat) — 9/9 pass, 45 expect() calls
- [x] Full `bun test` (plugins/chat) — 442/442 pass (no regressions in existing bridge.test.ts lifecycle tests)
- [x] `bun run type` — clean across the workspace
- [x] `bun run lint` — clean

Closes #2693

### DIFF
--- a/plugins/chat/src/bridge-actions.test.ts
+++ b/plugins/chat/src/bridge-actions.test.ts
@@ -1,0 +1,341 @@
+/**
+ * Tests for `dispatchApproveDenyAction` — the approve/deny button
+ * handler dispatch path extracted from the `chat.onAction(...)` lambda
+ * inside `createChatBridge` (#2693).
+ *
+ * Asserted invariants (issue #2693 AC, narrowed to the approve/deny
+ * subset — Gap 2 OAuth tests are owned by `slack-oauth-handler.test.ts`
+ * post-#2689):
+ *
+ *   - On approve: `actions.get(actionId)` then `actions.approve(actionId, "chat-sdk:<userId>")`,
+ *     then `adapter.editMessage` with the formatted result.
+ *   - On deny: same shape with `actions.deny`.
+ *   - On missing action (`get` returns null): fallback markdown edit.
+ *   - On already-resolved (approve|deny returns null): "already resolved" edit.
+ *   - On thrown error: error log + fallback edit + survive editMessage failure.
+ *   - Empty `event.value`: warn + return (no `actions.get` call).
+ *
+ * The chat-sdk's `ActionEvent` is structurally compatible with the
+ * helper's `ApproveDenyActionEvent` subset, so the test passes plain
+ * objects without constructing real chat-sdk types.
+ */
+
+import { describe, it, expect, mock } from "bun:test";
+import {
+  dispatchApproveDenyAction,
+  type ApproveDenyActionEvent,
+} from "./bridge";
+import type { ActionCallbacks } from "./config";
+import type { PluginLogger } from "@useatlas/plugin-sdk";
+
+// ---------------------------------------------------------------------------
+// Test fixtures
+// ---------------------------------------------------------------------------
+
+function makeLogger(): {
+  log: PluginLogger;
+  infos: Array<unknown[]>;
+  warns: Array<unknown[]>;
+  errors: Array<unknown[]>;
+} {
+  const infos: Array<unknown[]> = [];
+  const warns: Array<unknown[]> = [];
+  const errors: Array<unknown[]> = [];
+  return {
+    infos,
+    warns,
+    errors,
+    log: {
+      info: (...args: unknown[]) => infos.push(args),
+      warn: (...args: unknown[]) => warns.push(args),
+      error: (...args: unknown[]) => errors.push(args),
+      debug: () => {},
+    } as unknown as PluginLogger,
+  };
+}
+
+function makeEvent(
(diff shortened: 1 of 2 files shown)


msywulak:
@codex review

PR #2722, opened 2026-05-23
test(api): #2714 — real-Postgres smoke for Better-Auth-dependent migrations

### LINKED ISSUE #2714: test: migrate-pg.test.ts smoke skips MANAGED_AUTH_MIGRATIONS — Better-Auth-touching migrations have no real-Postgres coverage
## Context

Surfaced by the #2713 (entitlement bundle) PR review. Migration \`0090_organization_is_operator_workspace.sql\` adds a column to Better Auth's \`organization\` table — a real production deploy runs Better Auth migrations first, then Atlas migrations against the resulting tables. The mock-pool tests cover the SQL planning shape, but the real-Postgres \`migrate-pg.test.ts\` smoke runs with \`skip: MANAGED_AUTH_MIGRATIONS\`, so 0090 (and 0027 before it, and any future Better-Auth-table migration) never executes against a real Postgres in CI.

A regression that introduces a planner-level error against the actual \`organization\` shape would not be caught until the next production deploy.

## What's missing

\`migrate-pg.test.ts\` doesn't run Better Auth migrations before Atlas migrations, so MANAGED_AUTH_MIGRATIONS entries are skipped. Adding a step like:

\`\`\`ts
// Before runMigrations:
await runBetterAuthMigrations(pool); // — needs to exist
await runMigrations(pool); // — no skip
\`\`\`

would let us cover:

1. Migration 0090 — \`ALTER TABLE organization ADD COLUMN is_operator_workspace\` actually applies.
2. Migration 0027 — \`organization\` SaaS columns.
3. \`WorkspaceInstallGate\`'s JOIN through \`organization.is_operator_workspace\` round-trips on real data.

The acceptance test would then INSERT an org with \`is_operator_workspace = true\`, run the gate's SQL JOIN, assert the operator-bypass branch fires.

## Approach options

A. **Add a Better Auth bootstrap step to migrate-pg.test.ts.** Better Auth ships a SQL DDL extract — invoke it before \`runMigrations\` and drop the skip. Simplest but couples the test to BA's migration output format.

B. **Use Better Auth's runtime auto-migrator.** \`auth.api.migrate\` (or whatever Better Auth's runtime equivalent is) creates the tables. Heavier — pulls Better Auth's full stack into the test.

C. **Add a separate \`migrate-pg-with-auth.test.ts\`** suite that bootstraps both. Keeps the existing smoke fast for non-auth-touching migrations; the new suite runs the slower full pipeline.

C is probably right — keeps the per-test cost bounded for the common case while adding the coverage we're missing.

## Key files

- \`packages/api/src/lib/db/__tests__/migrate-pg.test.ts\` — existing smoke
- \`packages/api/src/lib/db/internal.ts\` — \`MANAGED_AUTH_MIGRATIONS\` list
- \`packages/api/src/lib/integrations/install/workspace-install-gate.ts\` — the consumer of the missing-from-smoke column

## Acceptance criteria

- [ ] CI runs Better Auth + Atlas migrations end-to-end against a real Postgres in some test job.
- [ ] Inserting an organization row with \`is_operator_workspace = true\` and SELECT'ing through the WorkspaceInstallGate JOIN returns the operator-bypass branch.
- [ ] A regression that drops or renames the column fails the smoke test loudly.

## Dependencies

None — independent of #2701/#2702 (those shipped without this coverage). File this as a 1.5.3 or architecture-backlog item.

## Notes

Surfaced from PR #2713 review (pr-test-analyzer agent). Logging as a follow-up rather than expanding the entitlement bundle's scope.

msywulak:
## Summary

The companion `migrate-pg.test.ts` runs with `skip: MANAGED_AUTH_MIGRATIONS` because Better Auth's `user`, `session`, `organization` tables don't exist in the test fixture. That leaves six migrations (0027 / 0042 / 0048 / 0050 / 0061 / 0090) without real-Postgres coverage — a regression that breaks one of them would only surface on the next production deploy. Surfaced from the #2713 PR review (pr-test-analyzer): 0090 adds a column to the Better-Auth-managed `organization` table and is in the skip list.

New test file `migrate-pg-with-auth.test.ts` bootstraps a minimal Better-Auth schema (just the columns Atlas migrations reference on `user` / `session` / `organization`), then runs the FULL Atlas migration set with no skip. Hand-crafted DDL vs invoking Better Auth's own migrator: keeps the test surface focused on Atlas — a Better Auth schema-shape change won't break this test unless it actually intersects an Atlas migration.

## Coverage

- All 92 migrations apply end-to-end (vs 86 with the skip list)
- Idempotent re-run applies zero new migrations
- **0090**: `organization.is_operator_workspace` round-trips `true`/`false`; the same row pins **0027**'s `plan_tier` ('free' default) + `workspace_status` ('active' default)
- **0061**: `user.default_landing` carries the 'chat' default
- **0048**: `trusted_device` FK to `user` enforces ON DELETE CASCADE

## Verification

Verified locally against `postgres:16`:
```
TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5499/atlas \
  bun test src/lib/db/__tests__/migrate-pg-with-auth.test.ts
# 5 pass / 0 fail / 9 expect() calls / 1459ms / applied=92, total=92
```

Skips cleanly when `TEST_DATABASE_URL` is unset (same gate as the companion smoke).

## Test plan

- [x] `bun run type` — clean
- [x] `bun run lint` — clean
- [x] Local PG run: 5/5 pass with 92 migrations applied (full set, no skip)

Closes #2714

### DIFF
--- a/packages/api/src/lib/db/__tests__/migrate-pg-with-auth.test.ts
+++ b/packages/api/src/lib/db/__tests__/migrate-pg-with-auth.test.ts
@@ -1,0 +1,214 @@
+/**
+ * Real-Postgres migration smoke INCLUDING Better-Auth-dependent
+ * migrations (#2714).
+ *
+ * The companion `migrate-pg.test.ts` runs with
+ * `skip: MANAGED_AUTH_MIGRATIONS` because Better Auth's `user`,
+ * `session`, and `organization` tables don't exist in the test
+ * fixture. That leaves the migrations in `MANAGED_AUTH_MIGRATIONS`
+ * (0027 / 0042 / 0048 / 0050 / 0061 / 0090) without any real-Postgres
+ * coverage — a regression that breaks one of those files would only
+ * surface on the next production deploy.
+ *
+ * This suite bootstraps a minimal Better Auth schema first (`user`,
+ * `session`, `organization` with the columns Atlas migrations
+ * reference, plus enough primary-key / FK structure that the
+ * Better-Auth-dependent ALTERs apply cleanly), then runs the FULL
+ * Atlas migration set with no skip. Catches planner-level errors that
+ * mock-pool tests can't see — same shape as the existing smoke, just
+ * with the auth-table fixture in place.
+ *
+ * Hand-crafted DDL vs invoking Better Auth's own migrator: the test
+ * focuses on Atlas migrations, not Better Auth's. Keeping the DDL
+ * minimal means a Better Auth schema-shape change doesn't break this
+ * test unless it actually intersects an Atlas migration. The set of
+ * columns below comes from greping the six MANAGED migrations for
+ * Better-Auth table references — anything new an Atlas migration adds
+ * against `user` / `session` / `organization` should be exercised here
+ * (or land in `MANAGED_AUTH_MIGRATIONS` with a stub update).
+ */
+
+import { afterAll, beforeAll, describe, expect, it } from "bun:test";
+import { Pool } from "pg";
+import { runMigrations } from "@atlas/api/lib/db/migrate";
+
+const TEST_DB_URL = process.env.TEST_DATABASE_URL;
+const describeIfPg = TEST_DB_URL ? describe : describe.skip;
+const PG_TEST_TIMEOUT_MS = 30_000;
+
+/**
+ * Minimal Better Auth bootstrap. Only the columns Atlas migrations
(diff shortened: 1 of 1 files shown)


msywulak:
@codex review

msywulak:
@codex review

Issue #2811, opened 2026-05-25
Track: file upstream bun bug — mock.module() not intercepting under --isolate (1.5.4 follow-up)


msywulak:
> **Updated diagnosis (was: mock.module() not intercepting under --isolate).** The original framing was wrong. Probe fixtures in `packages/api/src/__tests__/_bun-isolation-experiment/pair-{6,8,9,10,11}.experiment.ts` ruled out the `mock.module()` and env hypotheses and pinned the actual root cause. Original write-up preserved in the edit history.

## The bug

Under bun 1.3.14 `--isolate` (and therefore `--parallel`, which implies it), loading a module that uses top-level `await` does NOT wait for the await chain to settle before exposing the module's bindings. Importers reading those exports get:

```
ReferenceError: Cannot access 'X' before initialization.
```

Affects every import pattern: top-level `await import`, `await import` inside `beforeAll`, static `import` (hoisted), `await import` inside a test body. There is **no in-test workaround**.

## Minimal repro — zero Atlas deps (~15 lines)

Lives at `packages/api/src/__tests__/_bun-isolation-experiment/pair-11-tla-bare.experiment.ts` + `pair-11-tla-target.ts`:

```typescript
// target.ts
await new Promise<void>((resolve) => setTimeout(resolve, 50));
export const ready = "DONE";
```

```typescript
// test.ts
import { test, expect } from "bun:test";
const mod = await import("./target");
test("ready is DONE", () => expect(mod.ready).toBe("DONE"));
```

| Command | Result |
|---|---|
| `bun test ./test.ts` | ✅ passes |
| `bun test --isolate ./test.ts` | ❌ `ReferenceError: Cannot access 'ready' before initialization` |
| `bun test --parallel ./test.ts` | ❌ same |

## Version bisect

| bun | bare | --isolate | --parallel |
|---|:---:|:---:|:---:|
| 1.3.11 | ✅ | ✅ | ✅ |
| 1.3.13 | ✅ | ✅ | ✅ |
| 1.3.14 | ✅ | ❌ TDZ | ❌ TDZ |

Single-version regression at 1.3.14. The `bun-v1.3.13..bun-v1.3.14` commit range against pair-11 will pinpoint the offending change.

## Likely cause (circumstantial)

The bun v1.3.14 release ships TLA/TDZ-related ESM fixes (from the [blog](https://bun.com/blog/bun-v1.3.14)):

[...]

Issue #2802, opened 2026-05-25
Cutover: bun test --parallel + delete custom isolated runner (1.5.4 slice 6)


msywulak:
Parent: #2796

The win condition. Once slices 1-5 land and `bun test --parallel` runs clean against the full api suite + others, this slice is the mechanical cutover: swap all test scripts and delete the custom runner.

This is essentially the diff that PR #2794 originally attempted before we learned the custom runner was load-bearing. With slices 1-5 done, that coupling is gone and the cutover is safe.

## Key files

- `packages/{api,mcp,cli,web,react}/package.json` + `ee/package.json` — `"test": "bun run scripts/test-isolated.ts"` → `"test": "bun test --parallel"` (6 files)
- `.github/workflows/ci.yml:325` — `bun run scripts/test-isolated.ts --shard N/4` → `bun test --parallel --shard=N/4`
- Delete: `packages/{api,mcp,cli,web,react}/scripts/test-isolated.ts` + `ee/scripts/test-isolated.ts` (6 files, ~1200 LoC)
- `packages/api/package.json` — drop the now-redundant `test:fast` script
- `CLAUDE.md` + `AGENTS.md` "Testing" section — replace `bun run test, never bun test` rule with `bun test --parallel, never bare bun test`. Replace the `--affected` / `--since` docs with `--changed=origin/main` / `--changed=HEAD~3`
- `.claude/commands/{ci,next}.md` — same substitutions
- `apps/docs/content/docs/contributing/ci.mdx` — same substitutions
- Test-file comments referencing `scripts/test-isolated.ts` get a quick read-through (~2 files)

## Acceptance criteria

- [ ] All 6 package test scripts swapped
- [ ] `ci.yml` `api-tests` matrix uses `bun test --parallel --shard=N/4`
- [ ] All 6 `scripts/test-isolated.ts` files deleted
- [ ] CLAUDE.md / AGENTS.md / skill docs / contributor docs updated
- [ ] CI green end-to-end including all 4 `api-tests` shards
- [ ] `scripts/check-test-discipline.sh` allowlist is empty (or only documented exceptions remain)
- [ ] Wall-clock comparison in PR description: full api suite time before vs. after

## Dependencies

[...]

PR #2824, opened 2026-05-25
test(api): #2811 — pin bun to 1.3.13 to work around --isolate TLA regression

### LINKED ISSUE #2812: Refactor ~120 api route tests off mock.module() to Effect Layers (1.5.4 slice 6 alternative path)
The big refactor that would let us actually delete the custom subprocess test runner. Alternative path for closing out milestone 1.5.4 (#2796) if the bun upstream bug (#2811) doesn't get fixed.

## Context

Slice 6 (#2802) intended to swap the 6 `scripts/test-isolated.ts` subprocess runners for native `bun test --parallel`. Empirically blocked: under `bun test --isolate` (which `--parallel` implies), top-level `mock.module()` calls do NOT intercept the SUT's import resolution. See #2811 for the upstream-bug filing.

The api route test suite uses this pattern everywhere — ~120 files in `packages/api/src/api/__tests__/` mock middleware / handlers / services via top-level `mock.module()` and then top-level `await import("@atlas/api/server")`. Under `--isolate` the mocks don't apply; under the custom runner each file is a fresh subprocess so the mock timing works.

The bun fix may or may not land. This issue is the parallel path: refactor the test files off `mock.module()` to a pattern that works under `--isolate` regardless of bun's fix.

## The replacement pattern

Slices 3 (#2799) and 4 (#2800) established the Effect Layer-based DI pattern that already lives in `packages/api/src/__test-utils__/layers.ts`. Examples that already work:

- `createConnectionTestLayer()` — fresh `ConnectionRegistry` per test, no global mutation
- `createPluginRegistryTestLayer()` — fresh `PluginRegistry` per test
- `TestAppLayer` / `TestAdminLayer` / `TestPlatformLayer` — composed Layer.provide that wires fake services into the Hono bridge
- `runHandler(c, label, async () => { ... })` — already context-aware via the Effect bridge

For a route test, the refactor is roughly:

```typescript
// BEFORE (mock.module pattern — breaks under --isolate):
mock.module("@atlas/api/lib/auth/middleware", () => ({
  authenticateRequest: mockAuth,
}));
const app = (await import("@atlas/api/server")).default;

// AFTER (Layer-based DI — works under --isolate):
import { runTest, TestAppLayer } from "@atlas/api/__test-utils__/layers";
const TestAuth = Layer.succeed(AuthService, { authenticate: mockAuth });
const layer = TestAuth.pipe(Layer.provideMerge(TestAppLayer));
await runTest(app.fetch(...), layer);
```

Some routes still call legacy non-Effect helpers via `mock.module()` (e.g. `authenticateRequest` lives in `lib/auth/middleware.ts`, not behind a Tag). Those need a small adapter Tag in `lib/effect/services.ts` first, then the test moves to `Layer.succeed`.

## Key files

- `packages/api/src/api/__tests__/` — ~120 test files using the `mock.module()` + top-level-`await import()` pattern
- `packages/api/src/__test-utils__/layers.ts` — destination pattern; extend with auth/middleware/rate-limit Tags as needed
- `packages/api/src/lib/effect/services.ts` — add new Tags for currently-non-Effect helpers (authenticateRequest, checkRateLimit, getClientIP, etc.) that route tests need to swap
- `packages/api/src/lib/auth/middleware.ts` — wrap exports behind a Tag so tests can `Layer.succeed` them

## Acceptance criteria

- [ ] Audit `packages/api/src/api/__tests__/` and produce a precise file count + per-file dependency list (which `mock.module()` targets each file uses)
- [ ] Introduce Tags for `authenticateRequest`, `checkRateLimit`, `getClientIP`, and any other non-Effect helpers consumed by route tests
- [ ] Migrate route tests in batches, one route-family at a time (actions, admin-orgs, conversations, etc.) — each batch is its own PR
- [ ] After all batches: `bun test --parallel src/api/` reports 0 failures
- [ ] Then do the actual #2802 cutover: swap 6 `package.json` test scripts to `bun test --parallel`, delete the 6 `scripts/test-isolated.ts` files, update CLAUDE.md / AGENTS.md
- [ ] Include before/after wall-clock comparison for the full api suite in the cutover commit body

## Dependencies

- Sibling: #2811 (upstream bun bug filing) — if bun fixes the bug first, this work becomes unnecessary
- Blocks: #2802 (the cutover itself)
- Built on: #2799 (PluginRegistryTestLayer), #2800 (connections._reset patterns) — same DI strategy

## Estimated scope

~120 files, mechanical but tedious. Likely 5–10 PRs of 10–20 files each. Each PR runs `bun test --parallel src/api/<route-family>/` to verify before merging.

## Source

Discovered during slice 5b/6 work on branch `claude/practical-hamilton-Ycwto`. Single-file empirical (`actions.test.ts`: 62 pass → 5 pass / 57 fail) makes the blocker concrete. Full verdict on #2801.

msywulak:
## Summary
- Pin `engines.bun` to `>=1.3.13 <1.3.14` (workspace root + `create-atlas` + `create-atlas-plugin`) to work around a top-level-await regression in bun 1.3.14's `--isolate` mode.
- Adds investigative fixtures under `packages/api/src/__tests__/_bun-isolation-experiment/pair-{6,8,9,10,11}` that narrowed the failure and produced a minimal upstream-bug repro.
- Rewrites the experiment dir README to reflect actual findings.

## What broke

Under bun 1.3.14 `--isolate` (and `--parallel`, which implies it), modules with top-level `await` don't get fully awaited before their bindings are exposed to importers. Importers get `ReferenceError: Cannot access 'X' before initialization`. Affects every import shape (top-level `await import`, `await import` inside `beforeAll`, static `import`, in-test await import) — there is no in-test workaround.

`pair-11-tla-bare.experiment.ts` is the minimized repro: ~15 lines, zero Atlas deps. Single-version regression at 1.3.14 (works on 1.3.11 and 1.3.13). Full diagnosis + version bisect + changelog quotes on #2811.

## Why the engine pin

Cheaper than the alternatives:

1. **Engine pin (this PR)** — ~5 lines, restores `bun test --parallel` correctness today. Unpin when upstream lands a fix.
2. Wait for upstream — pair-11 is filable today; #2811 acceptance criteria tracks that.
3. Refactor SUTs off top-level await — far more invasive; was the premise of #2812 (now closed as not-needed).

## Validation

With 1.3.13, `packages/api/` runs clean under `bun test --parallel`:
```
3020 pass
0 fail
8344 expect() calls
Ran 3020 tests across 120 files. [19.58s]
```

The 1.5.4 slice 6 cutover (#2802) is unblocked under this pin and lands in a follow-up PR.

## Why these fixtures

The original #2811 framing blamed `mock.module()` and proposed a 120-file Effect-Layer refactor (#2812). Probe fixtures 6, 8, 9 ruled out the `mock.module()` and env-propagation hypotheses; pair-10 surfaced the TLA timing; pair-11 minimized it. Kept in-tree as evidence and as the upstream-bug repro source. All use the `.experiment.ts` suffix so the regular `**/*.test.ts` glob skips them.

## Follow-ups (not in this PR)

- #2802 slice 6 cutover: swap `bun run test` scripts to `bun test --parallel`, delete `scripts/test-isolated.ts` × 6 copies — separate PR.
- File the upstream bun issue using pair-11 as the repro body — see #2811 acceptance criteria.
- Optional `bun-v1.3.13..bun-v1.3.14` git-bisect to pinpoint the offending bun commit (prompt drafted in the thread that produced this PR).

## Issue references

- Resolves the diagnostic / unblock portion of #2811
- Closes #2812 (premise wrong; closed in the issue tracker, not this PR)
- Unblocks #2802

## Test plan
- [x] `bun test --parallel ./src/api/` in `packages/api/` on bun 1.3.13: 3020/3020 pass
- [x] `bun test --isolate ./src/__tests__/_bun-isolation-experiment/pair-11-tla-bare.experiment.ts`: passes on 1.3.13, fails on 1.3.14 (regression verified both ways)
- [ ] CI runs against the pinned bun version

🤖 Generated with [Claude Code](https://claude.com/claude-code)


msywulak:
Thanks @chatgpt-codex-connector — that was a real and important miss, not a nit. The `engines.bun` field is advisory only (npm/bun warn but still install/run), so without changing the workflow `BUN_VERSION` the pin accomplishes nothing for the actual failure this PR is meant to fix.

Pushed `de8d475d` updating six workflow files from `1.3.14` → `1.3.13`:

- `ci.yml`
- `eval.yml`
- `deploy-validation.yml`
- `lighthouse.yml`
- `sync-starters.yml`
- `publish.yml` (the inline `bun-version`, not via env)

### Intentionally out of scope (flagged for follow-up)

Eight Dockerfiles still base on `oven/bun:1.3.14@sha256:e10577f...`:

- `deploy/api/Dockerfile`, `deploy/web/Dockerfile`, `deploy/docs/Dockerfile`
- `packages/sandbox-sidecar/Dockerfile`
- `examples/docker/Dockerfile`
- `create-atlas/templates/docker/Dockerfile{,.sidecar}`, `create-atlas/templates/docker/sidecar/Dockerfile`

The TLA regression (#2811) is in `bun test --isolate`, not in the standard runtime path that production uses. Production won't hit it. Pinning the Dockerfiles to 1.3.13 for coherence is worth doing (and would silence the `engines` warning at `bun install` time inside the image) but isn't load-bearing for this fix. Holding for a separate PR — happy to roll it in here if reviewers prefer.

Issue #2727, opened 2026-05-23
Slice 1: Demo lead → Twenty person (fire-and-forget)


msywulak:
## Parent

#2726

## What to build

Push every demo signup on useatlas.dev into Twenty CRM at crm.useatlas.dev as a `Person` record, stamped with first-touch and last-touch source attribution. This is the foundational tracer bullet — it builds the *minimum* of every deep module needed for the integration, but only wires the demo capture path.

End-to-end behavior:
1. Prospect submits the `/demo` gate on useatlas.dev
2. Atlas writes to `demo_leads` (existing behavior, unchanged)
3. After insert, `captureDemoLead()` yields the new `SaasCrm` Tag and calls `upsertLead({source: "demo", email, ip, userAgent})`
4. On SaaS (enterprise enabled), the real `SaasCrmLayer` invokes the Twenty plugin's `upsertPerson` action via `TwentyClient`, matching on `emails.primaryEmail`
5. On self-hosted (no enterprise), the Noop layer returns success without dispatching
6. If Twenty is unreachable: log a warning, do not throw, do not block the demo response. Lead is lost (upgraded to durable retry in the outbox slice)

**Multi-source attribution.** Two columns on the Twenty `Person` object capture funnel position:
- `atlasFirstSource` — sticky, set once on first dispatch for that email, never overwritten thereafter
- `atlasLastSource` — updated on every dispatch with the current event's source

Implementation lives in `TwentyClient.upsertPerson` (the only place that touches Twenty's record state). `LeadNormalizer` stays pure — it outputs the event's source under a single `eventSource` field, and `upsertPerson` translates that into first/last on write. The standard approach: GET person by email → if exists and `atlasFirstSource` is set, send `{ atlasLastSource }` only; otherwise send `{ atlasFirstSource, atlasLastSource }`. Implementer may swap for an equivalent atomic Twenty REST primitive if one exists.

Modules ships in this slice:
- `plugins/twenty/` scaffold via `definePlugin()`: `TwentyClient` with `upsertPerson` only, `TwentyCredentialResolver` (env-var path only — DB-row precedence is the admin-UI slice), config schema with `apiKey` marked `secret: true`
[...]

PR #2830, opened 2026-05-26
feat(api): #2729 — durable lead outbox + Scheduler-backed flusher

### LINKED ISSUE #2729: Slice 2: Durable lead outbox (replaces fire-and-forget)
## Parent

#2726

## What to build

Replace the fire-and-forget Twenty dispatch from #2727 with a durable outbox. Lead events are written to `crm_outbox` synchronously, then a background flusher processes them with retry-on-failure, exponential backoff, and **sub-step idempotency** — so a partial dispatch (upsertPerson succeeded, createNote crashed) does not create duplicate Notes on retry.

End-to-end behavior:
1. Caller invokes `SaasCrm.upsertLead(event)`
2. `LeadOutbox.enqueue(event)` inserts a `crm_outbox` row with `status="pending"` and returns immediately
3. The Scheduler-backed flusher polls pending rows, marks them `in_flight`, normalizes via `LeadNormalizer`, and dispatches sub-steps in order
4. **Sub-step idempotency**: after `upsertPerson` succeeds, the returned `personId` is persisted on the row (`twenty_person_id`). After `createNote` succeeds, the returned `noteId` is persisted (`twenty_note_id`). On replay, sub-steps with their resource ID already recorded are skipped — only the remaining sub-steps run
5. On success (all sub-steps done): row marked `done`, `processed_at` set
6. On transient failure (5xx, network error): row remains `pending`, `attempts++`, `last_error` recorded, exponential backoff applied to next attempt
7. On permanent failure (4xx non-retryable status, or max attempts exceeded): row marked `dead`, `log.warn` emitted with row id + last_error
8. On graceful shutdown: flusher resets `in_flight` rows back to `pending` so they're picked up on restart

Schema: `crm_outbox` (id, event_type text, payload jsonb, status enum (`pending` | `in_flight` | `done` | `dead`), attempts int, last_error text, **twenty_person_id text nullable**, **twenty_note_id text nullable**, created_at, processed_at). Internal infra table — not user-surfaced content, so the mode system does not apply.

## Acceptance criteria

- [ ] `SaasCrm.upsertLead` enqueues to `crm_outbox` instead of dispatching directly to Twenty
- [ ] Inducing a Twenty 5xx (test fixture) causes `attempts` to increment with exponential backoff delay between attempts
- [ ] After max attempts, the row transitions to `dead` and a `log.warn` is emitted with the row id + last_error
- [ ] **Sub-step idempotency**: simulating an `upsertPerson` success + `createNote` failure → retry reuses the recorded `twenty_person_id` and only re-runs `createNote`. Verified by asserting `upsertPerson` is called exactly once across retries when the personId column is populated
- [ ] **Sub-step idempotency on full success replay**: if a row is marked `in_flight` with both resource IDs already populated (simulating a crash after both sub-steps succeeded but before `status="done"`), the next flush moves it to `done` without making any Twenty calls
- [ ] The demo lead path from #2727 works end-to-end through the outbox (no regression — demo signup → Person in Twenty)
- [ ] Restarting the API does not drop pending rows — `in_flight` rows are reset to `pending` on startup and the flusher resumes them
- [ ] `LeadOutbox` integration tests against `TEST_DATABASE_URL` cover: enqueue, flush-success, flush-failure-retry, max-attempts-dead, restart-resumes-in-flight, **sub-step replay with stored resource IDs**. Follow the `migrate-pg.test.ts` shape (skip when env var unset)
- [ ] `crm_outbox` migration has a matching `pgTable` definition in `schema.ts`; `check-schema-drift.sh` passes
- [ ] `migrate-pg.test.ts` runs cleanly with the new migration
- [ ] Flusher honors graceful shutdown — no in-flight row is left in `in_flight` state on clean stop
- [ ] Manual smoke: temporarily break the Twenty URL, fire a demo signup, restore the URL, observe the row goes from `pending` → `done` after the next flusher tick

## Blocked by

- #2727

msywulak:
## Summary

Slice 2 of 1.6.0. Replaces the fire-and-forget `SaasCrm.upsertLead` dispatch from slice 1 (#2727 / PR #2785) with a durable `crm_outbox` table and a Scheduler-backed flusher. A Twenty outage, API crash, or partial sub-step failure no longer drops the lead.

- **Migration `0102_crm_outbox.sql`** — `id, event_type, payload jsonb, status text+CHECK, attempts, last_error, twenty_person_id, twenty_note_id, created_at, processed_at` + partial index on `(status, created_at) WHERE status IN ('pending','in_flight')` so the flusher poll stays fast as done/dead rows accumulate. Drizzle mirror added in the same PR; schema-drift gate is green.
- **`packages/api/src/lib/lead-outbox/`** — new deep module owning generic queue mechanics (`enqueue`, `flushBatch` via single-statement `UPDATE … FOR UPDATE SKIP LOCKED RETURNING`, `recoverInFlight`, deterministic per-tier backoff: 30s / 2m / 8m / 30m / 2h, dead-letter at attempt 6). Dispatcher is parameterised — Twenty-specific dispatch lives in `ee/src/saas-crm/` so the `core → ee` inversion stays enforced by `check-ee-imports.sh`.
- **Scheduler wiring (`lib/effect/layers.ts:makeSchedulerLive`)** — yields `SaasCrm`, runs startup recovery (UPDATE `in_flight` → `pending`) **before** the tick starts, forks an `Effect.repeat(Schedule.spaced)` poll fiber, and registers a finalizer that interrupts the fiber **and** runs a final recovery sweep so SIGTERM never strands rows. Tick interval is `ATLAS_CRM_OUTBOX_TICK_SECONDS` (default 5s); batch limit is 50.
- **Idempotent sub-step replay** — dispatcher sees the row's existing `twenty_person_id` / `twenty_note_id` and skips any sub-step whose ID is already populated. After each successful sub-step the dispatcher's `persist` callbacks UPDATE the column immediately. The "upsertPerson succeeded, createNote crashed before commit" path is now safe; the next flush goes straight to createNote.
- **Permanent vs transient classification** — 4xx other than 429 → dead immediately; 5xx / 429 / transport / unknown → retry with backoff. Dispatcher contract violation (a thrown error) is treated as transient so a dispatcher bug doesn't dead-letter every row.
- **Demo path unchanged** — `captureDemoLead` still calls `crm.upsertLead`; that call now enqueues instead of dispatching, but the catch-and-log around it stays for defense-in-depth.

## Test plan

- [x] `bun run lint` — clean.
- [x] `bun run type` — clean.
- [x] `bun run test` — 474/474 api test files + all other packages pass (one pre-existing flake in `ai-saas.test.ts` cleared on retry; untouched by this PR).
- [x] Schema-drift, template-drift, railway-watch, oauth-helper, security-headers, test-discipline, syncpack — all green.
- [x] Real-PG integration tests (`outbox-pg.test.ts`) pass against a real Postgres: 5xx → retry / 401 → dead / partial-success replay (upsertPerson called exactly once) / row with both IDs straight to done / startup recovery flips `in_flight` → `pending` / backoff filters recently-failed rows.
- [x] Full migration smoke (`migrate-pg.test.ts`) — all 103 migrations apply cleanly end-to-end.
- [x] `ee-imports` gate — `lib/effect/enterprise-layer.ts` remains the only `@atlas/ee` import in core.

Closes #2729

### DIFF
--- a/ee/src/saas-crm/__tests__/saas-crm.test.ts
+++ b/ee/src/saas-crm/__tests__/saas-crm.test.ts
@@ -1,10 +1,13 @@
 /**
- * SaasCrm layer tests — boot verification + dispatch + end-to-end Live wiring.
+ * SaasCrm layer tests — boot verification + outbox enqueue + dispatcher.
  *
- * Exercises the EE-side Layer (`SaasCrmLive`) with a mocked fetch impl,
- * isolating from both the plugin's HTTP wrapper (covered in
- * `plugins/twenty/__tests__/`) and from Twenty itself. There are NO
- * live calls to crm.useatlas.dev.
+ * Exercises the EE-side Layer (`SaasCrmLive`) with a mocked fetch impl
+ * AND a stubbed `internalQuery` (so the outbox enqueue is observable
+ * without a real Postgres). The dispatcher itself (`dispatchOutboxRow`)
+ * is unit-tested directly — it's the call the flusher makes from
+ * inside the Scheduler Layer.
+ *
+ * There are NO live calls to crm.useatlas.dev.
  */
 
 import { describe, test, expect, beforeEach, mock } from "bun:test";
@@ -25,8 +28,30 @@ mock.module("@atlas/api/lib/logger", () => ({
   }),
 }));
 
+// ── Stub the internal DB so SaasCrmLive can observe enqueue ─────────
+let internalDbAvailable = true;
+let lastEnqueueArgs: { sql: string; params: unknown[] } | null = null;
+let enqueueCount = 0;
+let nextEnqueueId = "row-id-0";
+mock.module("@atlas/api/lib/db/internal", () => ({
+  hasInternalDB: () => internalDbAvailable,
+  internalQuery: async (sql: string, params?: unknown[]) => {
+    lastEnqueueArgs = { sql, params: params ?? [] };
+    if (/^\s*INSERT INTO crm_outbox/i.test(sql)) {
+      enqueueCount++;
+      return [{ id: nextEnqueueId }];
+    }
+    return [];
+  },
+}));
+
 // ── Now we can import the layer + its helpers ───────────────────────
-const { verifyCustomFields, dispatchLead, SaasCrmLive } = await import("../index");
+const {
+  verifyCustomFields,
+  SaasCrmLive,
+  dispatchOutboxRow,
+  classifyTwentyError,
+} = await import("../index");
 const { SaasCrm } = await import("@atlas/api/lib/effect/services");
 
 // ── Fixture helpers ─────────────────────────────────────────────────
@@ -60,11 +85,19 @@ function withFetch<T>(impl: typeof globalThis.fetch, run: () => Promise<T>): Pro
   });
 }
 
+function resetStubs(): void {
+  internalDbAvailable = true;
+  lastEnqueueArgs = null;
+  enqueueCount = 0;
+  nextEnqueueId = "row-id-0";
+}
+
 // ── verifyCustomFields ──────────────────────────────────────────────
 
 describe("verifyCustomFields", () => {
   beforeEach(() => {
     enterpriseEnabled = true;
+    resetStubs();
   });
 
   test("returns ok=true when both required custom fields are present", async () => {
@@ -93,39 +126,9 @@ describe("verifyCustomFields", () => {
     });
   });
 
-  test("returns ok=false when atlasLastSource is missing", async () => {
-    const fetchImpl = (async () =>
-      metadataResponse(["id", "atlasFirstSource"])) as unknown as typeof globalThis.fetch;
-
-    await withFetch(fetchImpl, async () => {
-      const result = await verifyCustomFields({
-        apiKey: "k",
-        baseUrl: "https://crm.test.local",
-      });
-      expect(result).toEqual({ ok: false });
-    });
-  });
-
-  test("returns ok='transient' on 5xx (transient upstream failure)", async () => {
-    const fetchImpl = (async () =>
-      new Response(JSON.stringify({ messages: ["oops"] }), {
-        status: 503,
-        headers: { "Content-Type": "application/json" },
-      })) as unknown as typeof globalThis.fetch;
-
-    await withFetch(fetchImpl, async () => {
-      const result = await verifyCustomFields({
-        apiKey: "k",
-        baseUrl: "https://crm.test.local",
-      });
-      if (result.ok !== "transient") throw new Error("expected transient");
-      expect(result.reason.length).toBeGreaterThan(0);
-    });
-  });
-
-  test("returns ok='transient' on network failure", async () => {
+  test("returns ok=transient on network failure", async () => {
     const fetchImpl = (async () => {
-      throw new Error("ECONNREFUSED");
+      throw new Error("ECONNRESET");
     }) as unknown as typeof globalThis.fetch;
 
     await withFetch(fetchImpl, async () => {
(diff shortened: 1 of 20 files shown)


msywulak:
## Review + Codex findings addressed in d65b8152

### Review (HIGH/MEDIUM, all fixed)
- **H1** `classifyHttpStatus` now routes 408 + 425 → `transient` (was silently dead-lettering retryable transport failures). `classify.test.ts` covers both.
- **H2** Retry-After plumbed end-to-end: `TwentyClientError.retryAfterMs` parsed via RFC-9110 parser; `DispatchOutcome.transient` carries `retryAfterMs` + `httpStatus`; new `retry_after TIMESTAMPTZ` column; `CLAIM_SQL` WHERE uses `COALESCE(retry_after, created_at + tier)` so the upstream's requested delay always wins. Unit + PG tests pin the gate.
- **M1** `dispatchOutboxRow` splits the upsertPerson + persist call into separate `try` blocks — a pg blip is now labelled `persist.setTwentyPersonId failed after upsertPerson succeeded`, not as a Twenty failure. New saas-crm test pins the label.
- **M2** Sales-form rows return `permanent` with an operator-visible message (was `log.debug` + silent `done`). Test pins the dead-letter path.
- **M3** `SaasCrmShape` is now a discriminated union — `available: true` narrows `dispatcher` to non-null. The 5 hand-synchronised return sites in `SaasCrmLive` collapse to two shapes; the Tag.available rule violation is gone.
- **M4** Stale comments cleaned across `backoff.ts`, `classify.ts`, `outbox.ts`, `0102_crm_outbox.sql`, `ee/src/saas-crm/index.ts`.
- **M5** Missing tests added:
  - real-PG concurrent claim race (`FOR UPDATE SKIP LOCKED` correctness)
  - real-PG long-backoff non-block (fresh row B is claimed even when row A is gated)
  - shutdown finalizer (new `outbox-flusher-lifecycle.test.ts` — boot + dispose both hit `recoverInFlight`)
  - dispatcher returns 2xx-with-no-id → permanent
  - end-to-end restart (recover → flush → done with zero Twenty calls)
  - AC-3 dead-letter log assertion (rowId + last_error in structured fields)
  - persist-failure label test
  - Retry-After PG gate test
  - CLAIM_DELAY_SQL ↔ DELAYS_MS PG lockstep test

### Polish (LOW, all fixed)
- **L1** Terminal-status UPDATEs wrapped in single-retry helper (`markStatusWithRetry`) so an isolated pg connection blip doesn't strand the row until next pod boot.
- **L2** Operator query for stranded-pending poison-pill rows added to `RECOVER_IN_FLIGHT_SQL` comment.
- **L3** PG test pins CLAIM_DELAY_SQL EPOCH lockstep with `DELAYS_MS` at every tier.
- **L4** `outboxTick` logs `FlushResult` per non-empty tick (`lead_outbox.tick_complete`).
- **L5** `TODO(#2729-followup)` annotations on `twenty_person_id` / `twenty_note_id` calling out the vendor-name leak (generic `resource_ids JSONB` once a second CRM ships).

### Codex bot findings (all fixed)
- **P1** (multi-pod recovery race): added `claimed_at TIMESTAMPTZ` stamped by `CLAIM_SQL`; `recoverInFlight` is now age-gated via `staleAgeMs` (startup: 5 min; shutdown: 30 s) so a sibling pod's actively-dispatched in_flight row is **not** reset by another pod's shutdown sweep. Prevents double-dispatch under rolling deploys.
- **P2** (timer overflow): `getTickIntervalMs` clamps `ATLAS_CRM_OUTBOX_TICK_SECONDS` to `[1, 3600]`. A typo'd `3600000` (operator meant ms) would have overflowed bun's 32-bit timer to 1ms and hammered Postgres + Twenty.
- **P3** (exhausted in_flight rows): `recoverInFlight` now dead-letters rows with `attempts >= DEAD_AFTER_ATTEMPTS` instead of resetting them to `pending` (where the claim WHERE filters them out and they hide forever). Returns `RecoveryResult { reset, deadLettered }`.

### Gates
`lint`, `type`, `schema-drift`, `ee-imports`, `test-discipline`, `security-headers`, `railway-watch`, `oauth-helper`, `template-drift`, `syncpack` — all pass. Full `bun run test` passes modulo two known parallel-pressure flakes (`ai-saas`, `scim-provenance-enforcement`) that pass cleanly in isolation, matching the slice-1 (#2785) exception.

PR #2844, opened 2026-05-26
feat(api): #2734 — crm_outbox depth gauges + threshold warn

### LINKED ISSUE #2734: Slice 8: Outbox depth metrics + threshold alert
## Parent

#2726

## What to build

Surface `crm_outbox` health as a first-class observability signal so a sustained Twenty outage is visible before it becomes a backlog problem. Without this, pending rows accumulate silently and operators learn about the issue only when marketing-ops complains that leads are missing.

End-to-end behavior:
1. The existing Telemetry / OTel path exposes two gauges:
   - `atlas.crm_outbox.pending_count` — current rows with `status = "pending"`
   - `atlas.crm_outbox.dead_count` — current rows with `status = "dead"`
2. Both gauges are updated by the outbox flusher on each polling tick (no separate polling job needed)
3. When `pending_count` crosses a configurable threshold (default 100), `log.warn` emits with the current depth and the oldest pending row's `created_at`
4. When any row transitions to `dead`, `log.warn` emits with row id, event_type, attempts, last_error (already in #2729's acceptance — this slice just verifies the dead-count gauge reflects it)
5. The threshold is overridable via `ATLAS_CRM_OUTBOX_WARN_THRESHOLD` env var

## Acceptance criteria

- [ ] `atlas.crm_outbox.pending_count` and `atlas.crm_outbox.dead_count` gauges are exposed via the existing Telemetry path and visible in the metrics endpoint
- [ ] Both gauges update on each flusher tick — no separate counting job
- [ ] Inducing a synthetic backlog of 101+ pending rows triggers a single `log.warn` per minute (rate-limited to avoid log spam), reporting current depth + oldest pending row's age
- [ ] `ATLAS_CRM_OUTBOX_WARN_THRESHOLD` env var overrides the default 100
- [ ] No new dependencies — uses the existing Telemetry Tag and Scheduler Layer
- [ ] Unit test: gauge values reflect the underlying row counts under various status distributions
- [ ] `bun run type`, `bun run lint`, `bun run test` all pass

## Blocked by

- #2729

msywulak:
## Summary

- `atlas.crm_outbox.pending_count` and `atlas.crm_outbox.dead_count` synchronous OTel gauges, recorded by the slice-2 flusher (#2729) on every tick BEFORE dispatch — so the value an operator sees between ticks reflects queue depth at tick start, not residual depth after the dispatch drains some rows.
- Rate-limited `log.warn` (`event: lead_outbox.depth_threshold_warn`) once per 60s per pod when pending depth crosses `ATLAS_CRM_OUTBOX_WARN_THRESHOLD` (default 100). Payload carries `depth`, `threshold`, `oldestPendingCreatedAt` (ISO), and `oldestPendingAgeMs` so a dashboard can plot backlog age without parsing log strings.
- Per-tick logic extracted to `runOutboxTick(deps)` in `lib/lead-outbox/tick.ts` — pure async function the unit test drives with stub DB / dispatcher / gauges / logger, so the slice's contract is enforced at the type level instead of via `mock.module`. `layers.ts` wires production deps and allocates one `OutboxWarnRateLimiter` per scheduler Layer scope.

## Acceptance criteria — issue #2734

- [x] `atlas.crm_outbox.pending_count` + `.dead_count` gauges exposed via the existing Telemetry path (`@atlas/api` meter, identical to the abuse/MCP counters)
- [x] Both gauges update on each flusher tick — no separate counting job (one aggregate `SELECT ... FROM crm_outbox` with `FILTER (WHERE status = ...)` clauses, single round-trip per tick)
- [x] 101+ pending rows triggers one `log.warn` per minute with depth + oldest pending row's age (verified by `runOutboxTick` test — two within-window ticks emit exactly one warn)
- [x] `ATLAS_CRM_OUTBOX_WARN_THRESHOLD` env var overrides the default 100 (with clamp-and-warn discipline matching `getTickIntervalMs`)
- [x] No new dependencies — uses the existing OTel meter in `lib/metrics.ts` and the Scheduler Layer in `lib/effect/layers.ts`
- [x] Unit test: gauge values reflect underlying row counts across PG-shaped result variants (string counts, Date / ISO oldest, null oldest, missing row)
- [x] `bun run lint`, `bun run type`, `bun run test` all pass

## Notes

- Snapshot order: snapshot SELECT runs **before** dispatch. Reversing would make the gauge under-report during a sustained backlog (the dispatch drains some rows, the snapshot sees the residual, depth-warn never trips even when the queue is growing tick-over-tick).
- Rate-limit is time-based, not edge-based: a 101+ depth that drops to 0 and re-rises within 60s does not re-warn (gate is `now - lastWarnAt < intervalMs`); past the window, it re-warns.
- The lifecycle test (`outbox-flusher-lifecycle.test.ts`) still passes — its `WHERE status = 'in_flight'` regex correctly ignores the new `FILTER (WHERE status = 'pending')` snapshot SQL.
- Self-hosted is unaffected: the `NoopSaasCrm` layer makes `saasCrm.available === false`, so the flusher fiber never forks and no observations land on the gauges.

## Test plan

- [x] `bun run lint` — 0 warnings
- [x] `bun run type` — 0 errors
- [x] `bun run test` — full suite, 220+ files, 0 failures
- [x] `bun x syncpack lint` — no issues
- [x] All drift gates: template / security-headers / railway-watch / schema / oauth-helper / test-discipline — green
- [x] `bun test packages/api/src/lib/lead-outbox/__tests__/depth.test.ts` — 17/17 pass (5 snapshot, 8 rate-limiter, 4 orchestrator)
- [ ] Post-merge: confirm `atlas.crm_outbox.pending_count` lights up on Grafana for `api`/`api-eu`/`api-apac` after next dispatch tick

Closes #2734.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/2844"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782401162&installation_id=135337507&pr_number=2844&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F2844&signature=a6de93472d645d9d31b0a9a7ec970db7ce9cefaec948f440d6b8ed40b1fdce2b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -241,6 +241,7 @@ ATLAS_ADMIN_EMAIL=admin@useatlas.dev
 # === Twenty CRM (SaaS-side lead capture — #2727) ===
 # TWENTY_BASE_URL=https://crm.useatlas.dev   # Twenty REST base URL (defaults to crm.useatlas.dev). Self-hosters set to their own Twenty hostname.
 # TWENTY_API_KEY=                            # Bearer key from Twenty Settings → API & Webhooks. Required for SaaS demo / signup / contact-form dispatch into Twenty; unset = SaasCrm.available is false. Twenty Person object must have custom fields `atlasFirstSource` AND `atlasLastSource` (verified at boot).
+# ATLAS_CRM_OUTBOX_WARN_THRESHOLD=100        # Pending-row threshold for the rate-limited backlog warning (#2734). When the flusher sees more than this many `crm_outbox` rows in `pending` status, it emits one `log.warn` per minute with the depth + oldest pending row's age. Pair with the `atlas.crm_outbox.pending_count` gauge for dashboards. Out-of-range values clamp to [1, 1_000_000] and log a warn.
 
 # === Talk-to-sales form (#2730) ===
 # Server side (api): verifies Cloudflare Turnstile tokens submitted by the
--- a/packages/api/src/lib/effect/layers.ts
+++ b/packages/api/src/lib/effect/layers.ts
@@ -53,14 +53,20 @@ import { EnterpriseLayer, type EnterpriseSubsystem } from "./enterprise-layer";
 import { AuditPurgeScheduler, SaasCrm } from "./services";
 import {
   recoverInFlight as recoverOutboxInFlight,
-  flushBatch as flushOutboxBatch,
+  runOutboxTick,
   getTickIntervalMs as getOutboxTickIntervalMs,
+  getWarnThreshold as getOutboxWarnThreshold,
+  OutboxWarnRateLimiter,
   FLUSH_BATCH_LIMIT as OUTBOX_FLUSH_BATCH_LIMIT,
   STARTUP_RECOVERY_STALE_MS as OUTBOX_STARTUP_STALE_MS,
   SHUTDOWN_RECOVERY_STALE_MS as OUTBOX_SHUTDOWN_STALE_MS,
   type OutboxDB,
   type RecoveryResult as OutboxRecoveryResult,
 } from "@atlas/api/lib/lead-outbox";
+import {
+  crmOutboxPendingCount,
+  crmOutboxDeadCount,
+} from "@atlas/api/lib/metrics";
 
 const log = createLogger("effect:layers");
 
@@ -1480,12 +1486,24 @@ export function makeSchedulerLive(
         );
 
         const outboxTickIntervalMs = getOutboxTickIntervalMs();
+        // One rate limiter per Layer scope — `lastWarnAt` lives on the
+        // instance, so a sustained 101+ pending depth fires exactly
+        // one log.warn per minute regardless of how many ticks elapse.
+        const outboxWarnLimiter = new OutboxWarnRateLimiter(getOutboxWarnThreshold());
         const outboxTick = Effect.tryPromise({
           try: () =>
-            flushOutboxBatch(outboxDb, outboxDispatcher, OUTBOX_FLUSH_BATCH_LIMIT),
+            runOutboxTick({
+              db: outboxDb,
+              dispatcher: outboxDispatcher,
+              batchLimit: OUTBOX_FLUSH_BATCH_LIMIT,
+              limiter: outboxWarnLimiter,
+              pendingGauge: crmOutboxPendingCount,
+              deadGauge: crmOutboxDeadCount,
+              logger: log,
+            }),
           catch: (err) => (err instanceof Error ? err : new Error(String(err))),
         }).pipe(
-          Effect.tap((result) =>
+          Effect.tap(({ flush: result }) =>
             Effect.sync(() => {
               // Only log non-empty ticks so an idle queue doesn't fill
               // the log with `claimed: 0`. The per-row dead-letter /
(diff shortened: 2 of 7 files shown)


msywulak:
## Review pass — Codex + 5 specialist agents

Addressed in `094b5a72`:

### Codex P2s
- **depth.ts:49** — Snapshot now scopes `WHERE status IN ('pending', 'dead')` so per-tick cost does not scale with unbounded `done` history. Pending side hits the slice-2 partial index `idx_crm_outbox_pending_created`; dead side falls back to a small sequential scan over terminal rows.
- **depth.ts:66** — Empty-aggregate-row path now **throws** instead of returning zeros + log.warn. The slice-2 `lead_outbox.tick_failed` envelope catches it, OTel gauges retain last-recorded values, and a sticky pool failure can no longer make the queue look healthy.

### silent-failure-hunter
- `parseCount` / `parseTimestamp` fallbacks now emit structured warns (`lead_outbox.snapshot_count_unparseable` / `_timestamp_unparseable`) so a pg type-parser drift is audible.
- `getWarnThreshold` rejects `parseInt`-prefix typos (`"100abc"`, `" 100"`, `"50.5"`) via `/^-?\d+$/` strict regex + a `lead_outbox.threshold_unparseable` warn — operator fat-finger no longer silently accepts the truncation.

### comment-analyzer
- Fixed `tick.ts:6` wrong test-file path (`__tests__/tick.test.ts` → `__tests__/depth.test.ts`).
- Fixed `OutboxWarnRateLimiter` JSDoc — the "re-rises immediately re-warns" example contradicted the time-based gate. Now reads as time-based, re-emission only after `intervalMs` elapses.

### pr-test-analyzer
- Renamed the strict-`<` boundary test + comment that lied about the assertion. Test now reads "strict-less-than window: -1ms suppresses, exact boundary and +1ms re-emit" with a fresh limiter for the +1ms check.
- Added 10 tests (27 total / 68 expect calls / all green):
  - 5× `getWarnThreshold` env-var parser (AC #3 was previously untested)
  - `oldestPendingAgeMs` field assertion in the threshold warn test (AC #2)
  - `runOutboxTick` propagates snapshot errors → `tick_failed`
  - `NaN`/`Infinity` count clamp
  - non-integer string count clamp
  - unparseable `oldest_pending_at` string

### Skipped (with rationale)
- `clampEnvInt` helper extraction (type-design-analyzer): refactor-after-N=3 per their own rec, currently N=2.
- Brand types on `pending`/`dead` (type-design): locality is tight, field-name access, no swap risk at call sites.
- Adding a gauges-record-before-claim ordering check (pr-test #3): the SQL-call-order assertion already covers the load-bearing invariant; adding event-log instrumentation would couple the test to the orchestrator's internal sequencing.
- `tickId` correlation ID (silent-failure 4b): future-debuggability nice-to-have, no present defect.

All gates green: `bun run lint` / `type` / 41 affected tests / 7 drift gates.

PR #2872, opened 2026-05-27
fix(api): per-email serialization in crm_outbox CLAIM_SQL (1.6.0 hotfix)

### LINKED ISSUE #2870: bug(api): crm_outbox flusher doesn't serialize per-email — attribution swap on demo→signup pairs
## What

The CRM outbox flusher (`packages/api/src/lib/lead-outbox/tick.ts` + `ee/src/saas-crm/index.ts`) claims and dispatches rows without serializing per-email. When two events for the same email land in the outbox close together, they can apply to Twenty in arbitrary order — flipping `atlasFirstSource` / `atlasLastSource` (the stickiness contract C10 was designed to test) and leaking last-write-wins fields like `atlasIp` to whichever dispatch happens to finish last.

Surfaced during the #2866 closeout verification of `atlas ops smoke-crm` against `api.twenty.com` ([comment](https://github.com/AtlasDevHQ/atlas/issues/2866#issuecomment-4550277666)).

## Reproduction

Reproducible across both Run A (happy path on `main`) and Run C (filter inversion):

**C10 stickiness pair — `gworth@globexcorp.com`** (Run C):
```
- gworth@globexcorp.com :: atlasFirstSource: expected="DEMO", observed="SIGNUP"
- gworth@globexcorp.com :: atlasLastSource: expected="SIGNUP", observed="DEMO"
```
Fixture enqueues `demo` then `signup` for the same email. Expected: first=DEMO, last=SIGNUP. Observed: SWAPPED — signup completed before demo despite being enqueued second.

**B8 idempotency pair — `msdyson@cyberdynesys.com`** (both runs):
```
- msdyson@cyberdynesys.com :: atlasIp: expected="203.0.113.41", observed="203.0.113.40"
```
Fixture enqueues two demo events with different IPs (`.40` then `.41`). Observed `atlasIp` is the first event's IP, indicating the first dispatch completed last (or the dispatcher doesn't overwrite IP on repeat events — needs investigation).

## Why this matters

A real user demo'ing and then signing up within the same flusher tick window can have their attribution flipped — exactly the regression the C10 stickiness pair was designed to catch in the fixture.

The smoke surfaces this as exit 3 on `main`, which means `atlas ops smoke-crm` can't currently serve its stated purpose ("run before any CRM PR") — every clean run is dirty until this is fixed.

## Key files

- `packages/api/src/lib/lead-outbox/outbox.ts` — `claimNextBatch` / dispatch loop.
- `packages/api/src/lib/lead-outbox/tick.ts` — tick orchestration.
- `ee/src/saas-crm/index.ts` — the dispatcher Tag-bound by `SaasCrmLive`.
- `scripts/test-fixtures/crm-personas.yml` — the C10 + B8 fixtures that surface this.

## Possible shape

1. **Per-email serialization at claim time** — the claim SQL adds a `WHERE NOT EXISTS (SELECT 1 FROM crm_outbox o2 WHERE o2.email_key = crm_outbox.email_key AND o2.status = 'in_flight')` predicate so two rows for the same email never run concurrently. Needs an `email_key` computed column or index.
2. **Per-email merge-PATCH that's order-insensitive** — dispatcher reads the row's `event_at`, refuses to apply if a later-event row for the same email already landed (last-event-wins by timestamp, not by dispatch completion order).
3. **Smoke-side**: drop `atlasIp` from the diff for repeat-source events (B8 case). This is the cheaper change but doesn't address the underlying C10 attribution risk.

(1) or (2) fix the real bug; (3) is band-aid.

## Acceptance criteria

- [ ] `bun run atlas -- ops smoke-crm --personas scripts/test-fixtures/crm-personas.yml --wipe-twenty` against `api.twenty.com` exits 0 on `main` (with whatever fixture/code adjustment is appropriate).
- [ ] A unit test reproduces the gworth swap deterministically (e.g. by injecting a delay between two same-email dispatches in a mocked flusher) and asserts the post-fix invariant.
- [ ] The fix preserves throughput — no per-email serialization that would make a workspace with 1000 leads-per-minute back up.

## Dependencies

- #2869 (the smoke command itself — landed, used to detect this).
- #2729 (the outbox flusher).

## Out of scope

- The #2865 regression-guard mock test (separate follow-up).
- Wiring smoke into CI (parked on CI-only Twenty workspace, per #2866).

msywulak:
## Summary

- **Migration `0104_crm_outbox_email_key.sql`**: adds `email_key TEXT` to `crm_outbox`, backfills from `payload->>'email'`, indexes `(email_key, status, created_at) WHERE status IN ('pending', 'in_flight')`.
- **`enqueue`**: populates `email_key` via `extractEmailKey(payload)` — lowercase + trim, matching the lead normalizer; returns `null` for payloads without `email`.
- **`CLAIM_SQL`**: rewritten as a CTE. Inner SELECT excludes any row whose `email_key` already has an `in_flight` sibling; `DISTINCT ON (COALESCE(email_key, id::text))` then picks the earliest pending row per email per tick. NULL email_key rows fall back to per-id dedup so non-email-keyed future event types stay independent.
- **Drizzle schema**: mirrors the new column + index (drift check passes).
- **Coverage**: 6 new in-memory unit tests + 3 new real-Postgres tests covering the C10 gworth swap, B8 msdyson idempotency, NULL independence, throughput preservation, and the cross-tick NOT EXISTS gate.

Closes #2870.

## Why this fix (vs. options 2 / 3 in the issue body)

- **Option 1 (this PR)** — fixes the bug at the durable layer (DB-side claim), works across pods, no Twenty schema changes, throughput preserved for distinct emails. Worst case for a workspace: same-email events drain one-per-tick (5s default).
- **Option 2 (order-insensitive merge-PATCH)** — would require a new `atlasLastEventAt` custom field on Twenty Person + dispatcher changes to read existing state and compare timestamps. Heavier; introduces an external schema dependency.
- **Option 3 (smoke band-aid)** — doesn't fix the underlying ordering risk; rejected per the issue body ("(1) or (2) fix the real bug; (3) is band-aid").

## Test plan

- [x] `bun run lint`
- [x] `bun run type` 
- [x] `bash scripts/check-schema-drift.sh`
- [x] `bash scripts/check-test-discipline.sh`
- [x] `bash scripts/check-ee-imports.sh`
- [x] `bun run syncpack lint --dependency-types prod,dev`
- [x] In-memory: `bun test src/lib/lead-outbox/__tests__/outbox.test.ts` — 31/31 pass (was 25, +6 for #2870)
- [x] Real-PG (via `TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5432/atlas`): `bun test src/lib/lead-outbox/__tests__/outbox-pg.test.ts` — 16/16 pass (was 13, +3 for #2870)
- [ ] `atlas ops smoke-crm` against `api.twenty.com` (DoD #1 from the issue — owner verification, needs the SaaS `TWENTY_API_KEY` which isn't in this worktree)
- [ ] CI green

## Notes

- The unit test for the gworth swap is deterministic: tick 1 claims only `demo`; tick 2 claims `signup`. The for-loop in `flushBatch` then guarantees `upsertPerson` for `demo` (POST: first=DEMO, last=DEMO) lands before `signup`'s PATCH (last=SIGNUP, first stays DEMO sticky). Source-swap is impossible post-fix.
- `migrate-pg.test.ts` will exercise 0104 against real Postgres in CI as part of the existing per-PR smoke.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/2872"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782437484&installation_id=135337507&pr_number=2872&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F2872&signature=1f66f36bf74f183635d9cb1e93f862f7503e98c49aec9ad79dd5cea453e442a3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/api/src/lib/auth/__tests__/migrate.test.ts
+++ b/packages/api/src/lib/auth/__tests__/migrate.test.ts
@@ -349,6 +349,7 @@ describe("migrateAuthTables", () => {
             { name: "0101_approval_surface_gchat.sql" },
             { name: "0102_crm_outbox.sql" },
             { name: "0103_converge_salesforce_pillar.sql" },
+            { name: "0104_crm_outbox_email_key.sql" },
           ],
         };
       }
--- a/packages/api/src/lib/db/__tests__/migrate.test.ts
+++ b/packages/api/src/lib/db/__tests__/migrate.test.ts
@@ -118,8 +118,9 @@ describe("runMigrations", () => {
     // + 0101 (add 'gchat' to approval-surface CHECK enums + promote
     //   gchat catalog row to 'available', #2754)
     // + 0102 (crm_outbox durable queue, #2729)
-    // + 0103 (converge salesforce row to ADR-0006 canonical pillar) = 104.
-    expect(count).toBe(104);
+    // + 0103 (converge salesforce row to ADR-0006 canonical pillar)
+    // + 0104 (crm_outbox.email_key for per-email serialization, #2870) = 105.
+    expect(count).toBe(105);
 
     // Advisory lock acquired before anything else
     expect(queries[0]).toContain("pg_advisory_lock");
@@ -252,6 +253,7 @@ describe("runMigrations", () => {
         "0101_approval_surface_gchat.sql",
         "0102_crm_outbox.sql",
         "0103_converge_salesforce_pillar.sql",
+        "0104_crm_outbox_email_key.sql",
       ],
     });
 
--- a/packages/api/src/lib/db/migrations/0104_crm_outbox_email_key.sql
+++ b/packages/api/src/lib/db/migrations/0104_crm_outbox_email_key.sql
@@ -37,19 +37,24 @@ ALTER TABLE crm_outbox ADD COLUMN IF NOT EXISTS email_key TEXT;
 -- backfill is idempotent (`WHERE email_key IS NULL`) so running this
 -- migration twice is a no-op.
 --
--- `NULLIF(LOWER(TRIM(...)), '')` mirrors the runtime `extractEmailKey`
--- helper exactly: a whitespace-only payload (`"   "`) passes a naive
--- pre-TRIM `<> ''` check but collapses to `''` after TRIM. We MUST
--- land NULL rather than `''` for those rows — a non-NULL empty-string
+-- Mirror the runtime `extractEmailKey` helper exactly:
+--   `NULLIF(LOWER(REGEXP_REPLACE(value, '^\s+|\s+$', '', 'g')), '')`
+-- A whitespace-only payload (`"   "`, `"\t\n"`, etc.) collapses to ''
+-- after the trim and we MUST land NULL — a non-NULL empty-string
 -- email_key would collide with every other whitespace-only row in
 -- CLAIM_SQL's NOT EXISTS gate, causing all such rows to mutually
--- block each other and drain at one-per-tick globally. `jsonb_typeof`
--- guards against `"email": null` and non-string types up front.
+-- block each other and drain at one-per-tick globally. The regex
+-- form matches JavaScript `String.prototype.trim()` for ASCII
+-- whitespace (space, tab, newline, CR, FF, VT); plain `TRIM()` in
+-- Postgres only strips spaces by default and would leak rows whose
+-- payload uses tabs/newlines, asymmetric with what `extractEmailKey`
+-- writes at runtime. `jsonb_typeof` guards against `"email": null`
+-- and non-string types up front.
 UPDATE crm_outbox
-SET email_key = NULLIF(LOWER(TRIM(payload->>'email')), '')
+SET email_key = NULLIF(LOWER(REGEXP_REPLACE(payload->>'email', '^\s+|\s+$', '', 'g')), '')
 WHERE email_key IS NULL
   AND jsonb_typeof(payload->'email') = 'string'
-  AND NULLIF(LOWER(TRIM(payload->>'email')), '') IS NOT NULL;
+  AND NULLIF(LOWER(REGEXP_REPLACE(payload->>'email', '^\s+|\s+$', '', 'g')), '') IS NOT NULL;
 
 -- Partial index supports both the DISTINCT-ON dedupe inside CLAIM_SQL
 -- and the NOT EXISTS gate. Filtering on the active statuses keeps the
--- a/packages/api/src/lib/db/migrations/0104_crm_outbox_email_key.sql
+++ b/packages/api/src/lib/db/migrations/0104_crm_outbox_email_key.sql
@@ -1,0 +1,52 @@
+-- 0104_crm_outbox_email_key.sql
+--
+-- Add `email_key` for per-email serialization at claim time (#2870).
+--
+-- Without this column, two `crm_outbox` rows for the same email can be
+-- claimed in the same flusher tick (or by different pods) and dispatch
+-- to Twenty in arbitrary order. That flips `atlasFirstSource` /
+-- `atlasLastSource` (the C10 stickiness contract) and leaks last-write-
+-- wins fields like `atlasIp` to whichever dispatch happens to finish
+-- last. CLAIM_SQL post-migration dedupes by `email_key` AND gates each
+-- claim on `NOT EXISTS (in_flight row for same email_key)` so:
+--
+--   1. Within a single tick batch, only the EARLIEST pending row per
+--      email is claimed. Siblings wait for the next tick after the
+--      claimed row reaches a terminal state.
+--   2. Across ticks / pods, a row for an email already in_flight is
+--      excluded from the candidate set (the partial index makes this
+--      check cheap as the table grows).
+--
+-- `email_key` is application-populated by `enqueue` rather than a
+-- `GENERATED ALWAYS AS` column so legacy payloads with a missing /
+-- malformed `email` field stay claimable (NULL email_key never matches
+-- another NULL in the NOT EXISTS clause, so those rows are treated as
+-- independent — matching today's behaviour for any future event types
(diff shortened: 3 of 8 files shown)


msywulak:
Pushed `fdafc84d` addressing all three Codex findings plus follow-ups from a parallel review (code-reviewer, silent-failure-hunter, pr-test-analyzer, comment-analyzer agents). Summary:

### Codex P1 #1 — cross-pod race ✅
Confirmed real. Two flushers under READ COMMITTED can each see the other's pre-commit UPDATE as invisible; both pass NOT EXISTS, both end up with different same-email rows in_flight. Fixed via transaction-scoped advisory lock: `AND (email_key IS NULL OR pg_try_advisory_xact_lock(2870, hashtext(email_key)))`. Uses the two-key variant with `2870` as namespace to avoid collisions with any other advisory locks. New PG test exercises two pods (separate `pg.Pool`s, size 1 each) running `flushBatch` in parallel with a slow dispatcher; sampling `COUNT(*) WHERE status='in_flight'` during the hold confirms only one pod claims.

### Codex P1 #2 — retry-cooldown leapfrog ✅
Confirmed real. R1 (older, in transient backoff) drops out of `claimable` via the due-time filter; the old NOT EXISTS only excluded `in_flight` siblings, so R2 (newer, fresh) leapfrogged → atlasFirstSource flips. Broadened NOT EXISTS:
```sql
AND NOT EXISTS (
  SELECT 1 FROM crm_outbox o2
  WHERE o2.email_key IS NOT NULL
    AND o2.email_key = crm_outbox.email_key
    AND o2.id <> crm_outbox.id
    AND o2.status IN ('pending', 'in_flight')
    AND o2.created_at < crm_outbox.created_at
)
```
The `created_at <` clause is essential — without it every row blocks itself. New PG test seeds R1 (pending, `retry_after = now() + 1 hour`) and R2 (fresh same email); asserts `claimed === 0`.

### Codex P2 — `backfill-crm-leads.ts` bypass ✅
Confirmed. `buildBulkEnqueueSql` was INSERTing 3 columns (no `email_key`). Now 4 columns; calls the now-exported `extractEmailKey({ email: event.email })` per row. The `extractEmailKey` JSDoc claim "exported for the migration backfill script" is now factual.

### Additional fixes from internal review
- **Migration 0104 whitespace asymmetry** — `payload->>'email' <> ''` passed `"   "` payloads which then collapsed to `''` via TRIM, giving a non-NULL empty-string `email_key` that mutually blocked every other whitespace-only row. Switched to `NULLIF(LOWER(TRIM(...)), '')` to mirror runtime `extractEmailKey` exactly. Added `jsonb_typeof(payload->'email') = 'string'` guard up front. New PG test pins it.
- **Warn-log on null email_key for known event types** — `enqueue` now `log.warn`s with `event: "lead_outbox.email_key_missing"` when a `demo`/`signup`/`sales-form`/`conversion` payload lands without an extractable email. Surfaces silent serialization-disabled rows.
- **`id` tie-breaker** added to every `ORDER BY created_at` so claim order is deterministic when timestamps tie (bulk INSERT scenario).
- **Comment accuracy** — corrected misleading "NOT EXISTS sees about-to-be in_flight" claim (it's DISTINCT ON that does intra-tick dedup); scoped CLAIM_SQL JSDoc to match impl; fixed `extractEmailKey` JSDoc normalization order.

### Local verification
- `bun run type` ✅
- `bun run lint` ✅
- `bash scripts/check-schema-drift.sh` ✅
- `bash scripts/check-test-discipline.sh` ✅
- `bash scripts/check-ee-imports.sh` ✅
- `bun x syncpack lint` ✅
- `cd packages/api && bun test src/lib/lead-outbox/__tests__/outbox.test.ts` → 33/33 (was 31, +2 warn-log tests)
- `bun run scripts/test-isolated.ts --affected` → 11/11 files pass (incl. `backfill-crm-leads.test.ts` after the 3-placeholder swap)
- Real-PG tests rely on `TEST_DATABASE_URL`; api-tests workflow exercises them in CI.

@codex review — would appreciate another pass to confirm the advisory-lock + broadened NOT EXISTS shape closes both P1 scenarios.

msywulak:
CI: ✅ all 24 checks green on `be9fe37e` (Codex round 2 fixes).

The previous failure on `fdafc84d` was the migration whitespace-backfill test I added, which caught a real `TRIM` semantics mismatch — Postgres default `TRIM()` strips spaces only, JS `String.trim()` strips the full whitespace set. Fixed in `24505820` by switching both sides (migration + test) to `REGEXP_REPLACE(value, '^\s+|\s+$', '', 'g')`. Round 2 then layered the three Codex follow-ups (same-`created_at` tie-break, in_flight-blocks-regardless-of-age, `stamp-conversion` literal) on top in `be9fe37e`.

Local verification on real Postgres: 36 in-memory + 19 real-PG outbox tests pass. CI: all 4 api-tests shards + Symlink Stub Build + Deploy Validation pass.

Ready for re-review when convenient.

PR #2888, opened 2026-05-28
fix(web): hard-nav after auth to dodge stale Router Cache → /login bounce

msywulak:
## Summary

After invitation signup → verify-email OTP → accept-invitation, the freshly-authenticated user was being bounced back to `/login`. Reproduced in Playwright against prod (PR for `#2885` was the symmetric companion fix; this is a different bug surfaced in dogfood after that landed).

**Root cause (not what I first thought):** the session cookie IS being set, Domain-scoped `.useatlas.dev`, sent on cross-subdomain requests, and proxy.ts DOES read it correctly. The bug is **Next.js Router Cache poisoning**:

1. User arrives at `/accept-invitation/<id>` while signed out → public-route branch renders.
2. Next.js auto-prefetches `/` (Link in the page chrome).
3. Prefetch hits `proxy.ts` → no session → **307 → /login**.
4. Next.js caches the prefetch result keyed by `/`.
5. User signs up, verifies OTP, clicks "Join Atlas" → server-side acceptance + set-active both succeed → session cookie is valid.
6. `router.push("/")` consults the cache, replays the **stale 307**, never hits the proxy again. User lands at `/login` with a valid session in the cookie jar.

This is a well-known Next.js gotcha (middleware redirect + Link prefetch + Router Cache). `Cache-Control: no-store` on the proxy redirect doesn't help — the Router Cache is in-memory client-side and ignores HTTP cache directives.

**Fix:** at every "auth state just changed" boundary, use `window.location.assign` instead of `router.push`. Hard nav drops the Router Cache entry for the destination, so the post-auth session is honored.

Centralized as `navigatePostAuth(url)` in `packages/web/src/lib/auth/post-auth-nav.ts` so the rationale lives in one place and tests have one mock point. Patched exits:

- `accept-invitation` handleAccept success
- `/signup` direct success (verification off) and post-OTP-verify
- `/login` email+password (non-2FA branch), passkey button, passkey autofill, post-OTP for email-unverified-on-sign-in
- `/login/two-factor` TOTP + backup code verify

The 2FA-redirect branch (`/login` → `/login/two-factor`) intentionally stays on `router.push` — no session exists yet there and the form needs SPA continuity for prefilled state. Documented inline.

## Test plan

- [x] `bun run lint` — 0/0
- [x] `bun run type` — 0
- [x] `bun run --filter '@atlas/web' test` — 166/166 files pass
- [x] login + 2FA page tests cover both the toHaveBeenCalledWith and not.toHaveBeenCalled assertions on `navigatePostAuthMock` (46/46)
- [x] All /ci drift gates green (lint, type, syncpack, template, security headers, railway watch, schema drift, oauth helper drift, test discipline, twenty resolver, published symbols, EE imports)
- [ ] Re-run the Playwright reproduction against this branch — bounce no longer fires

### Known pre-existing flake (unrelated)

Full `bun run test` flagged `packages/api/src/api/__tests__/admin-roles.test.ts` failing in the parallel-isolated runner. It passes 24/24 in isolation (`bun test src/api/__tests__/admin-roles.test.ts`). My changes are entirely in `packages/web/` so this cannot be related; main is currently green on this same test. Documenting per /ci's "note flake but don't ignore" rule.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/2888"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782521035&installation_id=135337507&pr_number=2888&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F2888&signature=863508d1a1f5c1aeb85be56b86ad7df08cb35b09da1dfff92f19c608a90a2b69"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/web/src/app/accept-invitation/[id]/page.tsx
+++ b/packages/web/src/app/accept-invitation/[id]/page.tsx
@@ -18,6 +18,7 @@ import { Button } from "@/components/ui/button";
 import { Alert, AlertDescription, AlertTitle } from "@/components/ui/alert";
 import { Database, Mail, AlertTriangle, LogOut } from "lucide-react";
 import { authClient, type OrgInvitationDetail } from "@/lib/auth/client";
+import { navigatePostAuth } from "@/lib/auth/post-auth-nav";
 
 interface PageProps {
   params: Promise<{ id: string }>;
@@ -121,7 +122,8 @@ export default function AcceptInvitationPage({ params }: PageProps) {
             );
           });
       }
-      router.push("/");
+      // Hard nav out of the auth boundary — see post-auth-nav for why.
+      navigatePostAuth("/");
     } catch (err) {
       setAcceptError(err instanceof Error ? err.message : "Failed to accept invitation.");
     } finally {
--- a/packages/web/src/app/login/page.test.tsx
+++ b/packages/web/src/app/login/page.test.tsx
@@ -44,12 +44,18 @@ mock.module("@/ui/hooks/use-webauthn-supported", () => ({
 }));
 
 const routerPushMock = mock((_path: string) => {});
+const navigatePostAuthMock = mock((_path: string) => {});
 // Per-test mutable search params so tests can drive the `?invitationId=…`
 // branch without re-mocking the whole module.
 const searchParamsStore: Record<string, string | null> = { invitationId: null };
 mock.module("next/navigation", () => ({
   useRouter: () => ({ push: routerPushMock, replace: () => {}, back: () => {} }),
   useSearchParams: () => ({ get: (k: string) => searchParamsStore[k] ?? null }),
+}));
+// Sign-in exits go through navigatePostAuth (hard nav) — keep one mock
+// point so tests don't have to spy on `window.location.assign` per-case.
+mock.module("@/lib/auth/post-auth-nav", () => ({
+  navigatePostAuth: navigatePostAuthMock,
 }));
 
 // `LoginPage` fetches the social-provider list + password-reset status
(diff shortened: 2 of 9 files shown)


msywulak:
Codex P2 (invited-2FA still bounces) addressed in `1ac12996`:

- Extracted the branch decision as `requiresRouterPushAfterSignIn(next)` next to `getPostSignInRoute`, prefix-matching `/login/two-factor` so `?callbackURL=…` is covered.
- Added the regression test in `post-sign-in-route.test.ts` (with a Codex-catch comment so the case stays anchored).
- Updated the page-side comment to explain *why* the 2FA branch must stay on `router.push` — no session cookie exists yet, and proxy.ts only treats exact-match `/login` as an auth route, so a hard nav would 307 back to `/login` and the challenge would be unreachable.

PR #2892, opened 2026-05-28
fix(api,web): admin semantic detail routes by storage key (#2891)

### LINKED ISSUE #2891: fix(api,web): admin semantic entity detail 404s on every entity whose YAML name differs from file stem
On `/admin/semantic`, clicking any entity in the file tree throws "Failed to load \"<Entity>\": Entity \"<Entity>\" not found." Reproduces on app.useatlas.dev/admin/semantic for every YAML in the production workspace (reported request ID `f015790c-c7fa-4848-9cb5-3caeb1923405`, group `g_prod`).

Root cause is a key mismatch between the list and detail responses on the `/api/v1/admin/semantic/entities/...` surface:

- **List** (`parseRowToAdminSummary`, `packages/api/src/lib/semantic/admin-source.ts:226`) returns `name: nameField ?? parsed.data.table` — i.e. the YAML's `name:` field (the *display* name), falling back to the `table:` field. The DB row's actual `name` column is **never returned** to the frontend.
- **Storage** writes the DB row's `name` column from the *file stem* or *table name*:
  - `_scanYamlDir` (`packages/api/src/lib/semantic/sync.ts:591`): `name = file.replace(/\.yml$/, "")` — file stem.
  - Wizard save (`packages/api/src/api/routes/wizard.ts:712`): `name: path.basename(e.tableName)` — table name.
- **Detail** (`getAdminEntity` → `getEntity`, `packages/api/src/lib/semantic/entities.ts:807`) queries `WHERE name = $3` against the storage column.

So for any entity where the YAML's `name:` differs from the file stem — which is the dominant pattern (`customers.yml` with `name: Customers`, `audit_log.yml` with `name: AuditLog`, etc.) — the frontend builds `?file=entities/Customers`, the detail handler runs `WHERE name = 'Customers'`, the row stored as `name = 'customers'` doesn't match, and the response is 404.

Repo confirms the pattern: every entity in `semantic/entities/*.yml` uses a CamelCase `name:` field over a snake_case `table:`. The user reports "really any yml in that semantic gives the error" — consistent with this being the bug for all production entities.

## Key files
- `packages/api/src/lib/semantic/admin-source.ts:178-240` — `parseRowToAdminSummary` projects DB row to summary; drops the storage `row.name` and returns display name only.
- `packages/api/src/lib/semantic/entities.ts:779-883` — `getEntity` queries `WHERE name = $3` against the storage column.
- `packages/api/src/api/routes/admin.ts:1635-1687` — `getEntityRoute` handler that returns the 404 the user sees.
- `packages/web/src/app/admin/semantic/page.tsx:524-546` — list response normalizer (only reads `e.name`, ignores any storage-key field).
- `packages/web/src/app/admin/semantic/page.tsx:637` — detail URL built from `selection.name` (the display name).
- `packages/web/src/app/admin/semantic/search-params.ts:31-38` — `fileParamToSelection` roundtrips display name through `?file=entities/<name>`.
- `packages/api/src/lib/semantic/sync.ts:589-608` — disk-import writes file-stem `name`.
- `packages/api/src/api/routes/wizard.ts:710-714` — wizard writes table-name `name`.

## Findings
- The architecture comment at `admin-source.ts:65-69` is aware that display name differs from table (\"Collapsing the two was the conflation bug…\"). But neither side surfaces the **storage** `name` to the frontend, so the URL routing key is silently the display name, which doesn't match the DB.
- The bug is not specific to AuditLog — it fires for every entity whose YAML `name:` is not byte-identical to the file stem/table. That's most of the repo's demo entities and likely most user data created via the wizard.
- The list+detail tests cover the parse path (`parseRowToAdminSummary`, `getAdminEntity`) but appear not to assert a roundtrip: list returns name X → detail by name X succeeds. Suite green despite the production break.
- Edit/delete uses the same surface (`/api/v1/admin/semantic/entities/edit/{name}` → `getEntity`), so anyone clicking Edit or Delete on an entity in SaaS also hits 404. Worth re-checking whether wizard-saved entities can ever be edited today.

## Suggested fix
Surface the storage key on the list response and have the frontend route URLs by it:

1. Add a `storageName: row.name` (or `key`) field to `AdminEntitySummary` (DB branch). For the disk branch, the storage key is the file stem.
2. Update the frontend's normalizer in `page.tsx:525` and the `SemanticSelection` shape to carry both `displayName` (for rendering) and `name` (for routing).
3. Roundtrip the storage key through `selectionToFileParam`/`fileParamToSelection`.

Alternative (more invasive but kills the duplication for good): backfill all rows so `semantic_entities.name = YAML.name` consistently, then update both write sites (`_scanYamlDir`, `wizard.ts:712`) to do the same. The list+detail key becomes the YAML `name:` everywhere and the URL is unambiguous.

Tactical hotfix-friendly version: in `parseRowToAdminSummary`, just return `name: row.name` (the storage key) and accept that the file tree will render snake_case until the deeper fix lands. Less disruptive but reverts the display-name affordance.

Add a list→detail roundtrip integration test before closing.

## Acceptance criteria
- [ ] Clicking any entity in `/admin/semantic` on SaaS loads its detail view (no 404).
- [ ] Display name in the file tree continues to use the YAML `name:` field where present (no visual regression).
- [ ] Integration test that asserts every name returned by `GET /admin/semantic/entities` resolves to a 200 on `GET /admin/semantic/entities/{name}` (covers both disk and DB sources, with and without a `connection_group_id`).
- [ ] Edit/Delete + raw-YAML view + version-history sub-routes all keep working for entities whose YAML name differs from the storage `name` column.
- [ ] Verified on `app.useatlas.dev/admin/semantic` post-deploy.

msywulak:
## Summary

- **Bug** ([#2891](https://github.com/AtlasDevHQ/atlas/issues/2891)): clicking any entity on `/admin/semantic` in production 404'd. The list returned the YAML `name:` (display) field but the detail handler queried `WHERE name = $1` against the storage column (the file stem / table name). For demo-data style YAMLs where `name: Customers` over `table: customers`, the URL routed by "Customers" while the DB row stored "customers" → 404.
- **Fix**: split the wire shape. `name` is the storage key; new `displayName` field carries the label the file tree renders. URLs route by `name` (matches the detail handler's lookup); the file tree continues to render the YAML display name via `displayName`.
- **Surfaces**: `parseRowToAdminSummary` returns `row.name`; `diskToAdminSummary` returns the file stem; `discoverEntities` populates both; `page.tsx` + `SemanticFileTree` consume both. Frontend falls back to `name` when `displayName` is absent so older mirrors keep rendering.

## Test plan

- [x] `bun run test` (full `packages/api` isolated suite — 487 files green)
- [x] `bun run test` in `packages/web` (166 files green)
- [x] `bun run lint`, `bun run type`
- [x] `scripts/check-schema-drift.sh`, `check-ee-imports.sh`, `check-template-drift.sh`, `check-test-discipline.sh`, `check-twenty-resolver-imports.sh`
- [x] New unit test `parseRowToAdminSummary: displayName is YAML name, name is storage key (#2891)` pins the bug.
- [x] New integration test `every name from the list endpoint resolves on the detail endpoint (#2891)` iterates the fixture list so any future drift between display + storage trips immediately.
- [x] New disk fixture `audit_log.yml` with `name: AuditLog` reproduces the production-bug shape; existing `companies` / `orders` fixtures keep passing.
- [ ] Verified on `app.useatlas.dev/admin/semantic` post-deploy — every entity in the file tree loads its detail on click.

Closes #2891.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/2892"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782565764&installation_id=135337507&pr_number=2892&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F2892&signature=b5e18b0f4e6f5f0e5f37b8dd4046bf802abe83be72e43f62208879936b4eb7f4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/packages/api/src/api/__tests__/admin.test.ts
+++ b/packages/api/src/api/__tests__/admin.test.ts
@@ -72,6 +72,23 @@ dimensions:
 `,
   );
 
+  // #2891 fixture — YAML's `name:` ("AuditLog") deliberately differs
+  // from the file stem ("audit_log") so list→detail roundtrip tests
+  // exercise the storage-key path. Pre-fix this was the production
+  // failure mode reported on app.useatlas.dev: clicking the entity
+  // hit `/api/v1/admin/semantic/entities/AuditLog` and 404'd because
+  // the DB row / file stem stored `audit_log`.
+  fs.writeFileSync(
+    path.join(tmpRoot, "entities", "audit_log.yml"),
+    `name: AuditLog
+table: audit_log
+description: Audit trail rows
+dimensions:
+  id:
+    type: integer
+`,
+  );
+
   fs.writeFileSync(
     path.join(tmpRoot, "metrics", "total_companies.yml"),
     `name: total_companies
@@ -756,8 +773,10 @@ describe("GET /api/v1/admin/overview", () => {
     // an org context the helper falls through to runtime `default` (which
     // exists in the test mock), so we still see 1.
     expect(body.connections).toBe(1);
-    // 2 disk entities (companies + warehouse/orders) via `listAdminEntities`.
-    expect(body.entities).toBe(2);
+    // 3 disk entities (companies + warehouse/orders + audit_log) via
+    // `listAdminEntities`. `audit_log` is the #2891 fixture whose YAML
+    // `name:` deliberately differs from its file stem.
+    expect(body.entities).toBe(3);
     expect(body.plugins).toBe(1);
     // Deployment-scaffold tiles (metrics, glossaryTerms, pluginHealth)
     // moved to /api/v1/platform/overview per #2489 — assert they're gone
@@ -879,7 +898,8 @@ describe("GET /api/v1/admin/semantic/entities", () => {
     expect(res.status).toBe(200);
 
     const body = (await res.json()) as { entities: Array<Record<string, unknown>> };
-    expect(body.entities.length).toBe(2);
+    // Count covers all three fixtures (companies, orders, audit_log).
+    expect(body.entities.length).toBe(3);
 
     const companies = body.entities.find((e) => e.table === "companies");
     expect(companies).toBeDefined();
@@ -892,6 +912,15 @@ describe("GET /api/v1/admin/semantic/entities", () => {
     expect(orders).toBeDefined();
     expect(orders!.source).toBe("warehouse");
     expect(orders!.connection).toBe("warehouse");
+
+    // #2891: list response must carry both the storage `name` (file
+    // stem) and the YAML's `displayName`. The frontend routes URLs by
+    // `name` (so detail lookup matches the DB / disk key), but
+    // renders the file tree off `displayName`.
+    const auditLog = body.entities.find((e) => e.table === "audit_log");
+    expect(auditLog).toBeDefined();
+    expect(auditLog!.name).toBe("audit_log");
+    expect(auditLog!.displayName).toBe("AuditLog");
   });
 });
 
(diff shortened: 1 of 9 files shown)


msywulak:
Addressed both Codex P2s in cb32bb6e:

1. **Per-source disk dedup** — disk `diskToAdminSummary` now lifts `e.source` into `connectionId` for non-default sources (`shopify/entities/orders.yml` + `stripe/entities/orders.yml` dedup on `(orders, shopify)` vs `(orders, stripe)` instead of colliding on `(orders, null)`). Source becomes the env badge on the file tree, which is also how multi-source semantic layouts read most cleanly. Disk arm of `AdminEntitySummary` widens `connectionId` from `null` to `string | null`.
2. **Drift drawer table lookup** — `DriftRoutingEntity` gains a required `table` field and `driftDrawerTargetFor` returns `match.table` instead of `selection.name`. The drawer's diff lookup is keyed by `table` server-side; routing by storage name surfaced "no drift detected" whenever the two diverged.

New unit tests pin both — admin-source roundtrip for two-source same-stem disk + a divergent `audit_log_v2` / `table: audit_log` drift-routing case.

The lighthouse failure on the previous commit was a transient `POST PR comment` GitHub API 404 right after PR creation — the run itself was `MOBILE_OUTCOME: success`. The re-queued run on cb32bb6e should pass.

PR #2964, opened 2026-05-29
fix(chat): linear-time feedback slash parser (js/polynomial-redos)

msywulak:
## Summary

CodeQL flagged `plugins/chat/src/proactive/feedback.ts:103` with **`js/polynomial-redos`** — *Polynomial regular expression used on uncontrolled data* (code scanning alert #26). The slash-command parser ran:

```ts
trimmed.match(/^feedback(?:\s*[:-]\s*|\s+)(.+)$/i)
```

The `\s*[:-]\s*` / `\s+` alternation has overlapping whitespace quantifiers, and the trailing `(.+)` also matches whitespace — an ambiguous boundary the engine can backtrack over super-linearly. The args come from the Chat SDK (`command="/atlas"`, `args="feedback …"`), i.e. library/user-controlled input, so it's a genuine ReDoS surface (slow on `feedback\t` + many tabs).

## Fix

Replace the single combined regex with a word-boundary keyword test plus single-quantifier, anchored strips:

```ts
if (!/^feedback\b/i.test(trimmed)) return { kind: "not-feedback" };
const body = trimmed
  .slice("feedback".length)
  .replace(/^\s+/, "")
  .replace(/^[:-]\s*/, "")
  .trim();
return { kind: "feedback", text: body };
```

Each pattern has one anchored quantifier with nothing that can fail after it, so matching is linear — no backtracking ambiguity. `\b` keeps `feedbackhello` from being treated as an invocation.

## Behavior

All existing `parseFeedbackSlashArgs` tests pass unchanged (`feedback <text>`, `feedback: <text>`, `feedback - <text>`, bare `feedback`, case-insensitive keyword, non-feedback args). Added regression tests:
- word-boundary rejection (`feedbackhello`, `feedback123` → not-feedback)
- a 100k-tab pathological input that now resolves in <100ms

```
13 pass / 0 fail
```

Closes code scanning alert #26.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/2964"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782610509&installation_id=135337507&pr_number=2964&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F2964&signature=396fa45cb0ae3ec2cc7d9a26e50481b0210d03a6e0fd15e5c466c9f178e310d0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/plugins/chat/src/proactive/__tests__/feedback.test.ts
+++ b/plugins/chat/src/proactive/__tests__/feedback.test.ts
@@ -73,19 +73,30 @@ describe("parseFeedbackSlashArgs", () => {
     });
   });
 
-  it("requires a separator after the keyword", () => {
-    // No word boundary after `feedback` — not a feedback invocation.
+  it("requires whitespace or a `:`/`-` separator after the keyword", () => {
+    // The keyword must abut end-of-string or one of {whitespace, :, -}.
+    // Trailing word chars or other punctuation are not invocations.
     expect(parseFeedbackSlashArgs("feedbackhello")).toEqual({ kind: "not-feedback" });
     expect(parseFeedbackSlashArgs("feedback123")).toEqual({ kind: "not-feedback" });
+    expect(parseFeedbackSlashArgs("feedback.note")).toEqual({ kind: "not-feedback" });
+    expect(parseFeedbackSlashArgs("feedback@x")).toEqual({ kind: "not-feedback" });
   });
 
-  it("matches a pathological whitespace-heavy input in linear time", () => {
-    // Regression for the polynomial-ReDoS shape (js/polynomial-redos):
-    // `feedback` + a long run of tabs must resolve instantly, not blow up.
-    const input = `feedback${"\t".repeat(100_000)}x`;
-    const start = performance.now();
-    expect(parseFeedbackSlashArgs(input)).toEqual({ kind: "feedback", text: "x" });
-    expect(performance.now() - start).toBeLessThan(100);
+  it("treats a bare separator (no body) as an empty feedback invocation", () => {
+    expect(parseFeedbackSlashArgs("feedback:")).toEqual({ kind: "feedback", text: "" });
+    expect(parseFeedbackSlashArgs("feedback -")).toEqual({ kind: "feedback", text: "" });
+  });
+
+  it("handles a huge whitespace-only body without choking", () => {
+    // The original `js/polynomial-redos` finding is on the regex *shape*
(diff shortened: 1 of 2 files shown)


msywulak:
## Review follow-up (commit `faaf232a`)

Ran a multi-agent review (code / tests / comments). Two material findings, both addressed:

**1. Separator alphabet was over-broadened.** The first cut used `/^feedback\b/i`, but `\b` fires before *any* non-word char — so `feedback.note`, `feedback@x`, etc. flipped from `not-feedback` to `feedback` with the punctuation leaking into the body. Replaced with a fixed-width lookahead `/^feedback(?=$|[\s:-])/i` that restricts acceptance to the **original** `{whitespace, ':', '-'}` separator set (or end-of-string), and stays linear. Added rejection tests for `feedback.note`/`feedback@x`.

**2. The wall-clock regression test was a false guard.** `parseFeedbackSlashArgs` calls `args.trim()` *before* the regex, which strips the trailing whitespace/newline run that triggers the polynomial blowup. Verified empirically: even the **old** vulnerable regex returned in <1ms for every whitespace-heavy input through the function — so a timing assertion is green on both old and new code and can never catch a regression. Dropped the `performance.now()` assertion; the ReDoS *shape* is guarded by CodeQL (which flags the regex directly), and the test now just asserts the strips collapse a large separator run to an empty body.

**Intentional, benign behavior change:** bare separators with no body now normalize to an empty feedback invocation — `feedback:` and `feedback -` → `{ kind: "feedback", text: "" }` (previously `not-feedback` / a stray-separator body). This matches the documented bare-`feedback` path and is the safe direction per the downstream contract (the feedback subcommand must never fall through to the SQL query flow). Pinned with tests.

14 parser tests pass; eslint clean.

Issue #3034, opened 2026-05-31
REST datasource install leaks the customer credential to third-party spec hosts (probeSpec)


msywulak:
## Summary

Surfaced during review of #3032 (notion-data candidate, slice 6b). When a built-in data candidate's `openapiUrl` points at a **public third-party host** (GitHub raw), the shared install path still attaches the customer's API credential as an `Authorization` header while fetching the spec — disclosing the secret to that host. This is **pre-existing slice-6a behavior** and already affects the merged **`stripe-data`** candidate, not just notion-data.

## Mechanism (verified)

- `data-candidate-form-handler.ts` install delegates to `persistOpenApiDatasourceInstall({ authKind: candidate.authKind, authValue: data.auth_value, openapiUrl: candidate.openapiUrl, … })` with `authKind: "bearer"`.
- `openapi-generic-form-handler.ts` builds `auth = buildResolvedAuth("bearer", token)` → `{ kind: "bearer", token }` and calls `probeSpec(openapiUrl, auth, …)`.
- `probe.ts` (`probeSpec`, ~L259-267) fetches `openapiUrl` with `headers: { Accept, ...authHeadersForProbe(auth) }`; for bearer that is `{ Authorization: "Bearer <token>" }`. **No host-match gate** — the credential is attached unconditionally.
- `stripe-data.openapiUrl` = `raw.githubusercontent.com/stripe/openapi/…`; `notion-data.openapiUrl` = `raw.githubusercontent.com/makenotion/notion-mcp-server/…`. Both leak the customer's secret to `raw.githubusercontent.com` on install and on every "Rediscover schema" re-probe.

The unconditional attach is intentional for the **Twenty** case, where the spec endpoint (`/open-api/core`) is on the same authenticated host as the API and legitimately needs the credential (see the design note in `probe.ts`). The flaw is that the same logic fires when the spec lives on an unrelated public host.

## Proposed direction (needs decision)

Only send probe auth when the spec URL host matches the resolved API base-URL host — OR make auth-on-probe a per-candidate / per-install opt-in (Twenty opts in; public-spec candidates don't). Either keeps Twenty working while never sending Stripe/Notion tokens to GitHub. Decision needed because of the Twenty same-host requirement.

## Key files
- `packages/api/src/lib/openapi/probe.ts` (`probeSpec`, `authHeadersForProbe`)
- `packages/api/src/lib/integrations/install/openapi-generic-form-handler.ts`
[...]

Issue #2978, opened 2026-05-29
v0.0.3: scheduler-driven auto re-discovery of OpenAPI specs on the configured interval


msywulak:
## Parent

#2868 — PRD: Generic OpenAPI agent datasource

## What to build

A periodic scheduler fiber — modeled on the scheduler-driven `byot-catalog-refresh.ts` and the `Effect.repeat(Schedule.spaced(...))` periodic-fiber pattern already in `layers.ts` — that walks OpenAPI datasource installs whose `spec_refresh_interval` is **due**, re-probes each spec endpoint, re-normalizes to an `OperationGraph`, updates the stored snapshot, and bumps the `cache_invalidated_at` watermark.

This is the automated sibling of the manual "Refresh now": the agent's view of a custom API stays current without an admin clicking a button. Composes with the drift diff (#2976) so each scheduled refresh records what changed. Egress note: the resolver already probes the spec server-side at resolve time, so scheduled probing is the same egress on a timer — respect the SaaS allowlist / per-tenant base URL.

## Acceptance criteria

- [ ] A scheduler task selects installs whose interval has elapsed since last check and re-discovers each (re-probe → snapshot update → watermark bump).
- [ ] `off` installs are never auto-refreshed; the global tick interval is env-configurable (mirrors the `getExpertSchedulerIntervalMs` style).
- [ ] A slow or failing upstream fails soft for that install (negative-cache / skip) without stalling the loop or other installs; a failed scheduled probe never overwrites/degrades the live snapshot.
- [ ] Per-install last-checked + outcome is observable (log/audit).
- [ ] Tests over: due-selection, off-skip, fail-soft on a down upstream, watermark bump on success.

## Blocked by

- #2926 (per-install record + re-discovery path)
- #2977 (the `spec_refresh_interval` setting this loop reads)

msywulak:
Shipped in #3046 — Tier-2 per-install scheduler-driven re-discovery (the automated sibling of the manual "Refresh now" #3002). The PR had no `Closes` linkage so it stayed open; closing on reconciliation. All acceptance criteria met (due-selection, off-skip, fail-soft on down upstream, watermark bump, observability). v0.0.3 is now 4/5 — only the breaking-change drift signal (#2979) remains.

Issue #2977, opened 2026-05-29
v0.0.2: per-install spec refresh interval — customer-configurable knob + Refresh now


msywulak:
## Parent

#2868 — PRD: Generic OpenAPI agent datasource

## What to build

Add a per-install, customer-configurable **spec refresh interval** to an OpenAPI datasource's `workspace_plugins.config` (e.g. `off` (default) / `daily` / `weekly` / a bounded custom interval). Expose it as a control on the per-install detail page with validation, plus an explicit "Refresh now" that reuses the manual re-discovery path.

No background scheduler in this slice — just the stored setting + UI + manual trigger, so the knob exists and is honored on demand before automation lands (the scheduler is a separate slice). Per-install only; the value is never shared cross-tenant — private specs stay per-workspace per #2970's isolation constraint.

## Acceptance criteria

- [ ] `workspace_plugins.config` carries a validated `spec_refresh_interval` (bounded enum or min/max-clamped interval); default `off`.
- [ ] The admin detail page has a control to set/clear the interval and a "Refresh now" button wired to manual re-discovery (#2926).
- [ ] The write path validates + clamps the value; invalid input is rejected with an actionable message (no silent fallback).
- [ ] The field honors the install record's content-mode / draft rules if it participates; it holds no secret, so no new encryption surface.
- [ ] Tests cover set / clear / clamp and that "Refresh now" triggers a re-probe.

## Blocked by

- #2926 (per-install record + detail page + manual rediscover)

Issue #2976, opened 2026-05-29
v0.0.2: OpenAPI spec drift diff — structured changeset on (re-)discovery


msywulak:
## Parent

#2868 — PRD: Generic OpenAPI agent datasource

## What to build

When an OpenAPI datasource's spec is (re-)probed — today via the manual "Rediscover schema" action (slice 2, #2926) — compare the freshly normalized `OperationGraph` against the previously stored snapshot and produce a **structured spec diff**: operations added / removed / changed, plus per-operation schema-field changes (new / removed / retyped columns, joins, query-pattern hints). Persist the diff alongside the install snapshot and surface it on the per-install detail page ("2 new operations, 1 removed, 3 changed fields since last refresh").

The diff is a pure deep module over two `OperationGraph`s — no per-API branching. It makes any refresh (manual now, scheduled later) legible: an admin sees exactly what moved upstream instead of a silent snapshot swap. Removed/retyped operations the agent relied on are the high-signal case.

## Acceptance criteria

- [ ] A pure `diffOperationGraphs(prev, next)` (or equivalent) returns a structured, deterministic, order-insensitive changeset: added / removed / changed operations + added / removed / retyped schema fields.
- [ ] Re-discovery persists the new snapshot **and** the computed diff against the install; a first-ever discovery records a baseline (no diff).
- [ ] The per-install detail page renders the diff as a human-readable summary; an unchanged re-probe shows "no changes".
- [ ] The diff covers the agent-relevant surface: operation set, `$ref` joins, inline/custom fields, query-pattern params.
- [ ] Unit tests over added / removed / changed / unchanged graphs, including the empty-baseline case.

## Blocked by

- #2926 (per-install record + snapshot + manual "Rediscover schema" must exist to diff against)

PR #3046, opened 2026-05-31
feat(openapi): scheduler-driven Tier-2 per-install spec re-discovery (#2978)

msywulak:
## What

Implements **#2978** (v0.0.3 — Spec Lifecycle): a Tier-2, per-install periodic scheduler that keeps each installed `openapi-generic` datasource's persisted spec snapshot current — the **automated sibling of the manual "Refresh now"** (#3002).

On a timer it walks every install whose per-install `spec_refresh_interval` (#2977) has elapsed since its last check, re-probes the spec endpoint, re-normalizes to an `OperationGraph`, updates the persisted snapshot, records the structured drift diff (#2976), and bumps a new `spec_last_checked_at` watermark.

### Scope boundary (Tier-2, not Tier-1)
- `scheduler/openapi-spec-refresh.ts` (#2970) is the **Tier-1 SHARED cross-workspace cache** refresh — conditional-GETs public catalog specs, **never mutates a workspace snapshot**. Left untouched.
- This is **Tier-2**: per-install, customer-configurable, **mutates the persisted per-install snapshot** + records a drift diff. New, orthogonal loop — exactly the arm the Tier-1 docstring names.

## How

- **Shared core** `lib/openapi/rediscover.ts` — `performRediscovery` (decrypt → auth → `probeSpec` → snapshot → diff, discriminated result) + `persistRediscoverySnapshot` + `stampSpecLastChecked`. The manual route is refactored to call it; its HTTP responses, persist SQL, cache eviction, and audit are **byte-for-byte unchanged** (route shrinks ~120 lines). Because re-discovery flows through the same `probeSpec`, scheduled probing inherits the **identical SSRF guard + redirect-revalidating fetch + #3034 host-match credential gate + SaaS egress allowlist** as resolve-time — sharing the path is what prevents divergence.
- **Due-selection** — the interval is *per-install*, so a single SQL `now() - interval` gate doesn't fit. A coarse SQL pre-filter selects non-`off` candidates (ordered least-recently-checked-first, batch-bounded); the precise "elapsed?" decision is `evaluateSpecRefreshDue` (new, in `spec-refresh.ts`) against `max(spec_last_checked_at, snapshot.probedAt)` — so a fresh install **and** a manual "Refresh now" both reset the clock without the route having to write the watermark.
- **Fail-soft** — a down/slow upstream (probe is `AbortSignal.timeout`-bounded) stamps the watermark (persisted negative-cache → not re-probed until its interval elapses again) and **leaves the live snapshot intact**; a persist failure deliberately does **not** stamp, so the next tick retries rather than caching a half-applied refresh. Per-install failures never abort the loop or the workspace's other installs.
- **Observability** — per-install `connection.probe` audit rows (`triggeredBy: "scheduler"`, so an operator filtering by install id sees manual + scheduled uniformly) + a per-tick `connection.spec_refresh_cycle` row (scheduler-stopped forensic invariant, mirrors BYOT). Global tick env-tunable via `ATLAS_OPENAPI_REDISCOVER_INTERVAL_HOURS` (default 1h, mirrors `getExpertSchedulerIntervalMs`).
- **Wiring** — start + finalizer-stop in the startup Layer DAG, alongside the Tier-1 fiber; `setInterval` + `unref()` + single-running + in-flight overlap guards, no-op without an internal DB.
- **No migration** — the watermark lives in `workspace_plugins.config` JSONB (selective-field pattern), like `spec_refresh_interval` and `openapi_last_diff`.

## Acceptance criteria
- [x] Selects installs whose interval has elapsed since last check and re-discovers each (re-probe → snapshot update → watermark bump)
- [x] `off` installs are never auto-refreshed; global tick interval is env-configurable
- [x] Slow/failing upstream fails soft for that install (negative-cache/skip) without stalling the loop or other installs; a failed scheduled probe never overwrites/degrades the live snapshot
- [x] Per-install last-checked + outcome observable (per-install + cycle audit rows, pino)
- [x] Tests over: due-selection, off-skip, fail-soft on a down upstream, watermark bump on success

## Tests
New: scheduler cycle + lifecycle (`openapi-install-rediscover.test.ts`), `evaluateSpecRefreshDue`/`parseIsoToMs` units, and the real watermark-write SQL helpers (`rediscover.test.ts`). Existing `admin-openapi-datasources.test.ts` stays green through the refactor.

`/ci`: lint, type, **test (api 537 files / 0 failed + others)**, syncpack, template/security/railway/schema/oauth drift, test-discipline, twenty-resolver, published-symbols — all pass. `main` CI + Railway staging green.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/3046"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782836475&installation_id=135337507&pr_number=3046&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F3046&signature=612600d3947c6513d6622ae5e1eb1e6df0a9c210fa69432ecbbe4f94d5260826"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Shared OpenAPI rediscovery core and a background scheduler to periodically rediscover per-install specs, persist snapshots/diffs, and emit audited cycle summaries.
  * New audit action for scheduler-driven spec refresh cycles.

* **Configuration**
  * New env option to control rediscovery interval in hours.

* **Documentation**
  * Example env file updated to document the new interval setting.

* **Tests**
  * Expanded unit and integration tests covering rediscovery, scheduling, and edge cases.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -310,6 +310,14 @@ ATLAS_ADMIN_EMAIL=admin@useatlas.dev
 # ATLAS_OPENAPI_SHARED_SPEC_TTL_MS=3600000    # Freshness window for the install-time short-circuit — within it a fresh install of an already-cached public upstream skips the network entirely (default: 1h)
 # ATLAS_OPENAPI_SPEC_REFRESH_INTERVAL_HOURS=24 # Cadence of the periodic conditional-GET refresh of the shared public-spec cache (mirrors the BYOT / semantic-expert scheduler; default: 24h)
 #
+# Per-install auto re-discovery (#2978, Tier-2): a separate scheduler re-probes each
+# installed generic OpenAPI datasource whose per-install spec_refresh_interval (set
+# in Admin → Connections: off / daily / weekly / "<N>h") has elapsed, refreshing its
+# persisted snapshot + drift diff. Distinct from the shared-cache refresh above — it
+# mutates each install's own snapshot. This tunes the LOOP wake cadence, not the
+# per-install interval; keep it at/below the smallest per-install interval you set.
+# ATLAS_OPENAPI_REDISCOVER_INTERVAL_HOURS=1   # How often the loop wakes to scan for due installs (default: 1h)
+#
 # SSRF egress guard (#3006): host-side spec probes AND operation execution are
 # blocked from reaching private / loopback / link-local / CGNAT IPs and internal
 # hostnames (e.g. cloud metadata at 169.254.169.254). The guard is ON in every
--- a/packages/api/src/api/routes/admin-openapi-datasources.ts
+++ b/packages/api/src/api/routes/admin-openapi-datasources.ts
@@ -23,45 +23,29 @@ import { createLogger } from "@atlas/api/lib/logger";
 import { errorMessage } from "@atlas/api/lib/audit/error-scrub";
 import { logAdminAction, ADMIN_ACTIONS } from "@atlas/api/lib/audit";
 import { internalQuery } from "@atlas/api/lib/db/internal";
-import { decryptSecretFields, parseConfigSchema } from "@atlas/api/lib/plugins/secrets";
 import { runHandler } from "@atlas/api/lib/effect/hono";
 import {
   OPENAPI_GENERIC_CATALOG_ID,
-  OPENAPI_GENERIC_CONFIG_SCHEMA,
   coerceRepresentationMode,
   isValidSnapshot,
-  type OpenApiSnapshot,
   type OpenApiAuthKind,
 } from "@atlas/api/lib/openapi/catalog";
 import {
-  resolveAuthFromDecryptedConfig,
-  probeSpec,
-  buildSnapshot,
   snapshotToGraph,
   invalidateInstallGraphCache,
   summarizeOperations,
-  OpenApiProbeError,
 } from "@atlas/api/lib/openapi/probe";
 import { REPRESENTATION_MODES } from "@atlas/api/lib/openapi/representation";
 import {
   coerceSpecRefreshInterval,
   normalizeSpecRefreshInterval,
 } from "@atlas/api/lib/openapi/spec-refresh";
-import { buildOperationGraph } from "@atlas/api/lib/openapi/spec";
-import {
-  diffOperationGraphs,
-  summarizeSpecDiffRecord,
-  baselineSpecDiffRecord,
-  unparseablePriorDiffRecord,
-  type SpecDiffRecord,
-} from "@atlas/api/lib/openapi/diff";
-import type { OperationGraph } from "@atlas/api/lib/openapi/types";
+import { performRediscovery, persistRediscoverySnapshot } from "@atlas/api/lib/openapi/rediscover";
+import { summarizeSpecDiffRecord } from "@atlas/api/lib/openapi/diff";
 import { ErrorSchema, AuthErrorSchema } from "./shared-schemas";
 import { createAdminRouter, requireOrgContext, requirePermission } from "./admin-router";
 
 const log = createLogger("admin.openapi-datasources");
-
-const SECRET_SCHEMA = parseConfigSchema(OPENAPI_GENERIC_CONFIG_SCHEMA);
 
 /**
  * Raw install row shape this router reads. A `type` (not `interface`) so it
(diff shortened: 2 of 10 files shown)


msywulak:
Thanks for the reviews 🙏 — addressed the two genuinely-right findings in 7e2edc5 and documented the two deliberate-by-design ones.

### Fixed
- **CodeRabbit — prototype-key hardening (`spec-refresh.ts`).** `lower in SPEC_REFRESH_PRESET_HOURS` matched inherited keys, so `normalizeSpecRefreshInterval("toString")` was *accepted as a valid interval* and `getSpecRefreshIntervalMs("constructor")` returned `NaN`. Switched both lookups to `Object.hasOwn`; added prototype-pollution tests (`toString`/`constructor`/`__proto__`/… → rejected on write, `null` on read, never due in the scheduler).
- **Codex — batch starvation (P2).** Real bug in the candidate query. `ORDER BY` now uses `GREATEST(config->>'spec_last_checked_at', config->'openapi_snapshot'->>'probedAt')` — the *same* effective-activity watermark `evaluateSpecRefreshDue` compares against — instead of the bare watermark `NULLS FIRST`. Otherwise a batch full of freshly-installed-but-not-due rows (null watermark, recent `probedAt`) would be re-selected every tick (returned `not_due`, never stamped) and starve genuinely-overdue installs past the `LIMIT`. Added a default-query SQL test.

### Deliberate by design (documented in-code, not changed)
- **Codex — skip disabled installs (P2).** The scheduler is intentionally `status != 'archived'` **only**, to match the datasource *resolver* (`workspace-datasource.ts`), which also doesn't gate on `enabled`. The scheduler must refresh exactly the set the agent is served — gating only the scheduler on `enabled` would leave a still-served install carrying a stale snapshot. If the resolver ever gains an `enabled` gate, the predicate moves in lockstep (noted in the query comment).
- **Codex — replica gating (P2).** No distributed lock, consistent with every sibling scheduler (BYOT / Tier-1 / semantic-expert). SaaS pins each regional API to `numReplicas: 1` (`deploy/README.md`: "intentional, not aspirational"), and re-discovery is idempotent (a duplicate re-probe writes the same snapshot + watermark), so the worst case under an accidental scale-up is duplicate egress/audit rows, not corruption. A cross-scheduler singleton is the right home for this if that invariant is ever lifted (noted in the lifecycle docstring).
- **CodeRabbit — partial `mock.module()` factories in the new test.** Kept as-is: the mock export sets are identical to the canonical sibling `byot-catalog-refresh.test.ts` (which passes CI), and the per-file subprocess runner (`test-isolated.ts`) isolates each test, so the "missing export breaks other files" failure mode the rule guards against can't occur here.

`/ci` re-run green: lint, type, affected tests (73 files, 0 failures) including the new prototype-key + GREATEST-ordering tests.

msywulak:
Ran the full `/pr-review-toolkit` (code / errors / tests / types / comments) before merge. No correctness bugs or blockers surfaced; applied the worthwhile findings in 60543e22:

- **silent-failure-hunter (observability):** the two failure modes that had *opposite* retry semantics but collapsed into one audit `reason` are now distinct — a persist failure (good snapshot, write failed → retries next tick) audits `persist_failed` + emits a direct `log.warn`; a rediscover/probe fault (negative-cached a full interval) audits `rediscover_fault`. Added a `phase` discriminant + test assertions.
- **comment-analyzer (accuracy):** fixed the "SaaS egress allowlist" phrasing in two docstrings — the host-side spec probe is a *fail-closed SSRF guard* (`isSafeExternalUrl`, opt-out via `ATLAS_OPENAPI_ALLOW_INTERNAL_HOSTS`), not a hostname allowlist; and refreshed the stale #2977 forward-reference (the scheduler shipped `setInterval`-based, not `Schedule.spaced`).
- **type-design:** documented that `RediscoveryResult.ok.drift`'s `null` arm is defensive-only.

Acknowledged-but-deferred (internal polish, no behavior impact): `RediscoverCycleResult` status/error discriminated-union tightening, `InstallOutcome.config_skip` authKind split, and dedicated `_inFlight`/initial-cycle lifecycle tests — none change behavior and the invariants are already correctly held; happy to follow up if wanted.

type + lint + targeted tests (91) green. Enabling auto-merge.

Issue #2917, opened 2026-05-28
chore(infra): populate Railway staging env vars (staging slice 21/22, HITL)


msywulak:
## Parent
#2893

## What to build

**HITL — requires Railway project owner access + outputs from slices 12–18.**

Populate the `staging` environment's variable set in Railway with every secret in `.env.example` plus staging-specific additions. Source values from the provider provisioning slices:

| Variable | Source slice |
|----------|-------------|
| `SLACK_CLIENT_ID/SECRET/SIGNING_SECRET` | #2900 (slice 12) |
| `LINEAR_CLIENT_ID/SECRET` | #2901 (slice 13) |
| `GITHUB_APP_ID/PRIVATE_KEY/WEBHOOK_SECRET` | #2902 (slice 14) |
| `GOOGLE_CLIENT_ID/SECRET` | #2903 (slice 15) |
| `TWENTY_API_KEY/BASE_URL` | #2904 (slice 16) |
| `STRIPE_SECRET_KEY/WEBHOOK_SECRET` | #2905 (slice 17) — `sk_test_...` + `whsec_test_...` |
| `RESEND_API_KEY` | #2906 (slice 18) |
| `DATABASE_URL` | #2907 (slice 19) — managed PG |
| `ATLAS_DEPLOY_MODE` | hardcoded `saas` |
| `ATLAS_API_REGION` | hardcoded `staging` (existing residency discriminator — see PRD Region-keying section) |
| `ATLAS_DEPLOY_ENV` | hardcoded `staging` (deploy-env profile discriminator — SEPARATE from `ATLAS_API_REGION`; `resolveDeployEnv()` in `lib/env-profile.ts` defaults to `production` when unset, which would run the prod auth profile: cookie prefix `atlas` instead of `atlas-staging`. Surfaced by Codex review on #3087) |
| `BETTER_AUTH_SECRET` | freshly generated 32-byte random |
| `ATLAS_ENCRYPTION_KEYS` | freshly generated versioned keyset |
| `STAGING_MAIL_SINK` | `staging-mail@useatlas.dev` |
| `STAGING_ADMIN_PASSWORD` | freshly generated random |

Wire the Railway deploy-success webhook to GitHub `repository_dispatch` so the smoke-test workflow (slice 10) fires on every staging deploy.

Trigger a manual redeploy of all three services. Confirm api-staging boots green and `ensureStagingSeed` creates the seed.

## Acceptance criteria

[...]

PR #3047, opened 2026-05-31
fix(api): derive WebAuthn rpID from web origin, fail loud on mismatch (#3045)

### LINKED ISSUE #3045: fix(api): WebAuthn rpID silently defaults to prod domain — derive from web origin / fail loud on non-prod deploys
The Better Auth passkey plugin is configured with `rpID: process.env.ATLAS_RPID ?? "app.useatlas.dev"` (`lib/auth/server.ts:~1462`). When `ATLAS_RPID` is unset, **every deploy silently inherits the production RP ID** `app.useatlas.dev`. On any host whose origin isn't a registrable suffix of `app.useatlas.dev` — staging (`app.staging.useatlas.dev`), a self-hoster on a custom domain, a preview env — the browser rejects passkeys with the opaque `The RP ID "app.useatlas.dev" is invalid for this domain`. Found on staging, but the footgun is general.

The default is a hardcoded prod constant even though `getWebOrigin()` (`lib/web-origin.ts`) already knows the deploy's real app origin (first `ATLAS_CORS_ORIGIN` entry, then `BETTER_AUTH_TRUSTED_ORIGINS`). A better fallback is the host of the configured web origin — which equals `app.useatlas.dev` on prod (so no behavior change / no passkey invalidation there) and the correct value everywhere else. At minimum, a deploy whose effective rpID can't be valid for its own web origin should fail loud at boot rather than break silently in the browser later.

**Hard constraint:** `rpID` MUST stay stable across passkey enrollment — changing it invalidates every enrolled passkey (see the comment block above the `passkey({...})` call). An explicit `ATLAS_RPID` must always win, and the prod effective rpID must remain exactly `app.useatlas.dev`. Verify prod either sets `ATLAS_RPID` explicitly or that the derived fallback resolves to `app.useatlas.dev` before changing the default.

## Design fork (decide first)
- **(A) Derive fallback from web origin** — `rpID = process.env.ATLAS_RPID ?? new URL(getWebOrigin()).hostname`. Elegant, self-correcting; prod stays `app.useatlas.dev`. Risk: rpID becomes coupled to `ATLAS_CORS_ORIGIN` ordering — a reorder could shift the derived rpID and invalidate passkeys. Mitigate by always setting `ATLAS_RPID` explicitly in SaaS.
- **(B) Fail loud** — keep the explicit-env requirement, but assert at boot that the effective rpID is equal to / a registrable suffix of `getWebOrigin()`'s host; hard-fail with an actionable message otherwise (mirrors the existing cross-origin cookie-domain fail-loud a few lines down in `server.ts`). Lower invalidation risk; turns a confusing browser error into a clear boot failure.
- **Recommended: B as the floor, A as the fallback** — derive from web origin when `ATLAS_RPID` is unset, AND assert the effective rpID is valid for the web origin (fail loud if an explicit `ATLAS_RPID` is set but bogus for the origin). Explicit env always wins.

## Key files
- `packages/api/src/lib/auth/server.ts` (~1457–1463) — the `passkey({ rpID: ... })` config + the rpID-stability warning comment; ~2530–2565 has the analogous cross-origin cookie-domain fail-loud to mirror
- `packages/api/src/lib/web-origin.ts` — `getWebOrigin()`; the existing source of truth for the deploy's app origin
- `.env.example` (~120–124) — `ATLAS_RPID` / `ATLAS_RPNAME` docs (commented-out optional today); update guidance to reflect the new fallback/assertion
- `packages/web/src/lib/auth/parse-passkey-sign-in-error.ts` — where the client maps passkey errors; consider surfacing a clearer message for the invalid-RP-ID case

## Findings
- `rpID` falls back to a hardcoded prod domain (`app.useatlas.dev`), not to the deploy's actual origin.
- `getWebOrigin()` already resolves the correct app origin per-deploy and is unused by the passkey config.
- The failure mode is browser-side and opaque — no boot-time signal that the rpID can't possibly be valid for the configured origin.
- Affects staging (#2917 must set `ATLAS_RPID` regardless), self-hosters on custom domains, and preview envs.
- rpID changes invalidate enrolled passkeys, so the fix must guarantee prod's effective rpID is unchanged.

## Suggested fix
Implement the recommended hybrid in `server.ts`: derive the rpID from `getWebOrigin()`'s hostname when `ATLAS_RPID` is unset, and add a boot-time assertion (fail loud, actionable message) that the effective rpID is equal to or a registrable suffix of the web origin's effective domain. Keep explicit `ATLAS_RPID` as an always-winning override. Update `.env.example` guidance. Confirm prod's effective rpID stays `app.useatlas.dev` (no passkey invalidation). Add unit coverage for the derivation + the assertion (valid, suffix-valid, invalid-explicit, unset-derives).

## Acceptance criteria
- [ ] With `ATLAS_RPID` unset, the rpID derives from the configured web origin (prod resolves to `app.useatlas.dev` — no change; staging resolves to `app.staging.useatlas.dev`).
- [ ] An explicit `ATLAS_RPID` always overrides the derived value.
- [ ] A deploy whose effective rpID is not valid for its web origin fails loud at boot with an actionable message (not a silent browser-side error).
- [ ] Prod's effective rpID is verified unchanged (`app.useatlas.dev`) so no enrolled passkey is invalidated.
- [ ] Unit tests cover: unset→derive, explicit-override, suffix-valid, invalid-explicit→fail-loud.
- [ ] `.env.example` rpID guidance updated to describe the derive-from-origin fallback + when to set it explicitly.

## Notes
Filed under Staging Environment (#57) since staging surfaced it and the fix directly serves "passkeys work on staging without a manual env-var landmine," but the hardening benefits all non-prod / self-hosted deploys. The immediate staging unblock (set `ATLAS_RPID=app.staging.useatlas.dev`) is captured on #2917 and is independent of this code change.

msywulak:
## Summary

`rpID` fell back to the hardcoded prod constant `app.useatlas.dev` whenever `ATLAS_RPID` was unset, so **every non-prod deploy silently inherited prod's RP ID** and passkeys broke with an opaque browser error (`The RP ID "app.useatlas.dev" is invalid for this domain`) on staging, self-hosted custom domains, and preview envs.

- **Derive-from-origin fallback** — new `resolvePasskeyRpId(env, webOrigin)`: explicit `ATLAS_RPID` always wins → else derive from the configured web origin's host via `getWebOrigin()` → else the legacy default. Prod's web origin is `https://app.useatlas.dev`, so the derived value is **exactly `app.useatlas.dev` — unchanged, no passkey invalidation**. Staging now derives `app.staging.useatlas.dev`; localhost derives `localhost`.
- **Fail loud at boot** — whenever a web origin is configured, asserts the effective rpID is equal to or a registrable-domain suffix of the origin host, and throws an actionable error otherwise. (This is *stronger* than the cross-origin cookie-domain check in `getAuthInstance()`, which only warns — a wrong rpID can't work at all.) A bogus explicit `ATLAS_RPID` is caught here too; the derived path can never produce an invalid value, so unset deploys never throw.
- **Case-insensitive comparison** — DNS hostnames are case-insensitive and browsers lowercase the rpID before matching, so a valid mixed-case explicit `ATLAS_RPID` doesn't false-positive fail-loud. The returned rpID is left verbatim (rpID stability — never mutate the enrolled value).
- **No silent skip on a broken origin** — a configured-but-unusable origin (unparseable, or scheme-less `host:port` that `new URL()` parses as a protocol → empty host) is logged at **error** level naming the consequence + fix, rather than slipping past as a silent `null`.
- **No regression for minimal setups** — a `null` web origin (single-origin / self-hosted with neither var set) preserves today's behavior: legacy default, no hard-fail.
- **Clearer client error** — routes the browser's invalid-rpID `SecurityError` to actionable "contact your administrator" copy in `parsePasskeySignInError` (the residual self-hosted case the boot assertion can't reach).
- Updates `.env.example` rpID guidance + adds unit coverage.

## rpID stability (hard constraint)

Prod verified unchanged: `deploy/README.md` sets `ATLAS_CORS_ORIGIN=https://app.useatlas.dev`, and prod does not set `ATLAS_RPID` — both paths resolve to exactly `app.useatlas.dev`, so no enrolled passkey is invalidated. Explicit `ATLAS_RPID` always overrides. The e2e passkey tests set `ATLAS_RPID=localhost` explicitly, which still wins and validates against the `localhost` origin.

## Test plan

- [x] `ATLAS_RPID` unset + web origin → derives origin host (prod `app.useatlas.dev`, staging `app.staging.useatlas.dev`, localhost `localhost`)
- [x] explicit `ATLAS_RPID` always overrides (incl. parent-domain pin); mixed-case explicit value valid for origin → no false fail-loud
- [x] registrable-domain suffix valid (e.g. `useatlas.dev` for `app.staging.useatlas.dev`); mixed-case origin derives lowercase rpID
- [x] explicit `ATLAS_RPID` invalid for origin → throws at boot with actionable message
- [x] unrelated-domain suffix without dot boundary rejected (`useatlas.dev` ✗ for `notuseatlas.dev`)
- [x] `getWebOrigin()` null → legacy default, no hard-fail
- [x] unparseable / scheme-less `host:port` origin → no crash, validation skipped (logged at error), degrades to explicit/default
- [x] client parser routes invalid-rpID `SecurityError` to setup-issue copy (Chrome + suffix phrasings, wire + thrown)
- [x] `/ci` — lint, type, full test (api 535 + others), syncpack, all drift/discipline gates green
- [x] full `/pr-review-toolkit:review-pr` pass (code, tests, comments, errors); review findings addressed in the second commit

## Notes

This does **not** unblock staging passkeys on its own — staging still needs `ATLAS_RPID=app.staging.useatlas.dev` set in Railway (tracked on #2917, HITL). This change is the general hardening so the footgun never silently bites again.

Closes #3045

## Summary by CodeRabbit

* **Documentation**
  * Expanded WebAuthn/passkey guidance: detailed rpID derivation, origin mappings, multi-region/custom domain notes, and implications of changing rpID after passkeys are enrolled.

* **Bug Fixes**
  * Improved detection and clearer user-facing messages for misconfigured passkey (rpID) setups; advises contacting admin or using alternate sign-in.
  * Startup diagnostics now surface explicit invalid rpID errors.

* **Tests**
  * Added coverage for rpID resolution and related error-handling scenarios.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -117,9 +117,17 @@ ATLAS_ADMIN_EMAIL=admin@useatlas.dev
 #                                         # overridable via env by design.
 # ATLAS_MFA_ISSUER=Atlas                  # Issuer label shown in TOTP authenticator apps when users
 #                                         # enroll an MFA device. Default: "Atlas".
-# ATLAS_RPID=app.useatlas.dev             # WebAuthn Relying Party ID — the registrable domain that
-#                                         # passkeys are bound to. Self-hosted: set to your auth domain.
-#                                         # Changing this AFTER passkeys are enrolled invalidates them.
+# ATLAS_RPID=app.useatlas.dev             # WebAuthn Relying Party ID — the registrable domain passkeys are
+#                                         # bound to. When UNSET, it is derived from the configured web origin's
+#                                         # host (first ATLAS_CORS_ORIGIN entry, then BETTER_AUTH_TRUSTED_ORIGINS):
+#                                         # prod → app.useatlas.dev, staging → app.staging.useatlas.dev, localhost → localhost.
+#                                         # If no web origin is configured, falls back to app.useatlas.dev.
+#                                         # Boot FAILS LOUD if the effective rpID isn't valid for the web origin
+#                                         # (must equal the origin host or be a parent domain) — no silent browser breakage.
+#                                         # SaaS multi-region: ALWAYS set this explicitly so a reorder of
+#                                         # ATLAS_CORS_ORIGIN can't shift the derived value. Self-hosted on a custom
+#                                         # domain: set to your auth domain. Changing this AFTER passkeys are
+#                                         # enrolled invalidates every enrolled passkey.
 # ATLAS_RPNAME=Atlas                      # Human-readable string the OS shows in the passkey prompt
 #                                         # ("Use your saved passkey for <ATLAS_RPNAME>?"). Default: "Atlas".
 # ATLAS_ABUSE_QUERY_RATE=200              # Max queries per workspace per sliding window (abuse prevention)
--- a/packages/api/src/lib/__tests__/startup.test.ts
+++ b/packages/api/src/lib/__tests__/startup.test.ts
@@ -64,6 +64,8 @@ const MANAGED_VARS = [
   "BETTER_AUTH_SECRET",
   "BETTER_AUTH_URL",
   "BETTER_AUTH_TRUSTED_ORIGINS",
+  "ATLAS_CORS_ORIGIN",
+  "ATLAS_RPID",
   "ATLAS_AUTH_JWKS_URL",
   "ATLAS_AUTH_ISSUER",
   "ATLAS_AUTH_AUDIENCE",
@@ -91,6 +93,8 @@ beforeEach(() => {
   delete process.env.BETTER_AUTH_SECRET;
   delete process.env.BETTER_AUTH_URL;
   delete process.env.BETTER_AUTH_TRUSTED_ORIGINS;
+  delete process.env.ATLAS_CORS_ORIGIN;
+  delete process.env.ATLAS_RPID;
   delete process.env.ATLAS_AUTH_JWKS_URL;
   delete process.env.ATLAS_AUTH_ISSUER;
   delete process.env.ATLAS_AUTH_AUDIENCE;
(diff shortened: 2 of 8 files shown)


msywulak:
Thanks @chatgpt-codex-connector — all three P2s were valid; addressed in `bda77a7f`:

1. **Eager startup validation.** Correct that the `buildPlugins()` throw is lazy (first managed-auth request / boot migration, where `migrateAuthTables` catches it into a generic "Better Auth migration failed" log) — not a clean fail-loud-at-boot. Extracted the pure resolver into `lib/auth/rpid.ts` (no `better-auth` dep) and now call it from `checkManagedAuthMode` in `startup.ts`, surfacing a new `INVALID_RP_ID` diagnostic eagerly on `/health` and route 503s with the actionable message.
2. **IP-literal rpID.** Real gap — an IP origin derived an IP rpID that passes the suffix check but can never work in-browser. Now logged loud at error level. I did **not** throw (would break non-passkey managed auth for an IP dev host) and did **not** remap loopback→`localhost` (the page is still served from the IP, so the browser still rejects `localhost` as the rpID — the remap wouldn't actually make passkeys work).
3. **rpID under a cancellation code.** The `AUTH_CANCELLED` fast-path could swallow an invalid-rpID `SecurityError` that wears the same shape (the file's own outcome doc notes this). Now checks `isInvalidRpIdError(message)` inside the cancellation branch before returning `silent`; the two-term guard keeps genuine cancellations silent.

Tests added for all three; full `/ci` green. CodeRabbit reported no actionable comments.

PR #3051, opened 2026-05-31
fix(api): gate breaking-change classifier on effective-required ancestry (#3050)

### LINKED ISSUE #3050: Breaking-change classifier over-flags required children of optional request bodies/refs
Follow-up from Codex review on #3049 (#2979). Two related false-positives in `classifyBreakingChanges` (`packages/api/src/lib/openapi/breaking-change.ts`), both deferred from the PR because a correct fix needs the diff layer to carry structured ancestry rather than a string-path hack.

**The gap.** An added required field is flagged `field_required_added` (breaking) whenever it sits on a request surface (`requestBody:*` / `param:*`), regardless of whether the *enclosing* container is optional:

1. An upstream spec adds an **optional** request body (`requestBody.required: false`, the OpenAPI default) whose schema has a required property `foo`. The diff emits `requestBody:<media>.foo` with `after.required === true` → flagged breaking, even though existing calls omit the whole body and keep working. Same for an optional object property with a required child.
2. When a request body/param is a `$ref` to a component schema, an added required property surfaces only in `diff.schemas.changed` with a bare path and is (intentionally, for now) treated as additive — so it's the inverse miss. See the "ADDITIVE (quiet)" note in the module header for why bare schema-field paths can't be disambiguated today.

**Why it wasn't patched in #3049.** The classifier receives the structured *diff*, not the full post-state graph, so it can't see an unchanged ancestor's requiredness. Inferring ancestry from the flattened string path is unsafe — media types (`application/vnd.api+json`) and property names can contain dots, so naive parsing risks *false negatives* (missing real breaking changes), which is strictly worse than the current conservative over-flag. Impact today is low: a spurious signal is a dismissible, auto-clearing alert raised only on scheduled re-discovery.

## Key files
- `packages/api/src/lib/openapi/diff.ts` — `FieldDescriptor` / `flattenOperationFields`: thread enclosing request-body + ancestor requiredness into the descriptor (or expose request-surface reachability for named schemas).
- `packages/api/src/lib/openapi/breaking-change.ts` — `classifyField` / `classifyBreakingChanges`: consume the structured ancestry; only flag an added required request field breaking when every enclosing container is required.
- `packages/api/src/lib/openapi/__tests__/breaking-change.test.ts` — add cases: optional body + required child = additive; required body + added required child = breaking; required prop added to a request-reachable component = breaking.

## Acceptance criteria
- [ ] Adding an optional request body with a required child does NOT raise a breaking signal.
- [ ] Adding a required child to an already-required request container still IS breaking (no false negative).
- [ ] A newly-required property on a component reachable only from request surfaces is breaking; one reachable from a response surface stays quiet.
- [ ] No reliance on string-path parsing of dotted media types / property names.

## Dependencies
- Builds on the structured diff (#3041) and breaking-change classifier (#2979 / #3049).

msywulak:
Closes #3050. The last pre-tag follow-up gating the **v0.0.3 — Spec Lifecycle** milestone.

## Problem

The breaking-change classifier (`classifyBreakingChanges`) flagged **any** added required field on a request surface as breaking, regardless of whether the *enclosing* container was optional. Two related defects, both deferred from #3049 (Codex review) because a correct fix needs structured ancestry rather than a string-path hack:

1. **Over-flag (false positive).** An upstream adds an **optional** request body (`requestBody.required: false`) whose schema has a required property — or a required child under an optional ancestor. Existing calls omit the whole container and keep working, yet a spurious (dismissible, auto-clearing) drift alert was raised on scheduled re-discovery.
2. **Inverse miss (false negative).** When a request body/param is a `$ref` to a component, an added required property surfaced only as a bare schema-field path and was treated as additive — so a genuinely-breaking required field on a request-only component went unsignalled.

## Fix

Thread an `effectiveRequired` flag through the diff layer (`diff.ts`): **true only when a field is required AND every enclosing container up to a request surface is also required.** It propagates through required-preserving edges (required object properties, array items, `allOf` branches); optional containers and `oneOf`/`anyOf` members break the chain.

For `$ref`'d components, a graph walk (`computeRequestExclusiveSchemas`) marks schemas reachable from a **required request chain and never a response** as request-exclusive, seeding their fields' `effectiveRequired`. A schema also reachable from a response stays quiet (the standing conservative "ambiguous surface ⇒ additive" policy).

The classifier now reads the boolean instead of sniffing the dotted path — **no false negatives on media types / property names that contain dots** (`application/vnd.api+json`, dotted prop names).

`effectiveRequired` is deliberately **excluded** from `descriptorsEqual` / `serializeDescriptor`: it is a derived classification hint, so an ancestor's requiredness change must not read as a child retype, and composition-branch keys stay stable.

## Acceptance criteria (all covered by new tests)

- [x] Adding an optional request body with a required child does **not** raise a breaking signal.
- [x] Adding a required child to an already-required request container still **is** breaking (no false negative) — incl. the nested optional-vs-required-ancestor pair.
- [x] A newly-required property on a component reachable only from request surfaces is breaking; one reachable from a response surface stays quiet (+ the both-surfaces no-false-positive case).
- [x] No reliance on string-path parsing of dotted media types / property names.

## Tests

7 new cases in `breaking-change.test.ts` (built through the real `buildOperationGraph` + `diffOperationGraphs`). Full local `/ci` green (lint, type, test, syncpack, all drift gates, twenty-resolver, published-symbols).

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/3051"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782850304&installation_id=135337507&pr_number=3051&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F3051&signature=d0aff715c92d6f2aed0e32462f920604837fc6a3d5ad7a342a5860969716a6dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate breaking-change detection for added request fields: classification now uses whole-chain "effective required" semantics, respects unions vs. intersections, propagates through arrays/allOf, handles cycles, and ignores dotted media-type quirks so response-only components aren’t misclassified as request-breaking.

* **Tests**
  * Expanded test coverage validating required-field classification across nested schemas, $ref reachability, arrays, allOf/oneOf behavior, and invariants preventing spurious descendant changes.

### DIFF
--- a/packages/api/src/lib/openapi/__tests__/breaking-change.test.ts
+++ b/packages/api/src/lib/openapi/__tests__/breaking-change.test.ts
@@ -397,6 +397,155 @@ describe("classifyBreakingChanges — mixed changesets", () => {
   });
 });
 
+// ── #3050: effective-required gating (optional containers + ref'd components) ─
+
+describe("classifyBreakingChanges — effective-required gating (#3050)", () => {
+  /** Classify the diff between two arbitrary docs (not clones of BASE_DOC). */
+  function classifyBetween(prev: OpenApiDoc, next: OpenApiDoc) {
+    return classifyBreakingChanges(diffOperationGraphs(graph(prev), graph(next)));
+  }
+  /** A doc with one POST whose request body wraps `bodySchema` at the given requiredness. */
+  function postBodyDoc(bodyRequired: boolean, bodySchema: SchemaNode): OpenApiDoc {
+    return {
+      openapi: "3.0.0",
+      info: { title: "T", version: "1.0.0" },
+      paths: {
+        "/things": {
+          post: {
+            operationId: "createThing",
+            requestBody: { required: bodyRequired, content: { "application/json": { schema: bodySchema } } },
+            responses: { "200": { description: "ok" } },
+          },
+        },
+      },
+      components: { schemas: {} },
+    };
+  }
+
+  it("adding a required child to an OPTIONAL request body is additive", () => {
+    const prev = postBodyDoc(false, { type: "object", required: ["a"], properties: { a: { type: "string" } } });
+    const next = postBodyDoc(false, {
+      type: "object",
+      required: ["a", "b"],
+      properties: { a: { type: "string" }, b: { type: "string" } },
+    });
+    expect(classifyBetween(prev, next).breaking).toBe(false);
+  });
+
+  it("adding a required child under an OPTIONAL ancestor (in a required body) is additive", () => {
+    // Body IS required, but `address` is optional → a caller omitting it keeps working.
+    const addr = (zipRequired: boolean): SchemaNode => ({
+      type: "object",
+      ...(zipRequired ? { required: ["zip"] } : {}),
+      properties: { street: { type: "string" }, ...(zipRequired ? { zip: { type: "string" } } : {}) },
+    });
+    const prev = postBodyDoc(true, { type: "object", required: [], properties: { address: addr(false) } });
+    const next = postBodyDoc(true, { type: "object", required: [], properties: { address: addr(true) } });
+    expect(classifyBetween(prev, next).breaking).toBe(false);
+  });
+
+  it("adding a required child under a REQUIRED ancestor (in a required body) IS breaking", () => {
+    const addr = (zipRequired: boolean): SchemaNode => ({
+      type: "object",
+      ...(zipRequired ? { required: ["zip"] } : {}),
+      properties: { street: { type: "string" }, ...(zipRequired ? { zip: { type: "string" } } : {}) },
+    });
+    const prev = postBodyDoc(true, { type: "object", required: ["address"], properties: { address: addr(false) } });
+    const next = postBodyDoc(true, { type: "object", required: ["address"], properties: { address: addr(true) } });
+    const a = classifyBetween(prev, next);
+    expect(a.breaking).toBe(true);
+    expect(
+      hasReason(a.reasons, "field_required_added", {
+        operationId: "createThing",
+        path: "requestBody:application/json.address.zip",
+      }),
+    ).toBe(true);
+  });
+
+  /** A doc using `WidgetInput` only on a required request body — request-EXCLUSIVE. */
+  function requestOnlyComponentDoc(inputRequired: string[], extraProp?: string): OpenApiDoc {
+    const props: Record<string, SchemaNode> = { name: { type: "string" } };
+    if (extraProp) props[extraProp] = { type: "string" };
+    return {
+      openapi: "3.0.0",
+      info: { title: "T", version: "1.0.0" },
+      paths: {
+        "/widgets": {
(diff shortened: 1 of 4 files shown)


msywulak:
Addressed both bots in `a363fd4f`. Both findings were real classifier false positives — thanks.

**1. `allOf` branch root marked required** (Codex P2 @ `diff.ts:325`)
`describeNode` set `effectiveRequired = chainRequired` regardless of whether the node was *itself* required, so a required body gaining an all-optional `allOf` branch flagged the `…|allOf[n]` root as `field_required_added`. Now enforces the documented invariant `effectiveRequired ⟹ required` (`required && chainRequired`) — a container the chain merely passes through (an `allOf` branch root, an array-items node) is no longer flagged; genuinely-required descendants still are. Regression test: *"gaining an all-optional allOf branch on a required body is additive"*.

**2. Schema request-exclusivity seeded from the wrong graph** (CodeRabbit outside-diff + Codex P2 @ `diff.ts:468` & `:470`)
You're right that `nextExclusive` is the wrong seed. A breaking change breaks a **pre-existing** caller, and a caller only exists for a request surface that already existed. The schema-field `effectiveRequired` is now seeded from **`prevExclusive`** (the schema's role in the *prior* spec), which fixes both directions in one move:
- **False positive** (Codex @468 / CodeRabbit): a new required op reusing a previously response-only/unused component + a new required prop → prior-exclusive is `false` → additive. Regression test: *"a NEW required op reusing a previously response-only component + an added required prop is additive"*.
- **False negative** (Codex @470): a pre-existing required-request component that *newly* also appears in a response → prior-exclusive is `true`, so the fresh response reachability no longer masks the real request break. Regression test: *"a NEW response on a PRE-EXISTING required-request component does NOT mask an added required prop"*.

Both flatten sides use the same prior seed — `effectiveRequired` is excluded from `descriptorsEqual`/`serializeDescriptor`, so the seed only sets the verdict on the `added` side and never perturbs change detection.

Full local `/ci` green; 78 tests in `breaking-change.test.ts` + the `diff.ts` suite pass.

Issue #3065, opened 2026-06-01
S1b — Restore a conversation's scope when it is opened


msywulak:
## Parent
#3063

## What to build
Opening a saved conversation from the sidebar must restore **that conversation's** scope into the picker. Today `handleSelectConversation` loads the messages and sets the conversation id but never reads the persisted SQL routing (group / member / routing mode) back into the picker, so the picker keeps whatever the new-chat seed set — the per-conversation-authoritative promise is unfulfilled.

Wire the loaded conversation's routing into the picker on open, taking precedence over the sticky preference and default seed. Precedence on load: **conversation row > sticky preference > default seed.**

Per [ADR-0011](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0011-unified-conversation-scope.md) + CONTEXT.md → Conversation scope.

## Key files
- `packages/web/src/ui/components/atlas-chat.tsx` — `handleSelectConversation` (~522), seed/restore effect (~205-238)
- `packages/web/src/ui/components/chat/env-picker.tsx` — selection wiring
- conversation load path (`convos.loadConversation`) — ensure routing fields are available to restore
- Tests: `packages/web/src/ui/__tests__/env-picker.test.tsx`

## Acceptance criteria
- [ ] Opening a saved conversation restores its group / member / routing mode into the picker.
- [ ] A restored conversation's scope takes precedence over the sticky preference and the default seed.
- [ ] Switching between two conversations with different scopes updates the picker each time.
- [ ] A legacy conversation with null routing falls back safely (treated as pin to its `connectionId`).

## Dependencies
Blocked by #3064 (shares the seed/restore + provenance seam).

PR #3074, opened 2026-06-01
fix(web): preserve legacy group-less conversation pin (#3065)

msywulak:
## Summary
Started as test-recovery for #3073, but **Codex's review caught a real regression** in #3073's validator (merged) — so this is now a fix.

A legacy conversation row with `connectionGroupId: null` but a **non-null** `connectionId` (a pre-0067 row whose group was never backfilled) returned `seed` once groups were loaded. `handleSelectConversation` then clears the scope, the seed/restore effect applies the **sticky preference / group primary as a non-null member**, and the transport sends that override — **silently switching a conversation pinned to e.g. `eu-prod` onto the default environment**. My earlier recovery test used the primary `us-prod`, which *masked* the switch (Codex flagged exactly this).

**Fix:** when a row has no group but a live connection, locate the group that now **owns** that connection and restore the pin `{owningGroup, connectionId, mode}`. Only `seed` if the connection no longer exists in any visible group.

- This both preserves the user's pin **and** supplies a valid group, so no `invalid_connection_group`.
- Cold-start (`groups: []`) still restores the row verbatim — the transport sends the real `connectionId`, so the route pins correctly.

## Test plan
- [x] Tests updated to assert the owning-group restore with the **non-primary** `eu-prod` (regression can't be masked by the primary), plus a connection-genuinely-removed → `seed` case. 85 pass / 0 fail.
- [x] `bun run lint` / `bun run type` clean; 171 affected web tests green.

## Provenance
- This is why we wait for the bots: CodeRabbit found nothing and my in-house review agents earlier judged this case "defensible," but Codex correctly identified the silent env-switch. Validates holding merges for bot review.

Refs #3065

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restoration of previously saved and pinned conversations
  * Enhanced fallback handling when connection information is no longer available or has changed
  * Better support for legacy conversation data formats with refined recovery logic

* **Tests**
  * Expanded test coverage for conversation restoration scenarios and edge cases

### DIFF
--- a/packages/web/src/ui/__tests__/env-picker.test.tsx
+++ b/packages/web/src/ui/__tests__/env-picker.test.tsx
@@ -1647,6 +1647,31 @@ describe("resolveConversationScope — restore a conversation's scope on open (#
     ).toEqual({ kind: "seed" });
   });
 
+  test("seeds a legacy group-less row (connectionId but null group) once groups are loaded", () => {
+    // Reachable production state (migration 0067 only backfilled the group for
+    // rows whose connection still joined): connectionId set, connectionGroupId
+    // null. With groups loaded there is no group to validate, so seed — the
+    // chat route still pins execution by reading connection_id back off the row.
+    expect(
+      resolveConversationScope(
+        { connectionGroupId: null, connectionId: "us-prod", routingMode: "pin" },
+        groups,
+      ),
+    ).toEqual({ kind: "seed" });
+  });
+
+  test("optimistically restores a legacy group-less row (null group) on cold-start", () => {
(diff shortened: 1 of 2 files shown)


msywulak:
@codex review

Thanks — good catch on the silent env-switch. The earlier `seed` test masked it with the primary `us-prod`. Fixed in 76ca5bc3: a group-less row with a live connection now locates the **owning group** and restores the pin `{owningGroup, connectionId, mode}` (only `seed` if the connection is gone from every group). Tests now assert the restore with the non-primary `eu-prod`. Please re-review.

Issue #3067, opened 2026-06-01
S2b — REST-only focus (suspend SQL for a conversation)


msywulak:
## Parent
#3063

## What to build
Let a user **focus** a single REST datasource so the conversation targets it exclusively and `executeSQL` is **suspended** — the "ask Stripe only" case. Focus is per-conversation, persisted, and seeds new chats via the sticky preference. Clearing focus returns to the prior default-state scope (SQL routing + exclude-set are retained, inert, while focused).

End-to-end:
- Migration: `conversations.rest_focus_datasource_id text` (nullable) + Drizzle mirror + `migrate-pg` smoke.
- `@useatlas/types` `Conversation` gains `restFocusDatasourceId`.
- Resolver: focused → resolve **only** the focus target; agent loop suspends `executeSQL` (REST-only turn; prompt/toolkit reflect it).
- `ChatScopePicker`: focus gesture (select a REST datasource as focus) + chip state (`Stripe only`) + a way to clear focus.
- Chat request body + `GET /conversations/:id` + sticky-preference store carry the focus field.

Per [ADR-0011](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0011-unified-conversation-scope.md).

## Key files
- `packages/api/src/db/migrations/####_*.sql` + `packages/api/src/db/schema.ts`
- `packages/types/src/conversation.ts`
- REST resolver + agent loop / toolkit (`executeSQL` gating) + `packages/api/src/lib/tools/rest-operation.ts`
- `ChatScopePicker` (`packages/web/src/ui/components/chat/env-picker.tsx`) + `atlas-chat.tsx` + `chat-routing-preference-store.ts`

## Acceptance criteria
- [ ] Focusing a REST datasource suspends `executeSQL`; only that datasource is reachable for the turn.
- [ ] Chip shows the focused state (e.g. `Stripe only`); focus can be cleared back to default scope.
- [ ] Focus persists across reload and seeds new chats.
- [ ] Clearing focus restores the previously-selected SQL routing + exclude-set.
- [ ] A focused conversation whose datasource was uninstalled falls back safely to default scope.

## Dependencies
Blocked by #3066 (extends the REST-scope plumbing: resolver signature, picker rows, types).

Issue #3066, opened 2026-06-01
S2a — REST scope: exclude datasources from a conversation


msywulak:
## Parent
#3063

## What to build
Make REST datasources first-class, **excludable** rows in the scope picker (the picker becomes `ChatScopePicker`). By default every workspace-global / active-group REST datasource is in scope (per ADR-0010); unchecking one **excludes** it from the conversation so the agent no longer queries it. The exclusion is per-conversation (authoritative) and updates the sticky preference for new chats. SQL routing is unaffected.

End-to-end:
- Migration: `conversations.rest_excluded_datasource_ids text[] DEFAULT '{}'` + Drizzle `schema.ts` mirror **in the same PR** + real-Postgres `migrate-pg` smoke.
- `@useatlas/types` `Conversation` gains `restExcludedDatasourceIds` (publish-then-bump-refs ordering for 0.0.x).
- `resolveWorkspaceRestDatasources(orgId, activeGroupId?, { excluded? })` drops the exclude-set **after** ADR-0010's group-scope filter and **before** credential resolution (never-rejects fail-soft contract preserved).
- Confirm-replay path (`tools/rest-operation.ts` `resolveFromContext`, `activeGroupId === undefined`) ignores the exclude-set.
- Chat request body + `GET /conversations/:id` carry the field; sticky-preference store gains it.
- `ChatScopePicker` renders REST datasources as checkboxes (default checked); chip reflects the in-scope count (e.g. `Pin · apac-prod · 2/3 REST`).

Per [ADR-0011](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0011-unified-conversation-scope.md) (supersedes ADR-0010 picker-surface).

## Key files
- `packages/api/src/db/migrations/####_*.sql` + `packages/api/src/db/schema.ts`
- `packages/types/src/conversation.ts`
- REST resolver (`resolveWorkspaceRestDatasources`, `openapi/data-candidates.ts`) + `packages/api/src/lib/tools/rest-operation.ts`
- `packages/web/src/ui/components/chat/env-picker.tsx` → `ChatScopePicker`, `packages/web/src/ui/components/atlas-chat.tsx`, `packages/web/src/lib/stores/chat-routing-preference-store.ts`
- Routes: chat + `GET /conversations/:id` (+ `GET /api/v1/me/connection-groups` already returns `restDatasources`)

## Acceptance criteria
[...]

PR #3077, opened 2026-06-01
feat(scope): REST datasource exclude-set per conversation (#3066)

msywulak:
## What

S2a of the **v0.0.4 — Conversation Scope** milestone ([#3066](https://github.com/AtlasDevHQ/atlas/issues/3066)). REST datasources are in scope by default; a conversation can now **exclude** specific ones via the scope picker so the agent stops querying them. **SQL routing (Auto/Pin/All) is unaffected.** Per [ADR-0011](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0011-unified-conversation-scope.md).

## How (by layer)

1. **Migration + schema** — `0112_conversation_rest_excluded_datasources.sql` adds `conversations.rest_excluded_datasource_ids text[] NOT NULL DEFAULT '{}'` with a matching Drizzle `schema.ts` mirror (schema-drift gate) + real-Postgres `migrate-pg` smoke (passes). Empty `{}` = all in scope, so a newly-installed REST datasource is reachable with no action.
2. **Resolver** — `ResolveWorkspaceDeps.excluded` drops excluded `install_id`s **after** ADR-0010's group-scope filter and **before** credential build (never-rejects fail-soft contract preserved). The agent binds `executeRestOperation` to exactly the filtered set, so exclusion is enforced at **both** prompt-build and tool-execution. The confirm-replay path (`resolveFromContext`) stays deps-free, so a staged write replays regardless of the exclude-set.
3. **Chat + conversations routes** — mirror the `routingMode` pattern: body field + read-back from the row + persist (`updateConversationRestExcluded`, gated by `sameStringSet`). `GET /conversations/:id` carries the field.
4. **`@useatlas/types`** — `Conversation.restExcludedDatasourceIds` (type-only; patch-bumped to `0.1.10`).
5. **Web** — REST datasources render as **checkboxes** (default checked = in scope; unchecking excludes). The chip shows the in-scope count (e.g. `Pin · prod · 2/3 REST`). Threaded through `resolveEnvSelection` / `resolveConversationScope` provenance + the sticky-preference store (seeds new chats; restores on conversation open).

### ⚠️ The #3073 bug class (transport-omits-null)
`useAtlasTransport` **always** sends `restExcludedDatasourceIds` (even `[]`), and the route distinguishes *field present as `[]`* (re-include → clear the row) from *field absent* (inherit the row). Omitting the empty array on a re-include would silently keep a stale exclusion. Covered by a chat-route regression test.

## Acceptance criteria
- [x] Migration adds the column with a matching Drizzle mirror; `migrate-pg` smoke passes.
- [x] Unchecking a REST datasource excludes it for that conversation; the agent does not query it (resolver + bound-tool tests).
- [x] Default (`'{}'`) = all in scope; a newly-installed REST datasource is reachable without action.
- [x] Exclusion persists across reload (row) and seeds new chats (sticky preference).
- [x] Confirm-replay of a staged write ignores the exclude-set.
- [x] Re-including a previously-excluded datasource (`[]`) clears it on the row (no stale-exclusion fallback).

## Tests
- Resolver exclude-set (7 cases incl. group-scope ordering, fail-soft, strict-resolver).
- Chat route: ALS-frame propagation, empty-set stripped, **re-include `[]` persists**, omit-inherits-no-update.
- Data layer: `createConversation` / `getConversation` / fork param + SELECT updates.
- Web picker: chip count, excluded renders unchecked, toggle sends the full next set, re-check clears.

## CI
Local gates all green: lint, type, full test suite, syncpack, template/security/railway/schema/oauth drift, test-discipline, twenty-resolver, published-symbols.

## Follow-ups (not this PR)
- S2b [#3067](https://github.com/AtlasDevHQ/atlas/issues/3067) — REST-only focus (`rest_focus_datasource_id`, suspends `executeSQL`). Reuses these seams.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/3077"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782921576&installation_id=135337507&pr_number=3077&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F3077&signature=3df17f87cd19bd928d4d46c8357b7fc73841f1fb15373a903696b9887bacf262"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Per-conversation REST datasource exclusion: pick/exclude REST datasources in the chat env picker; exclusions persist with conversations, restore on reopen, can be cleared, and are honored during chats and agent runs.

* **Tests**
  * Expanded unit/integration tests covering exclude-set lifecycle, route behavior, transport payloads, resolver logic, tooling, and egress resolution.

* **Chores**
  * DB migration, schema/types, and OpenAPI/docs updated to store and document the per-conversation exclude-set.

### DIFF
--- a/.claude/research/ROADMAP.md
+++ b/.claude/research/ROADMAP.md
@@ -26,10 +26,10 @@ Slices form a near-linear chain (S1a → S1b → S2a → S2b), with S3 + the #30
 
 - [x] S1a — Restore the sticky scope preference on a fresh chat — fixes reset-on-reload ([#3064](https://github.com/AtlasDevHQ/atlas/issues/3064), bug, no deps) — PR #3070
 - [x] S1b — Restore a conversation's scope when it is opened ([#3065](https://github.com/AtlasDevHQ/atlas/issues/3065), bug) — PR #3072 + scope-validation follow-up
-- [ ] S2a — REST scope: exclude datasources from a conversation ([#3066](https://github.com/AtlasDevHQ/atlas/issues/3066), feature, blocked by #3065) ⟵ next
+- [ ] S2a — REST scope: exclude datasources from a conversation ([#3066](https://github.com/AtlasDevHQ/atlas/issues/3066), feature, blocked by #3065) ⟵ in flight (branch `feat/3066-rest-scope-exclude-set`)
 - [ ] S2b — REST-only focus: suspend SQL for a conversation ([#3067](https://github.com/AtlasDevHQ/atlas/issues/3067), feature, blocked by #3066)
 - [ ] S3 — Persist the active conversation in the URL ([#3068](https://github.com/AtlasDevHQ/atlas/issues/3068), refactor, independent — synergizes with #3065)
-- [ ] OpenAPI drift — `ConversationSchema` omits `routingMode` ([#3071](https://github.com/AtlasDevHQ/atlas/issues/3071), bug, independent doc-fix surfaced by S1b)
+- [x] OpenAPI drift — `ConversationSchema` omits `routingMode` ([#3071](https://github.com/AtlasDevHQ/atlas/issues/3071), bug, independent doc-fix surfaced by S1b) — PR #3075
 
 Parent [#3063](https://github.com/AtlasDevHQ/atlas/issues/3063) stays in the Architecture Backlog (left unmodified per `/to-issues`). The **Staging environment** track (below) continues in parallel, independent of the tag train.
 
--- a/apps/docs/openapi.json
+++ b/apps/docs/openapi.json
@@ -1857,10 +1857,7 @@
                             ]
                           },
                           "restExcludedDatasourceIds": {
-                            "type": [
-                              "array",
-                              "null"
-                            ],
+                            "type": "array",
                             "items": {
                               "type": "string"
                             }
@@ -2146,10 +2143,7 @@
                       ]
                     },
                     "restExcludedDatasourceIds": {
-                      "type": [
-                        "array",
-                        "null"
-                      ],
+                      "type": "array",
                       "items": {
                         "type": "string"
                       }
--- a/apps/docs/openapi.json
+++ b/apps/docs/openapi.json
@@ -99,6 +99,12 @@
                       "all"
                     ]
                   },
+                  "restExcludedDatasourceIds": {
+                    "type": "array",
+                    "items": {
+                      "type": "string"
+                    }
+                  },
                   "boundDashboardId": {
                     "type": "string",
                     "format": "uuid"
@@ -1850,6 +1856,15 @@
                               null
                             ]
                           },
+                          "restExcludedDatasourceIds": {
+                            "type": [
+                              "array",
+                              "null"
+                            ],
+                            "items": {
+                              "type": "string"
+                            }
+                          },
                           "starred": {
                             "type": "boolean"
                           },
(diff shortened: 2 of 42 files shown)


msywulak:
## Review responses (commit `ee4f47b9`)

Thanks @coderabbitai and Codex — addressed below.

**CodeRabbit**
- ✅ **`sameExcludeSet` duplicate false-positive (major)** — fixed; it compared raw lengths, now compares `Set` sizes, matching the API-side `sameStringSet` exactly.
- ✅ **Dedupe at validation** — `restExcludedDatasourceIds` now normalizes to `[...new Set(ids)]` via a Zod transform (optional/undefined preserved so the inherit-vs-clear branch still works).
- ✅ **Param-order comment** — updated to include `rest_excluded_datasource_ids`.

**Codex (P2)**
- ✅ **Python egress** — `defaultResolveRestDatasource` now threads the exclude-set, so an excluded datasource can't be probed via `executePython` egress (matches the "agent does not query it" AC). Test added.
- ✅ **Single-env picker visibility** — `shouldRenderEnvPicker` now renders when there are REST datasources to exclude AND ≥1 SQL group, so the feature is reachable on the common one-Postgres + one-REST workspace. Truth-table tests added.
- ⏭️ **Zero-group REST-only visibility** + **exclude-set lost when SQL scope is all-null** — deferred to #3078. Both need a SQL-less picker render path + an exclude-set lifecycle decoupled from the SQL-scope provenance (a coherent chunk that pairs with S2b #3067). As shipped, the all-null-scope clear is only reachable via direct API (the picker hides for zero-group workspaces), so no UI data-loss path ships here.

Also addressed the in-house review: pinned the confirm-replay-ignores-exclusions contract, the `sameStringSet` reorder skip-UPDATE case, and `updateConversationRestExcluded` SQL/scope/no_db coverage.

msywulak:
@codex review

Latest commit `4b42f361` addresses your earlier P2s (Python egress exclude-set, single-env picker visibility) and CodeRabbit's nitpicks (sameExcludeSet dup-equality, validation dedupe, empty-set = omit uniformly). Two findings were deferred to #3078 (zero-group REST-only visibility + exclude-set lifecycle when SQL scope is all-null). Please re-review the latest SHA.

msywulak:
@codex review

`fe7dee21` is a follow-up to CodeRabbit's last finding: `restExcludedDatasourceIds` is never null on output (column is NOT NULL, `rowToConversation` always yields `string[]`/`[]`), so I dropped `.nullable()` from the response schema + `| null` from the wire type and regenerated the OpenAPI spec. No behavior change. Please re-review the latest SHA.

Issue #3109, opened 2026-06-02
ConnectionRegistry: thread workspace context through bare connections.get readers + composite-entry lifecycle (follow-up to #2783)


msywulak:
## Parent

Follow-up from #2783 / PR #3110. That PR re-keyed the routing source of truth by `(workspace_id, install_id)` in `ConnectionRegistry` (config-only `workspaceEntries` map + `getForOrg` resolving the composite) and made install/uninstall symmetric (`unregisterDatasourceInstall(workspaceId, installId)` now clears the per-workspace config). It kept the blast radius to the registry + boot loader + bridge + installer. This issue tracks the two read-side / lifecycle items deferred from it.

## Problem

After #2783, the **routing** path (`getForOrg(orgId, installId)`, used in SaaS where `pool.perOrg` is configured → org pooling ON) clones each org pool from the correct per-`(workspace, install_id)` config, and uninstall/config-update clear/refresh that config. Two gaps remain:

1. **Bare `connections.get(installId)` read paths + metadata readers still resolve a bare `install_id` with no workspace context.** First-write-wins across workspaces sharing an `install_id`:
   - `src/api/server.ts:121` (agent toolkit `get`), `src/lib/tools/sql.ts:644` (resolveConnection else-branch / org-pooling-OFF), `src/lib/dashboards.ts:878`, `src/lib/tools/profile-table.ts:66`, `src/lib/tools/check-distribution.ts:69`, `src/lib/semantic/diff.ts:223`.
   - Metadata readers (`getDBType` / `getTargetHost` / `getValidator` / `getParserDialect` / `getForbiddenPatterns`) are called with the bare `install_id` in the hot path (`sql.ts:646`, `sql.ts:1213-1214`). These read the bare (first-write-wins) entry, so for two workspaces sharing an `install_id` they return the FIRST workspace's metadata. Two sub-cases (raised by CodeRabbit + Codex on PR #3110): (a) **same db type** — `getDBType`/`getValidator`/`getParserDialect` agree, only `getTargetHost` diverges → cosmetic audit-log inaccuracy; (b) **different db types** (e.g. one postgres, one mysql sharing an `install_id`) — the non-first workspace is validated/wrapped with the wrong dialect, which can reject valid queries (fails closed — no data leak, and routing is already correct post-#2783). Both need the metadata readers scoped to `(workspace_id, install_id)` like `getForOrg`.
[...]

PR #3110, opened 2026-06-02
fix(db): retire DISTINCT ON multi-tenant ConnectionRegistry collision (#2783)

### LINKED ISSUE #2783: ConnectionRegistry: per-(workspace, install_id) registration to retire DISTINCT ON multi-tenant collision
## Parent

Follow-up from #2744's 1.5.3 cutover. Filed because the cutover preserved (rather than fixed) the pre-existing multi-tenant collision in \`ConnectionRegistry\`.

## Problem

\`loadSavedConnections\` registers one base pool per \`install_id\` via \`DISTINCT ON (install_id)\`. When two workspaces own datasource installs sharing an \`install_id\` (e.g. both name their warehouse \`warehouse\` or both auto-own the demo at \`install_id='__demo__'\`), only one row's URL feeds the registry; the other workspace's pool silently routes to the wrong DB.

This is preserved pre-cutover behaviour — it existed in the legacy \`connections\` table model too (composite PK \`(id, org_id)\` with the same DISTINCT ON hack). The cutover (#2744) made it more visible because every workspace now owns its own \`demo-postgres\` install at \`install_id='__demo__'\`, so the collision is the rule, not the exception.

## What changes

\`ConnectionRegistry\` becomes keyed by \`(workspace_id, install_id)\` rather than \`install_id\` alone. \`getForOrg(orgId, installId)\` resolves the registry entry by that composite, lazy-registering on cache miss from \`workspace_plugins\` directly. The base \`default\` connection (env-driven, runtime-only) keeps its current shape.

## Key files

- \`packages/api/src/lib/db/connection.ts\` — \`ConnectionRegistry\` class; \`getForOrg\`, \`register\`, the \`entries\` map.
- \`packages/api/src/lib/db/internal.ts\` — \`loadSavedConnections\` (the DISTINCT ON site).
- \`packages/api/src/lib/db/__tests__/internal.test.ts\` — tests need updating for the multi-tenant case.

## Acceptance criteria

- Two workspaces sharing an install_id end up with independent base pools (no silent collapse).
- The \`default\` connection stays runtime-only / single-instance.
- A regression test asserts the collision is gone.

## Out of scope

- Routes / lib code that currently call \`connections.get(installId)\` without an orgId — those will need updating to thread orgId through.

msywulak:
## What & why

Fixes the multi-tenant `ConnectionRegistry` collision (#2783).

`loadSavedConnections` registered one base pool per `install_id` via `DISTINCT ON (install_id)`. In SaaS — where `pool.perOrg` is configured (`deploy/api/atlas.config.ts`), so org pooling is ON — **every** query routes through `getForOrg(orgId, installId)`, which cloned an org-scoped pool from that single shared base entry. Two workspaces owning datasource installs that share an `install_id` (the demo case: every workspace auto-owns the demo at `install_id='__demo__'`; or two workspaces both name their warehouse `warehouse`) therefore collapsed onto one URL — the second tenant's queries silently routed to the first's DB. That's a data-isolation defect, so correctness here trumps features.

## Approach

Blast radius kept to **registry + loader + bridge glue + tests** (per the issue's scope; the read-side `connections.get(installId)` callers are deferred to #3109).

- **`connection.ts`** — `ConnectionRegistry` gains a **config-only** `workspaceEntries` map keyed by `(workspace_id, install_id)` (holds no live pool — `getForOrg` clones on demand). `getForOrg` resolves the per-workspace config first, falling back to the bare `entries` row for the runtime-only `default`, region pools, plugin-managed connections, and self-hosted single-tenant installs. New primitives: `registerForWorkspace` / `hasForWorkspace` / `getTargetHostForWorkspace` / `unregisterForWorkspace`.
- **`internal.ts`** (`loadSavedConnections`) — drops `DISTINCT ON`, loads every non-archived datasource install (one row per `(workspace_id, install_id)`).
- **`datasource-registry-bridge.ts`** — registers per-workspace (config-only upsert) **and** keeps the bare registration, so install-id-keyed readers (`getDBType` / `getTargetHost` / validators) and self-hosted `connections.get(installId)` keep working unchanged — **no reader churn, no async, no installer changes**. Uninstall leaves a benign mode-gated-stale config entry (`isConnectionVisibleInMode` fail-closes archived installs before `getForOrg`); config updates refresh via the upsert.

The `default` connection stays runtime-only / single-instance.

## Acceptance criteria

- [x] Two workspaces sharing an `install_id` get independent base pools (no collapse) — `internal.test.ts` (loader), `datasource-registry-bridge.test.ts` (bridge), `tenant-pool.test.ts` (`getForOrg` clones independently).
- [x] `default` connection stays runtime-only / single-instance.
- [x] Regression test in `db/__tests__/internal.test.ts` asserts the collision is gone (RED-verified against the pre-fix bridge).

## Testing

- `bun run test` (full suite) — pass. Type, lint, syncpack, template/security/railway/schema/oauth/test-discipline/twenty/published-symbols drift gates — all pass.
- Regression tests verified RED: simulating the pre-fix bridge, the loader test reports `count=1` (collapse) and the bridge test's second-workspace registration returns `false`.

## Follow-up

#3109 — thread workspace context through the bare `connections.get(installId)` read paths + eager composite-entry lifecycle on uninstall/update.

Closes #2783

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/3110"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783014776&installation_id=135337507&pr_number=3110&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F3110&signature=291d4846175f260d69c70e27e151af33c2e085ff2cab9d437594f3235801902b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved multi-tenant routing collisions: workspaces that share the same datasource installation now keep independent connection settings, preventing config conflicts and ensuring correct connection selection per workspace.
  * Improved install/uninstall behavior so registrations and removals are workspace-aware and idempotent, reducing stale routing and improving stability.

### DIFF
--- a/packages/api/src/lib/db/__tests__/datasource-registry-bridge.test.ts
+++ b/packages/api/src/lib/db/__tests__/datasource-registry-bridge.test.ts
@@ -17,6 +17,11 @@ const registerCalls: Array<{ id: string; url: string; schema?: string; descripti
 const unregisterCalls: string[] = [];
 let registeredIds = new Set<string>();
 
+// Per-(workspace, install_id) registrations — the routing source of truth (#2783).
+const wsRegisterCalls: Array<{ workspaceId: string; installId: string; url: string; schema?: string; description?: string }> = [];
+let wsRegisteredKeys = new Set<string>();
+const wsKey = (workspaceId: string, installId: string) => `${workspaceId}::${installId}`;
+
 const mockRegister: Mock<(id: string, cfg: { url: string; schema?: string; description?: string }) => void> = mock(
   (id: string, cfg: { url: string; schema?: string; description?: string }) => {
     registerCalls.push({ id, url: cfg.url, schema: cfg.schema, description: cfg.description });
@@ -30,12 +35,23 @@ const mockUnregister: Mock<(id: string) => boolean> = mock((id: string) => {
   return had;
 });
 const mockHas: Mock<(id: string) => boolean> = mock((id: string) => registeredIds.has(id));
+const mockRegisterForWorkspace: Mock<(workspaceId: string, installId: string, cfg: { url: string; schema?: string; description?: string }) => void> = mock(
+  (workspaceId: string, installId: string, cfg: { url: string; schema?: string; description?: string }) => {
+    wsRegisterCalls.push({ workspaceId, installId, url: cfg.url, schema: cfg.schema, description: cfg.description });
+    wsRegisteredKeys.add(wsKey(workspaceId, installId));
+  },
+);
+const mockHasForWorkspace: Mock<(workspaceId: string, installId: string) => boolean> = mock(
+  (workspaceId: string, installId: string) => wsRegisteredKeys.has(wsKey(workspaceId, installId)),
+);
 
 mock.module("@atlas/api/lib/db/connection", () => ({
   connections: {
     register: mockRegister,
     unregister: mockUnregister,
     has: mockHas,
+    registerForWorkspace: mockRegisterForWorkspace,
+    hasForWorkspace: mockHasForWorkspace,
   },
 }));
 
@@ -46,16 +62,22 @@ beforeEach(async () => {
   bridge = await import("../datasource-registry-bridge");
   registerCalls.length = 0;
   unregisterCalls.length = 0;
+  wsRegisterCalls.length = 0;
   registeredIds = new Set<string>();
+  wsRegisteredKeys = new Set<string>();
   mockRegister.mockClear();
   mockUnregister.mockClear();
   mockHas.mockClear();
+  mockRegisterForWorkspace.mockClear();
+  mockHasForWorkspace.mockClear();
 });
 
 afterEach(() => {
   registerCalls.length = 0;
   unregisterCalls.length = 0;
+  wsRegisterCalls.length = 0;
   registeredIds = new Set<string>();
+  wsRegisteredKeys = new Set<string>();
 });
 
 const ROW = (slug: string) =>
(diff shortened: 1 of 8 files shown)


msywulak:
Thanks @coderabbitai and Codex — both flagged the same two items, and they're real but **deferred to #3109** (not regressions, so they don't block this PR):

**1. Cross-dbType shared `install_id` → wrong metadata** (`getDBType` / `getValidator` / `getParserDialect` / `getForbiddenPatterns` resolve the bare install_id, so if two workspaces share an `install_id` with *different* db types, the non-first workspace gets the first's dialect).
- This PR does **not** regress it: pre-PR `DISTINCT ON (install_id)` collapsed those workspaces onto one base pool, so they had wrong *routing* (the data-isolation bug #2783) **and** wrong metadata. This PR fixes the routing (correct per-workspace pool); the metadata readers are still bare-keyed.
- It fails **closed** (a mismatched parser dialect rejects queries — it does not leak data), and only triggers on the narrow cross-dbType-shared-`install_id` case. The common collision (`__demo__`, or two same-type `warehouse`s) is same-dbType → correct.
- The metadata-reader threading is the read-side work scoped in #3109.

**2. Stale org pool on config-update** (`registerForWorkspace` upserts the base config but doesn't drain a cached `orgEntries` clone, which keeps the old URL until LRU/restart).
- Pre-existing: `unregisterDatasourceInstall` never touched `orgEntries` before this PR either, so config-update staleness is unchanged. Tracked in #3109 (eager org-pool drain on update).

The uninstall fail-open finding (stale per-workspace config surviving `getForOrg`'s priority resolution) **was** fixed in this PR (commit 583840c — symmetric `unregisterDatasourceInstall(workspaceId, installId)` + sibling-safe bare removal), which CodeRabbit confirmed.

Issue #2016, opened 2026-05-03
feat(sdk): @useatlas/webhook-publisher — unify the 3 outbound webhook senders (HMAC + retry)


msywulak:
## Context

Surfaced in #1924 (sub-processor change feed). Atlas now has **three** independent outbound webhook senders, each with its own signing + retry conventions. The original "file this when a second consumer appears" trigger has fired (and then some) — extract a single `@useatlas/webhook-publisher` primitive and migrate all three onto it.

This issue is the **full** scope (decided 2026-06-02): a published `@useatlas/webhook-publisher` package consumed by both the server (`packages/api`, `ee/`) and the `@useatlas/webhook-action` plugin, with **pluggable signing** so no consumer's wire format breaks.

> **Sequencing:** picked up **after** the current 7 arch-backlog easy-wins land (#2983 / #2987 / #2989 / #2708 / #2937 / #2377 / #3109).

## Current state — three divergent senders

| Sender | Path | Signing | Replay guard | Retry |
|--------|------|---------|--------------|-------|
| sub-processor publisher | `packages/api/src/lib/sub-processor-publisher/index.ts` (`signRequest` / `deliver`) | `X-Webhook-Signature: sha256=<hex>` over `${ts}:${body}` | ✅ timestamp | ✅ 3× capped exp backoff, 10s timeout, 4xx-permanent / 5xx-retry, tagged outcomes |
| webhook-action plugin | `plugins/webhook-action/src/tool.ts` (`hmacSign` / `executeWebhookPost`) | `X-Atlas-Signature: <hex>` over raw body (Stripe/GitHub-style) | ❌ none | ✅ configurable, [250/1k/4k]ms, 30s timeout |
| SLA alerter | `ee/src/sla/alerting.ts` (`deliverAlert`) | ❌ **unsigned** | ❌ none | ❌ **single fetch, no retry** ← the real gap |

The inbound `@useatlas/webhook` plugin (`plugins/webhook/src/replay.ts`) verifies the sub-processor's format (`${ts}:${body}`, timestamp window, tolerates the `sha256=` prefix) — so the **timestamped convention is already the house standard** on both the send (#1) and receive sides.

**Headline finding:** the SLA alerter ships **unsigned and unretried** — a customer can't verify the alert originated from Atlas, and a single network blip silently drops it. The extraction is the natural place to fix this, not just DRY cleanup.

## Scope (full extraction)

[...]

PR #3118, opened 2026-06-02
feat(webhook-publisher): add @useatlas/webhook-publisher package (#2016)

msywulak:
**PR 1 of 3 for #2016** — extract a single outbound-webhook primitive and unify Atlas's three divergent senders. This PR ships **the package only**; no consumer is migrated yet.

## Why package-first (publish-then-migrate)

`prepare-templates.sh` copies `packages/api/src/lib` (Step 3) and `ee/src` (Step 5c) into the scaffold templates, and the scaffold installs `@useatlas/*` from the npm registry. Migrating the server consumers in the same PR would reference `@useatlas/webhook-publisher` before it exists on npm and break Deploy-Validation's scaffold smoke tests (`check-published-symbols.ts` guards exactly this). So the sequence is:

1. **This PR** — package source + tests + `publish.yml` entry. Merge.
2. **Tag** `webhook-publisher-v0.0.1` → OIDC publish workflow → on npm.
3. **PR 2** — migrate the two server consumers (sub-processor + SLA alerter), pin `0.0.1` in scaffold templates, docs + `ATLAS_SLA_WEBHOOK_SECRET`.
4. **PR 3** — migrate the `webhook-action` plugin (keep `X-Atlas-Signature` wire format + `hmacSign`/`executeWebhookPost` re-exports).

## What's in this PR

New `packages/webhook-publisher/` (mirrors `packages/plugin-sdk` layout — `dist/` build, exports map, MIT, OIDC publish). **Framework-free**: no Effect, no logger, no validation dep — just `fetch` + `node:crypto`.

- **`deliverWebhook({ url, payload, sign, retry, timeoutMs, fetcher?, sleep?, onFailedAttempt? })`** → tagged `DeliveryOutcome` (`ok` | `http_error` | `transport_error`). `4xx` is permanent (no retry, captures a bounded body excerpt); `5xx` / transport / timeout are transient and retried per an **explicit** `{ maxAttempts, delaysMs }` policy. Per-attempt `AbortController` timeout. Never throws for an expected delivery failure — callers map the outcome to their own convention (server logs it; the plugin will re-throw on non-ok).
- **Two signing strategies** (`src/sign.ts`), both byte-locked by golden vectors:
  - `timestamped` → `X-Webhook-Signature: sha256=<hmac(\`${ts}:${body}\`)>` + `X-Webhook-Timestamp` — the house standard, matches the inbound `@useatlas/webhook` verifier (`verifyHmacWithTimestamp`) and the documented customer verify-helper.
  - `rawBody` → `X-Atlas-Signature: <hmac(rawBody)>` — Stripe/GitHub style, kept exactly for the webhook-action plugin.
- **`cappedExponentialDelays()`** helper for capped exponential schedules.
- `onFailedAttempt` observation hook so each consumer emits retry breadcrumbs through its own logger (structured on the server, `console.warn` in the plugin).

## Tests

27 unit tests — golden signature vectors for both strategies (incl. cross-check against the inbound verify algorithm), and retry/backoff/4xx-vs-5xx/timeout/`onFailedAttempt` via injected `fetcher` + `sleep` seams (no wall-clock waits).

## CI

Local gates all green: lint, type, full test (package runs in-suite), syncpack, template-drift, security-headers, railway-watch, schema-drift, oauth-helper-drift, test-discipline, twenty-resolver, published-symbols.

Refs #2016

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/3118"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783025742&installation_id=135337507&pr_number=3118&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F3118&signature=a9624d5f50a516c32811b0467d9dd8d7ac0ef22376646fab57a68d245aa73ce1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added `@useatlas/webhook-publisher`: reliable outbound webhook delivery with pluggable HMAC signing, configurable retry/backoff and per-attempt timeouts.
* **Documentation**
  * Added usage, signing examples, retry semantics, failure classifications, and install/usage guidance.
* **Tests**
  * Comprehensive tests for signing, delivery behavior, retries, backoff, timeouts, and failure handling.
* **Chores**
  * CI updated to publish on matching tags; package metadata, license, ignore rules, build config, Docker contexts, and test script filters adjusted.


msywulak:
Thanks for the reviews — addressed in `8a3a551` + `a889da0`:

**Codex P2 — `timestamped` not inbound-verifiable / docs over-claim** (+ flagged by in-house review): the wire format is **intentionally** `sha256=`-prefixed because it must be byte-identical to the existing sub-processor sender — that's the whole point of the extraction (the issue requires "sub-processor byte-identical"). So this is a docs-accuracy fix, not a format change: corrected the module doc + JSDoc in `sign.ts`, the README, and the interop test comment to state that the `${ts}:${body}` signing input is shared with the inbound `@useatlas/webhook` verifier, but that verifier compares against **bare hex**, so receivers strip the `sha256=` prefix first (exactly what the documented customer verify-helper does). The inbound plugin serves a different direction (Zapier/Make/n8n inbound), so this isn't a broken Atlas-to-Atlas integration — just a previously-imprecise claim.

**Codex P2 — timestamp captured at factory time:** fixed — `timestamped()` now computes the timestamp inside the returned signer (at sign time). `deliverWebhook` signs once per delivery and reuses the result across retries, so byte-stability *within* a delivery is preserved, but a cached/reused signer now gets a fresh timestamp per delivery (no more stale-timestamp replay-window failures). Added a clock-mocked test that proves a reused signer emits a fresh timestamp per invocation.

**Codex P2 — unbounded 4xx body read:** fixed — `readExcerpt` now reads at most ~one chunk past the 200-char limit from the response stream and cancels the rest, so a hostile receiver returning a huge 4xx body can't force an unbounded allocation (falls back to `text()` when there's no stream).

Also added 4 tests from the in-house pass: `timestamped`-through-`deliverWebhook` byte-stability across retries, 5xx-with-body proves the excerpt read is skipped for transient failures, unreadable-body 4xx stays a tagged `http_error`, and full `FailedAttempt`-shape assertions for http + transport. 33 unit tests total; local gates (lint/type/test/syncpack/drift/published-symbols) green.

msywulak:
Round 2 addressed in `6a8f64a`:

- **CodeRabbit (deliver.ts) — guard `onFailedAttempt`:** done. Wrapped the observer in try/catch so a throwing callback can't violate the no-throw-on-delivery contract (surfaced at `console.debug`); added a test proving a throwing observer still yields a tagged `http_error`.
- **CodeRabbit (README) — unlabeled fences:** done, both wire-format blocks now `text`.
- **CodeRabbit (sign.ts:85) — "compute the timestamp when signing":** already fixed in `a889da0` — the timestamp is computed *inside* the returned signer (`sign.ts:70-71`), not at construction. The suggested diff matches the current code, so no change needed.
- **Codex (sign.ts) — "emit an inbound-verifiable signature":** intentionally **not** changing the wire format. The `sha256=`-prefixed `X-Webhook-Signature` is byte-identical to Atlas's existing sub-processor sender, and PR 2 migrates that sender onto this package with a hard "byte-identical" requirement — emitting bare hex here would break that. The inbound `@useatlas/webhook` plugin serves a different direction (Zapier/Make/n8n inbound) and expects bare hex; the `${ts}:${body}` signing input is shared, so a receiver strips the `sha256=` prefix first (the documented customer verify-helper does exactly this). Docs were corrected to state this precisely rather than over-claim direct interop. Making the inbound verifier tolerate the prefix (to enable Atlas-to-Atlas verification without the helper) is a separate, pre-existing enhancement — out of scope for this extraction.

34 unit tests; local gates green (lint/type/test/syncpack/drift/dockerfile-workspace/published-symbols).

PR #3123, opened 2026-06-03
chore(deps): within-major dependency sweep for v0.0.7 (Group A)

msywulak:
## What

Mechanical **within-major** refresh of **66 npm dependencies** across all workspaces, for the **v0.0.7** patch-rollup tag (dependency refresh; no milestone, like v0.0.5).

Generated reproducibly:
```bash
bun update --filter '*' <allowlist derived from `bun outdated` minus Group B majors>
bun x syncpack fix
```
`bun update` (no `--latest`) can't cross a `^` major and respects the 48h `minimumReleaseAge` supply-chain quarantine, so every bump is to the latest **gate-eligible** version within the current major.

## Notable bumps

- **better-auth** 1.6.9 → **1.6.13** (+ all `@better-auth/*` plugins)
- **@ai-sdk/\*** providers (anthropic, openai, gateway, bedrock, provider, react)
- **effect** 3.21.2 · **hono** 4.12.23 · **next** 16.2.6 · **typescript** 6.0.3 · **tailwindcss** 4.3.0
- **zod** 4.4.3 — **coupled** with **@modelcontextprotocol/sdk** 1.29.0 (zod 4.4 needs mcp-sdk ≥1.29, else `packages/mcp` fails `ZodString not assignable to AnySchema`)
- `@hono/zod-openapi`, `@opentelemetry/*`, `@tanstack/react-query`, `react-hook-form`, `mysql2`, `pg`, `jose`, `jsforce`, `snowflake-sdk`, … (full list in the diff)

## One code change

Removes the now-unused `@ts-expect-error` in `packages/web/src/lib/auth/client.ts` — better-auth **1.6.13** fixes the `apiKeyClient()` plugin client type that required the suppression at 1.6.9. (tsgo flags an unused `@ts-expect-error` as an error, so this is mandatory once the bump lands.)

## Held for separate PRs (Group B — major bumps, security-first)

Deliberately **excluded** from this sweep; each needs changelog review / soak:
`@vercel/sandbox` 1→2 (SaaS-pinned prod sandbox) · `just-bash` 2→3 (explore backend) · `stripe` 21→22 (billing) · `react-day-picker` 9→10 · `diff` 8→9 · low-risk/defer: `fumadocs-mdx`, `syncpack`, `shadcn`, `esbuild`, `@duckdb/node-api`.

Verified none crossed a major (lockfile-checked).

## Gates

[...]

PR #3124, opened 2026-06-03
docs(commands): add /deps-update — dependency-refresh workflow

msywulak:
## What

Adds a `/deps-update` slash command (`.claude/commands/deps-update.md`) that captures how dependency refreshes actually work in this repo, so future runs don't re-derive it. Renovate is Docker-only, so npm deps drift until a manual sweep — this documents that sweep end-to-end.

## Why

Distilled from the **v0.0.7 Group A sweep (#3123)**, which hit two failures that local `bun run type` / `bun run test` did **not** catch:
1. **Frozen-lockfile** — `syncpack fix` rewrites ranges, so you must `bun install` before committing or CI's `--frozen-lockfile` fails.
2. **kysely 0.29 break** — better-auth 1.6.13's kysely-adapter imports an export kysely 0.29 removed; only the Standalone Example Build (Turbopack) surfaced it. Fixed by pinning kysely 0.28.17.

Both, plus the `bun update` workspace-cascade gotcha, the coupled-version constraints, and the `rm -rf node_modules` tsc-shim trap, are now encoded so they're a checklist instead of a rediscovery.

## Contents

- Two-group model: within-major batched sweep (Group A, one PR) vs per-package major bumps (Group B, security-first, one PR each).
- Exact `bun outdated` → allowlist → `bun update` → `syncpack fix` → `bun install` sequence.
- Coupled-constraints table (zod↔mcp-sdk, better-auth↔kysely) — add a row when a new one bites.
- Gotchas + Rules sections.

Docs-only (single markdown command file); no code paths touched.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/3124"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783042806&installation_id=135337507&pr_number=3124&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F3124&signature=c05a0fa950994374d807e3020aaa804515d60ea8405fa47120cd7e1ffec5971d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive dependency-refresh runbook describing a strict two-stage workflow: a batched minor/patch sweep and one-PR-per-package major upgrades with staged validation.
  * Documents verification steps and CI parity gates (including lockfile and dedicated build checks), guidance for handling coupled-version constraints, operational gotchas to avoid risky update patterns, and merge/release tagging rules for patch rollups.

### DIFF
--- a/.claude/commands/deps-update.md
+++ b/.claude/commands/deps-update.md
@@ -1,0 +1,101 @@
+Refresh npm dependencies across the monorepo. Renovate is **Docker-only** (`renovate.json` → `enabledManagers: ["dockerfile"]`), so npm deps drift until a manual sweep. This command does the sweep the way it actually works in this repo — including the gotchas that only CI (not local type/test) catches. Ships as a **patch-rollup tag** (`v0.0.x`, no milestone — like v0.0.5), per [ADR-0008](../../docs/adr/0008-versioning-and-release-tags.md).
+
+Two-group model — **never mix them in one PR**:
+
+- **Group A — within-major sweep** (minor/patch). One batched PR. `bun update` can't cross a `^` major, so a single allowlisted sweep is safe to review as a unit.
+- **Group B — major bumps.** One PR **per package**, security-sensitive first. Each needs changelog/migration reading and (for sandbox/billing/auth) a staging soak. Held out of the Group A sweep by an explicit exclude list.
+
+---
+
+## Group A — within-major sweep
+
+**Step 1: Branch.**
+```bash
+git checkout main && git pull && git checkout -b chore/deps-group-a-<tag>   # e.g. chore/deps-group-a-v0.0.7
+```
+
+**Step 2: Derive the allowlist** — every outdated direct dep, minus the Group B majors. `bun outdated` only lists **direct** deps (transitives like `kysely` won't appear — see coupled-bump table).
+```bash
+EXCLUDE='^(stripe|react-day-picker|diff|just-bash|@vercel/sandbox|fumadocs-mdx|syncpack|esbuild|shadcn|@duckdb/node-api|chat|@chat-adapter/.*)$'
+bun outdated --filter '*' 2>/dev/null \
+  | grep -E '^\| ' | grep -vE '^\| Package' | grep -vE '\(peer\)|\(optional\)' \
+  | sed -E 's/^\| +//; s/ +\|.*$//; s/ \(dev\)//' \
+  | sort -u | grep -vE "$EXCLUDE" > /tmp/groupA.txt
+wc -l /tmp/groupA.txt && cat /tmp/groupA.txt   # eyeball it
+```
+Re-check the EXCLUDE list against `bun outdated` each time — when a Group B major lands, drop it from EXCLUDE; when a new major appears, add it.
+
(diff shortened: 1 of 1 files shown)


msywulak:
Thanks @chatgpt-codex-connector — addressed in `66ca25f4`:

**P2 #2 (Group B safe-patch exclusion) — ✅ fixed.** Good catch. Added a "coverage gap to own" callout to the Group B section: excluding a package from the within-major sweep also drops its safe within-major patches, so each Group B PR should baseline to latest-within-major and a within-major advisory on a held package gets a fast-track PR.

**P2 #3 (partial CI gate) — ✅ fixed.** Step 6 now leans on the **full `/ci`** (which is exactly where the template-drift / published-symbols / lint checks that catch sweep mistakes live), with the frozen-lockfile check and the standalone Turbopack build called out as the only two extras `/ci` doesn't cover.

**P2 #1 (`--filter '*'` unsupported) — ❌ respectfully declining; verified false on the pinned bun.** `engines.bun` pins `>=1.3.13 <1.3.14`, and on `bun@1.3.13`:
- `bun update --filter '*' hono` and `bun update --filter '*'` both exit 0 — no "unrecognised dependency format: *" error. (The v0.0.7 sweep, #3123, ran this exact form successfully across 66 packages.)
- `--filter '*'` is **not** redundant — it's required for the cross-workspace floor cascade. I downgraded `packages/api`'s `hono` floor to `^4.0.0` and ran `bun update hono` (no `--filter`): the sub-workspace floor stayed `^4.0.0`. Only `--filter '*'` rewrites sub-workspace `package.json` `^` floors; without it you get lockfile-only resolution. So dropping it would silently regress the sweep.

I clarified in the doc *why* both the filter and the names are load-bearing, and noted to re-test if a future bun changes `bun update`'s flag handling.

msywulak:
@coderabbitai — verified against current code and addressed in `181a20b5`, though not as written.

The premise is inverted: the **actual `/ci` command** (`.claude/commands/ci.md`) *does* run those checks — `check-published-symbols.ts`, `check-security-headers-drift.sh`, `check-schema-drift.sh`, `check-oauth-helper-drift.sh`, `check-test-discipline.sh`, `check-twenty-resolver-imports.sh`. The "lint + type + test + syncpack + template drift + railway-watch" contract you cited is the **CLAUDE.md one-liner**, which is the *stale, under-listing* artifact — not the source of truth. So trimming Step 6 to those six would make this doc **less** accurate, not more.

I took your *alternative* suggestion instead: dropped the inline enumeration and pointed Step 6 at `ci.md` as the source of truth, so it can't drift. That fixes the operator-confusion concern without asserting a false smaller gate.

(Real follow-up surfaced here, out of scope for this PR: the CLAUDE.md `/ci` summary should be updated to match `ci.md`'s actual check set — it's the doc that's genuinely drifted.)

PR #3228, opened 2026-06-05
test(dashboards): de-flake param-sync URL round-trip (one live write per test)

msywulak:
## Problem

`main` CI went red on `test-others` after #3227 — a flaky failure in `cross-filter-sync.test.tsx`:

```
✗ drill applies a filter, a second composes (AND)…
  TestingLibraryElementError: Unable to find an element by: [data-testid="filter-chip-stage"]
```

Passed in isolation, flaked **7/40 (~17%)** locally. It only surfaces under CI load, where the suite runs slower/contended.

## Root cause

The test drove **two back-to-back synthetic `fireEvent.click`s** (drill `region`, then drill `stage`) through a hand-rolled self-feeding nuqs testing adapter:

```tsx
const [search, setSearch] = useState(() => new URLSearchParams(initial));
<NuqsTestingAdapter searchParams={search} onUrlUpdate={(e) => setSearch(e.searchParams)} />
```

Two failure modes stack here:

1. **Override clobber** — each write composes on the `searchParams` **prop**, which only updates after React commits `setSearch`; nuqs's optimistic cache updates immediately. The second write reads the *stale* prop, so the `stage` drill clobbers the `region` override.
2. **Queue drop** — even with a synchronous store, nuqs's update queue can drop the **second** of two rapid synthetic writes before the first flushes; the compose step then composes on the first write and the second chip never appears.

Real drilldowns are user-paced seconds apart, so neither race exists outside the synthetic back-to-back test clicks. The single-write sibling `drilldown-sync` has always been stable for exactly this reason.

## Fix

**Drive exactly one live write per test.** Each test seeds the *prior* filter state into the URL via `withNuqsTestingAdapter({ searchParams, hasMemory: true })` and drives a single drill / chip-remove / clear-all, then asserts. With no two back-to-back writes, neither the clobber nor the queue-drop can occur — robust by construction, not by timing. Same contract coverage (apply / compose-AND / remove-one / toggle-deselect / clear-all / reload), now across 6 focused tests.

`hasMemory` (NuqsTestingAdapter's documented memory-backed mode) gives the faithful Next.js round-trip a bare initial-frozen adapter doesn't — writes land in nuqs's own synchronous store and re-render every hook subscribed to the key. The sibling `drilldown-sync` is modernized onto the same `hasMemory` adapter (single write, already stable).

## Validation

| | before | after |
|---|---|---|
| cross-filter-sync | 7/40 fail (17.5%) | **0/60 sequential, 0/16 under parallel CPU load** |

`eslint` clean, `tsgo --noEmit` clean. Test-only change; no product code touched.

### DIFF
--- a/packages/web/src/app/(workspace)/dashboards/[id]/__tests__/cross-filter-sync.test.tsx
+++ b/packages/web/src/app/(workspace)/dashboards/[id]/__tests__/cross-filter-sync.test.tsx
@@ -15,7 +15,6 @@
  * we pin the URL-state contract those two build on.
  */
 import { afterEach, describe, expect, test, mock } from "bun:test";
-import { useState } from "react";
 import { useQueryState } from "nuqs";
 import { NuqsTestingAdapter } from "nuqs/adapters/testing";
 import { render, cleanup, fireEvent, waitFor, screen } from "@testing-library/react";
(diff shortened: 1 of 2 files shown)


msywulak:
⚠️ **Converting to draft — partial fix.** CI shows `hasMemory` does eliminate the *override clobber* (the first filter is now correctly retained), but a **residual race remains**: under the slower CI runner the rapid **second** `fireEvent.click` (the compose step) gets dropped by nuqs's update queue while the first write is still mid-flush — the DOM freezes at exactly `{region: "us"}` with no `stage` chip. So this turns a ~17% clobber-flake into a lower-rate drop-flake, not a clean fix.

I can't reproduce the failure locally (0 fails in 100+ runs, even under 8× parallel CPU load — my machine is too fast to expose the GitHub-runner timing), so I can't pre-validate a deeper fix.

Signal for the real fix: the sibling `drilldown-sync` test issues a **single** write and is rock-stable; this suite issues **two rapid** writes and flakes. The robust fix is to stop driving two back-to-back live writes — e.g. seed the prior filter(s) via the URL and drive **one** live interaction per assertion (compose / remove / toggle-deselect), matching the single-write pattern that's already stable. Holding for a decision on approach before pushing further.

msywulak:
✅ **Updated approach — restructured to single-write-per-assertion (robust by construction).**

Root cause confirmed: nuqs's update queue drops the second of two *rapid* synthetic writes under slow CI before the first flushes; the compose step composes on the first, so it broke. `hasMemory` fixed the *clobber* but not this drop, and it can't be reproduced locally (0 fails in 100+ runs even under 8× parallel CPU load).

The fix drives **exactly one live write per test** — each test seeds the prior filter(s) via the URL (`withNuqsTestingAdapter` + `hasMemory`) and asserts a single drill / chip-remove / clear-all. No two back-to-back writes ⇒ the queue drop can't occur, the same reason the single-write sibling `drilldown-sync` has always been stable. Same contract coverage (apply / compose-AND / remove-one / toggle-deselect / clear-all / reload), now across 6 focused tests.

Validation: **0/60 sequential + 0/16 under parallel CPU load**, eslint + tsgo clean. Marking ready for review.

Issue #3240, opened 2026-06-06
feat(api): recognize semantic/groups/<group>/ for metrics, glossary & catalog discovery


msywulak:
Incidental finding while implementing #3232 (group-scoped **entity** loader, ADR-0012).

#3232 taught the **entity** read side about the canonical `semantic/groups/<group>/entities/` namespace (and added `groups` to `RESERVED_DIRS`). The sibling read paths for **metrics, glossary, and catalog** were left out of scope and do **not** yet recognize the namespace:

- `discoverMetrics` / `loadGlossary` in `packages/api/src/api/routes/admin.ts` scan only `root/<source>/metrics` and `root/<source>/glossary.yml` (local `RESERVED_DIRS = {entities, metrics}`), so `groups/<group>/metrics/` and `groups/<group>/glossary.yml` are never found.
- `loadMetricsFromDir` / glossary loaders in `packages/api/src/lib/semantic/lookups.ts` and the per-source scans in `packages/api/src/lib/semantic/search.ts` import scanner `RESERVED_DIRS` — which now includes `groups` — so they **skip** the `groups/` dir entirely rather than descending into `groups/<group>/{metrics/,glossary.yml}`.

Net: a non-default group's `entities/` load and partition correctly, but its `metrics/`, `glossary.yml`, and `catalog.yml` are silently invisible. Single-DB (flat root) and legacy `<source>/` layouts are unaffected.

This is a natural seam — #3232 was scoped to the entity loader/whitelist — but it should be closed before the grouped-tree view (#3235) and the generator (#3234) write metrics/glossary into the group namespace, or those artifacts won't render/load.

## Key files
- `packages/api/src/lib/semantic/scanner.ts` — consider a shared `getGroupDirs(root, subdir)` so metrics/glossary discovery reuses the same `flat | groups/<g> | legacy <source>` traversal as `getEntityDirs` (single source of truth for the layout).
- `packages/api/src/api/routes/admin.ts` — `discoverMetrics`, `loadGlossary` (+ their local `RESERVED_DIRS`).
- `packages/api/src/lib/semantic/lookups.ts` — metric/glossary per-source scans.
- `packages/api/src/lib/semantic/search.ts` — per-source scans (lines ~426/484).

## Acceptance criteria
- [ ] `semantic/groups/<group>/metrics/*.yml` are discovered and attributed to group `<group>`.
[...]

PR #3251, opened 2026-06-06
fix(api): fail-closed semantic whitelist on directory-scan failure (#3243)

### LINKED ISSUE #3243: fix(api): semantic whitelist should fail-closed (not drop to shared-default) when a directory scan fails
Surfaced by the silent-failure review of #3241 (group-scoped loader). Pre-existing class — applies to both the legacy `<source>/` scan and the new `groups/` scan added in #3232.

## Problem
When `fs.readdirSync` throws inside `getEntityDirs` (EACCES, ENOTDIR on a symlinked dir, EMFILE/ENFILE under fd pressure, EIO/network-mount stalls — all *distinct* from "directory absent", which is already guarded by `existsSync`), the catch sets `rootScanFailed = true` and the affected group's entities silently drop out of `byConnection`. Two bad downstream effects in `whitelist.ts`:

1. **Mis-attributed empty whitelist.** `getWhitelistedTables` emits *"No entities found for connection — whitelist is empty"*, blaming the connection rather than the scan failure. An operator concludes the group was never configured when its YAMLs exist on disk but were locked out by a transient/permission error.
2. **Silent drop to shared-default mode (the security-relevant one).** Partition mode is decided by `hasNonDefaultConnection = byConnection.keys().some(k => k !== "default")`. If the scan that *would* have produced the non-default group fails and only `default` remains, the system falls into **shared-whitelist mode** — every connection silently validates against the *default* group's tables. A query that should have been restricted to (or rejected by) the `warehouse` group's whitelist instead validates against `default`. The table whitelist is a SELECT-only **security boundary** (CLAUDE.md: "prefer errors over silent fallbacks" on security checks), so a swallowed FS error changing which whitelist a connection is validated against is a fail-toward-widening.

Note: this is **fail-closed at the point of enforcement** (`tools/sql.ts` rejects tables not in the set), so it cannot widen *beyond* the default group — but it can validate against the *wrong* (default) group instead of an empty/correct one.

## Proposed fix
- Thread `rootScanFailed` out of `getTablesByConnection`/`loadTablesByConnection` into `getWhitelistedTables`.
- When a scan failed and a group resolves to an empty set, **keep it empty (fail-closed)** with an accurate "semantic layer scan failed — load incomplete" message; do **not** fall back to shared-default mode.
- Consider splitting `rootScanFailed` into `groupsScanFailed` / `legacyScanFailed` (or a `failedScans: string[]`) so the escalated error names which namespace failed (today both collapse into one boolean logged as the generic "semantic root" message).

## Key files
- `packages/api/src/lib/semantic/scanner.ts` — `getEntityDirs` (two catch sites → one boolean).
- `packages/api/src/lib/semantic/whitelist.ts` — `loadTablesByConnection` / `getTablesByConnection` / `getWhitelistedTables` (partition decision + empty-whitelist message).

## Acceptance criteria
- [ ] A failed directory scan does not silently drop to shared-default mode; affected groups fail-closed.
- [ ] The empty-whitelist log distinguishes "scan failed (incomplete)" from "no entities configured".
- [ ] Escalated error names which scan (groups/ vs legacy) failed.
- [ ] Tests simulate a scan failure (e.g. unreadable dir) and assert fail-closed + accurate messaging.

## Dependencies
- Builds on #3232.

msywulak:
Closes #3243.

## Problem

The table whitelist is a SELECT-only **security boundary**. When `fs.readdirSync` throws inside `getEntityDirs` (EACCES, ENOTDIR on a symlinked dir, EMFILE/ENFILE under fd pressure, EIO/stalled network mount — all **distinct** from "directory absent", which `existsSync` already guards), the catch set `rootScanFailed = true` and the affected group's entities silently dropped out of `byConnection`. Two bad downstream effects in `whitelist.ts`:

1. **Mis-attributed empty whitelist.** The log blamed the connection ("No entities found for connection") rather than the scan failure — an operator concludes the group was never configured when its YAMLs exist on disk but were locked out by a transient/permission error.
2. **Silent drop to shared-default mode (the security-relevant one).** Partition mode is keyed on `hasNonDefaultConnection`. If the scan that *would* have produced the non-default group fails and only `default` remains, the system fell into shared-whitelist mode — a connection that should be restricted to (or rejected by) the `warehouse` group's whitelist instead validated against `default`. A swallowed FS error changing which whitelist a connection is validated against is **fail-toward-widening**.

> It still fails closed *at enforcement* (`tools/sql.ts` rejects unlisted tables), so it cannot widen *beyond* default — but validating against the *wrong* (default) group instead of an empty/correct one is the bug.

## Fix

- **`scanner.ts`** — replace the single `rootScanFailed` boolean with a typed `failedScans: ScanNamespace[]` (`"groups" | "legacy"`). Both catch sites push their namespace, so the escalated error can name **which** scan failed. Kept well-typed and minimal so the queued #3240/#3245 scanner work can build on it without churn.
- **`whitelist.ts`** — thread `failedScans` out of `loadTablesByConnection` / `getTablesByConnection` into a new shared `resolveGroupTables` helper (de-duplicates the live + cached decision paths). When **any** scan failed, never drop to shared-default mode: the requested group resolves to its own (possibly empty) set and **fails closed**. The empty-whitelist log now distinguishes **"scan failed (incomplete)"** (error, names the namespace) from **"no entities configured"** (warn) — different operator situations. The scan-failure signal is cached alongside `_tablesByConnection` so the decision stays consistent with the map it was built from.

`default` is unaffected by a `groups/` scan failure (it still serves its own tables); only the groups whose scan failed fail closed.

## Tests

- **`group-scoped-loader.test.ts`** — simulate a real FS error (EACCES) via `spyOn(fs, "readdirSync")` (the established pattern from `explore-backend.test.ts`; distinct from "absent"). Asserts: (a) the affected group fails **closed** (empty set, **not** default's tables); (b) does **not** drop to shared-default mode; (c) the escalated error **names** the failed namespace and is distinct from the no-entities warning; (d) `getEntityDirs` reports `failedScans` correctly for both `groups` and `legacy`; (e) a *successful* scan with a genuinely empty group keeps today's benign warning (no behavior change when scans succeed).
- **`sql-connection-whitelist.test.ts`** — a **registered** connection whose scan failed (empty whitelist) has its query **REJECTED**, proving the enforcement layer does not validate it against `default`.

RED confirmed before merge: with the fail-closed guard reverted to the old "drop to default" behavior, exactly the 3 fail-closed behavioral tests fail (the group inherits `default`'s `orders`, drops to shared mode, emits the wrong log); the scanner-threading and benign-no-entities tests correctly still pass.

## Acceptance criteria

- [x] A failed directory scan does not silently drop to shared-default mode; affected groups fail-closed.
- [x] The empty-whitelist log distinguishes "scan failed (incomplete)" from "no entities configured".
- [x] Escalated error names which scan (groups/ vs legacy) failed.
- [x] Tests simulate a scan failure (FS error) and assert fail-closed + accurate messaging.

## CI

Local gates all green: lint, type (full monorepo), full test suite, syncpack, template/security-headers/railway/schema/oauth-helper drift, test-discipline, twenty-resolver, published-symbols.

This is an **architecture-labeled** fix (module-boundary hardening); an architecture-wins entry will be recorded after merge.

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/3251"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783349600&installation_id=135337507&pr_number=3251&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F3251&signature=bd69dd2dff842c58d8bcd4615ccd753e2a8c0cab561c2c7d1714cb311472ab48"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of semantic directory scan failures with more granular error tracking and a fail-closed security strategy that restricts database access when scans encounter filesystem errors, preventing potential exposure of unvetted tables.
  * Enhanced error logging to clearly identify which scan operations failed during directory traversal.

* **Tests**
  * Added comprehensive tests verifying fail-closed behavior and error reporting when directory scans encounter permission issues.

### DIFF
--- a/packages/api/src/lib/semantic/__tests__/group-scoped-loader.test.ts
+++ b/packages/api/src/lib/semantic/__tests__/group-scoped-loader.test.ts
@@ -14,9 +14,10 @@
  * observable — see `docs/development/testing.md` and the existing
  * `config-deploy-mode-warning.test.ts` for the pattern.
  */
-import { describe, it, expect, beforeEach, afterEach, mock } from "bun:test";
+import { describe, it, expect, beforeEach, afterEach, mock, spyOn } from "bun:test";
 import { resolve, join } from "path";
 import { mkdirSync, writeFileSync, rmSync, existsSync } from "fs";
+import * as fs from "fs";
 
 type LogCall = { level: "error" | "warn" | "info" | "debug"; payload: unknown; message: string };
 const logCalls: LogCall[] = [];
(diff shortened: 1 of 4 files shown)


msywulak:
@codex review — this is a security-sensitive fix-closed change to the SQL table whitelist (a SELECT-only security boundary). Please scrutinize the fail-closed decision in `resolveGroupTables` and the scan-failure threading in `scanner.ts`/`whitelist.ts` for any path that could still drop to shared-default mode or validate against the wrong group.

Issue #3267, opened 2026-06-06
feat: ES/OpenSearch datasource — Query DSL tool (queryElasticsearch) + read-only validator


msywulak:
## Parent
#3259

## What to build
The second query surface: a dedicated `queryElasticsearch` tool for full-text/relevance and deeply-nested aggregation questions the SQL API can't express. Registered in the plugin's `initialize()` (mirroring how Salesforce registers `querySalesforce`). A **custom read-only DSL validator** accepts only read request shapes (`_search`, `_count`, `_msearch`, `_field_caps`, `_mapping`, read-only `_cat`) and **default-denies** everything mutating/administrative (`_bulk`, index/update/delete, `_delete_by_query`, `_update_by_query`, mutating scripts); the index whitelist is enforced in the tool. The **response normalizer is extended** to flatten `hits.hits[]._source` and the `aggregations` tree into `{ columns, rows }`. A size cap + `terminate_after`/timeout apply. The agent system prompt gains DSL guidance (when to prefer DSL over SQL).

## Acceptance criteria
- [ ] `queryElasticsearch` tool registered and callable by the agent
- [ ] DSL validator allows only read request shapes; default-denies all writes/admin shapes; adversarial cases (write smuggled into a search body) rejected — unit-tested
- [ ] Index whitelist enforced in the tool
- [ ] Normalizer flattens `hits` and `aggregations` (bucket + metric aggs) to `{ columns, rows }`; unit-tested
- [ ] Size cap + timeout applied to every DSL request
- [ ] DSL dialect guidance present in the agent system prompt
- [ ] End-to-end: a full-text and an aggregation question return correct tables

## Blocked by
- #3262

PR #3312, opened 2026-06-07
fix(plugins): enforce per-index/object semantic whitelist in ES Query DSL & SOQL tools (#3307)

### LINKED ISSUE #3307: fix: plugin tool object/index whitelist uses connection IDs, not semantic table names
## Summary

`AtlasPluginContext.connections.list()` returns registered **connection IDs**, not semantic-layer **table/index names** — but the plugin-registered query tools use it as their object/index whitelist source. So the whitelist's *membership* tier is ineffective for those tools.

Wired in `packages/api/src/api/server.ts`:

```ts
connections: {
  get: (id) => connections.get(id),
  list: () => connections.list(), // ConnectionRegistry.list() → Array.from(entries.keys()) = connection IDs
},
```

Both plugin tools source their whitelist from this:

- `plugins/salesforce/src/index.ts` → `getWhitelist: () => new Set(ctx.connections.list())` (feeds `validateSOQL`'s object whitelist)
- `plugins/elasticsearch/src/index.ts` (#3267) → `getWhitelist: () => new Set(ctx.connections.list())` (feeds `validateIndexAccess`'s index whitelist)

In static-datasource mode the set therefore contains values like `["salesforce-datasource"]` / `["elasticsearch-datasource"]`, never the real object/index names. The membership check then either over-rejects every legitimate object/index, or — where the tool falls back to structural-only on an "empty-ish" set — silently allows any named one. Salesforce hasn't surfaced this because it is OAuth/adapter-only now (the static `querySalesforce` path is legacy); Elasticsearch's static path is a first-class self-host case.

ES mitigates today with always-on structural rails (no wildcards / `_all` / system indices), so cross-index fan-out is still blocked — but the per-name semantic-layer whitelist is not actually enforced from the plugin context.

## Key files
- `packages/api/src/api/server.ts` — plugin context `connections` wiring
- `packages/plugin-sdk/src/types.ts` — `AtlasPluginContext` shape
- `plugins/salesforce/src/index.ts`, `plugins/elasticsearch/src/index.ts` — whitelist consumers
- `packages/api/src/lib/semantic/whitelist.ts` — `getWhitelistedTables(connectionId)` (the correct source, core-only today)

## Acceptance criteria
- [ ] The plugin context exposes a semantic-table accessor (e.g. `ctx.semantic.getTables(connectionId)` / `ctx.connections.tables(id)`) returning the semantic-layer entity/table names for a connection.
- [ ] `querySalesforce` and `queryElasticsearch` enforce their object/index whitelist against that accessor instead of `ctx.connections.list()`.
- [ ] Unit test: a query against an object/index NOT in the semantic layer is rejected in static mode (production-representative whitelist).

## Dependencies
- Surfaced by #3267 (ES Query DSL tool). Affects the Salesforce `querySalesforce` tool too.

msywulak:
Closes #3307.

## Problem
The plugin-registered query tools enforced an *ineffective* object/index membership whitelist in **static-datasource (self-host) mode**:

- **`queryElasticsearch`** (ES Query DSL, #3267) passed an **empty set** → `validateIndexAccess` ran in structural-only mode, so **any explicitly-named, non-system index was queryable** — including indices with no entity YAML.
- **`querySalesforce`** (SOQL) passed **`ctx.connections.list()`** → registered **connection IDs** (e.g. `salesforce-datasource`), never object names like `Account`, so `validateSOQL` either over-rejected every real query or fell back to structural-only.

Net: the semantic layer is the access boundary the **SQL path (`executeSQL`) enforces**, but the bespoke DSL/SOQL surfaces onto the *same cluster* did **not** honor it. This is read-only and bounded (always-on structural rails block wildcards / `_all` / system indices; endpoints are default-deny read-only), so it's a **confidentiality / defense-in-depth gap**, not a write/escalation hole — but it contradicts the documented invariant that only `semantic/entities/*.yml` tables are queryable.

## Fix
Add a `connections.tables(id)` accessor to the plugin context returning the semantic-layer entity names for a connection — the **same source `executeSQL` validates against** (`getWhitelistedTables`). Both tools whitelist against it, so the DSL/SOQL surfaces enforce the **identical per-object boundary** as the SQL path.

| Layer | Change |
|---|---|
| `@useatlas/plugin-sdk` `AtlasPluginContext` + internal `PluginContextLike` | new **required** `connections.tables(id): string[]` |
| `api/server.ts` | wires `tables: (id) => Array.from(getWhitelistedTables(id))` |
| `plugins/elasticsearch`, `plugins/salesforce` | `getWhitelist` reads `ctx.connections.tables(<plugin id>)` — the id the static connection registers under (`registerDirect(plugin.id, …)`), matching the SQL whitelist key |
| `api/plugins/mcp-boot.ts` | stays empty — tool registration is a **no-op** on that standalone path, so the bespoke tools are never served there; the real whitelist-backed context is the Hono boot's (comment added) |

Empty semantic layer → `tables()` returns `[]` → validators fall back to their always-on structural rails (unchanged).

## Tests
- **ES** (`dsl-tool.test.ts`): proves `tables()` (not `list()`) drives the whitelist — a non-member index is rejected, the **connection id itself** (what `list()` returned pre-fix) is rejected, and a semantic-layer index passes the gate.
- **SF** (`salesforce.test.ts`): a non-member object (`Contact` when `tables()=[Account]`) is rejected via the plugin's `initialize` wiring; fixed the pre-existing "whitelist load fails" test to inject failure on `tables()` (was `list()`).
- Mechanical: `tables: () => []` added to plugin-context mocks across plugin/api tests for the new required field; `createMockContext` helper updated.

## Notes
- Broad test touch is from the new **required** context field (deliberate — the type now guarantees every context supplies a real whitelist source).
- The 16 pre-existing `@typescript-eslint/no-unsafe-function-type` lint findings in `plugins/snowflake/__tests__` are unrelated and predate this branch (CI lint scope is `packages/ ee/`; `plugins/` test files aren't gated). Not touched.
- ✅ `bun run type` clean · ✅ `packages/` lint clean · ✅ affected ES/SF/api plugin tests green

---
<a href="https://app.blacksmith.sh/AtlasDevHQ/codesmith/atlas/pr/3312"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783440326&installation_id=135337507&pr_number=3312&repository=AtlasDevHQ%2Fatlas&return_to=https%3A%2F%2Fgithub.com%2FAtlasDevHQ%2Fatlas%2Fpull%2F3312&signature=5cb64f33f3047f74f5bd7a23f79ed3a503bf3fdae4484b540ea37335d7f71aae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Query tools now enforce per-index/object membership against the semantic-layer whitelist; when no semantic layer exists the structural validation remains.
  * Whitelist lookups now fail closed (errors surface instead of falling back to an empty whitelist).

* **Documentation**
  * Updated Elasticsearch query docs to describe per-index membership enforcement and resource bounding behavior.

* **Tests**
  * Expanded tests and mocks to cover semantic-layer whitelist validation across multiple plugins.

### DIFF
--- a/apps/docs/content/docs/plugins/datasources/elasticsearch.mdx
+++ b/apps/docs/content/docs/plugins/datasources/elasticsearch.mdx
@@ -151,12 +151,14 @@ and an optional Query DSL `body`.
 The surface is gated by a **default-deny** read-only validator: only read
 endpoints are allowed; mutating/administrative endpoints, smuggled bulk-write
 actions, mutating (`ctx`-referencing) scripts, and wildcard / `_all` /
-system-index targets are rejected. These structural rails are always enforced;
-per-index *membership* against the semantic-layer index list is **not** yet
-enforced from the DSL tool (the SQL surface enforces the real whitelist via the
-core pipeline) — tracked in [#3307](https://github.com/AtlasDevHQ/atlas/issues/3307).
-Every `_search` is resource-bounded — `size` clamped to `ATLAS_ROW_LIMIT`, a
-search `timeout`, and a non-aggregation `terminate_after`
+system-index targets are rejected. On top of those always-on structural rails,
+the tool enforces **per-index membership** against the semantic layer: the
+queried index must be a profiled entity (the same index/table whitelist
+`executeSQL` validates against), so the DSL surface honors the identical access
+boundary as the SQL surface ([#3307](https://github.com/AtlasDevHQ/atlas/issues/3307)).
+When a connection has no semantic layer the membership check is skipped and only
+the structural rails apply. Every `_search` is resource-bounded — `size` clamped
+to `ATLAS_ROW_LIMIT`, a search `timeout`, and a non-aggregation `terminate_after`
 (`ATLAS_ES_TERMINATE_AFTER`).
 
 The `queryElasticsearch` tool is registered **only** for a static,
--- a/e2e/surfaces/slack-proactive.test.ts
+++ b/e2e/surfaces/slack-proactive.test.ts
@@ -430,7 +430,7 @@ beforeAll(async () => {
 
   await plugin.initialize!({
     db: null,
-    connections: { get: () => { throw new Error("unused"); }, list: () => [] },
+    connections: { get: () => { throw new Error("unused"); }, list: () => [], tables: () => [] },
     tools: { register: () => {} },
     logger: {
       info: () => {},
--- a/packages/api/src/__tests__/plugin-mcp-tools.test.ts
+++ b/packages/api/src/__tests__/plugin-mcp-tools.test.ts
@@ -189,7 +189,7 @@ describe("wireMcpToolPlugins", () => {
     pluginRegistry.register(makePlugin("docs", [makeTool({ name: "list" })]));
     await pluginRegistry.initializeAll({
       db: null,
-      connections: { get: () => ({}), list: () => [] },
+      connections: { get: () => ({}), list: () => [], tables: () => [] },
       tools: { register: () => {} },
       logger: {},
       config: {},
@@ -215,7 +215,7 @@ describe("wireMcpToolPlugins", () => {
     } as PluginLike);
     await pluginRegistry.initializeAll({
       db: null,
-      connections: { get: () => ({}), list: () => [] },
+      connections: { get: () => ({}), list: () => [], tables: () => [] },
       tools: { register: () => {} },
       logger: {},
       config: {},
@@ -236,7 +236,7 @@ describe("wireMcpToolPlugins", () => {
     pluginRegistry.register(makePlugin("good", [makeTool({ name: "ok" })]));
     await pluginRegistry.initializeAll({
       db: null,
-      connections: { get: () => ({}), list: () => [] },
+      connections: { get: () => ({}), list: () => [], tables: () => [] },
       tools: { register: () => {} },
       logger: {},
       config: {},
@@ -262,7 +262,7 @@ describe("wireMcpToolPlugins", () => {
     } as PluginLike);
     await pluginRegistry.initializeAll({
       db: null,
-      connections: { get: () => ({}), list: () => [] },
+      connections: { get: () => ({}), list: () => [], tables: () => [] },
       tools: { register: () => {} },
       logger: {},
       config: {},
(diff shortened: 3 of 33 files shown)


msywulak:
Ran the full in-house review toolkit (code / tests / silent-failure / type-design / comments). Pushed hardening in `75d8968`:

- **SF whitelist now fails CLOSED** — the `getWhitelist` `catch` that returned an empty set (= structural-only = silent widening) is removed; an unloadable whitelist now refuses the query, symmetric with the ES tool. (silent-failure-hunter, HIGH)
- **Stronger tests** — ES positive case asserts `success: true` (was a weak negative match); ES wiring test pins that `tables()` is called with `"elasticsearch-datasource"` (guards the wrong-id → `[]` → structural-only silent downgrade). (test-analyzer / code-reviewer)
- **`tables(id)` returns `readonly string[]`** + doc scoped to self-host and the typo'd-id caveat. (type-design)
- **Comment accuracy** — "same source `executeSQL` validates against" scoped to self-host/static (SaaS uses `getOrgWhitelistedTables`); mcp-boot rationale strengthened (MCP serves `mcpTools()`, not `tools.register`). (comment-analyzer)

**Deferred (tracked in #3313):** fail-CLOSED on semantic-layer *scan failure* for plugin query tools — `getWhitelistedTables` returns `[]` on a scan error, which these tools read as structural-only (the inverse of the SQL path's #3243 deny-all). Distinguishing "no layer" (→ structural-only, intended) from "scan failed" (→ refuse) touches the whitelist core + the legitimate no-layer case, so it's separable. Note this PR already makes ES strictly better than its prior always-empty state.

✅ type-check · `packages/` lint · ES/SF/api plugin tests all green.

Issue #3269, opened 2026-06-06
feat: ES/OpenSearch datasource — index patterns / aliases / data-streams as entities


msywulak:
## Parent
#3259

## What to build
Represent index **patterns** (`logs-*`), **aliases**, and **data-streams** as a single logical entity across the whitelist, the CLI profiler, and both query surfaces — so time-partitioned indices are queryable as one source. Extends the whitelist to match patterns/aliases, the profiler to emit one entity per pattern/alias/data-stream, and both surfaces to resolve them.

## Acceptance criteria
- [ ] Index patterns, aliases, and data-streams representable as one entity in the semantic layer
- [ ] Whitelist matches a pattern/alias/data-stream entity (not just concrete index names)
- [ ] `atlas init` emits one entity per pattern/alias/data-stream
- [ ] Both SQL and DSL surfaces query a pattern/alias/data-stream entity correctly end-to-end

## Blocked by
- #3267
- #3268

PR #3318, opened 2026-06-07
fix(api): don't split dotted ES index-pattern names into bogus whitelist keys (#3317)

### LINKED ISSUE #3317: fix(api): whitelist splits dotted ES pattern table names, injecting a bogus unqualified key
## Context

Surfaced during the code review of #3316 (PR resolving #3269, ES/OpenSearch index patterns as entities). Low severity — filing so it isn't lost.

An Elasticsearch index-pattern entity can legitimately have a dotted base, e.g. indices `filebeat-7.10.0-2024.01.01` / `…02` collapse to an entity with `table: filebeat-7.10.0-*` (see `indexPatternBase` / `detectIndexPatterns` in `plugins/elasticsearch/src/mapping.ts`).

The whitelist loader treats every `table:` value as a potentially schema-qualified SQL identifier and splits it on `.` to derive an unqualified last-segment key:

```ts
const normalized = normalizeTableIdentifier(entity.table); // "filebeat-7.10.0-*"
const parts = normalized.split(".");                        // ["filebeat-7", "10", "0-*"]
tables.add(parts[parts.length - 1].toLowerCase());          // adds bogus "0-*"
tables.add(normalized.toLowerCase());                       // adds the real "filebeat-7.10.0-*"
```

So a `filebeat-7.10.0-*` entity registers **two** whitelist keys: the correct full name `filebeat-7.10.0-*` (queries validate fine) **and** a meaningless fragment `0-*`.

## Impact (low)

- The real query path is unaffected — the full-name key is also added, so `FROM "filebeat-7.10.0-*"` (SQL) and `queryElasticsearch(index: "filebeat-7.10.0-*")` (DSL) still validate.
- The bogus `0-*` key is a latent over-grant: it's a member of the DSL whitelist, so a query naming index `0-*` would pass `validateIndexAccess`'s wildcard-membership check. `0-*` is not a realistic index name and the agent has no reason to emit it, so impact is minimal — but it's noise that widens the allow-list.
- Affects all three load paths that do this split: `loadEntitiesFromDir`, `loadOrgWhitelist`, and `registerPluginEntities` in `packages/api/src/lib/semantic/whitelist.ts`.

## Root cause

`normalizeTableIdentifier` + the `.split(".")` last-segment heuristic assume a table name is a SQL `schema.table` identifier. That holds for Postgres/MySQL but not for ES pattern/data-stream/alias names, which can contain `.` as ordinary characters (version strings, date fragments).

## Key files

- `packages/api/src/lib/semantic/whitelist.ts` — `normalizeTableIdentifier`, `loadEntitiesFromDir` (~L177-181), `loadOrgWhitelist` (~L710-712), `registerPluginEntities` (~L532-534)
- `plugins/elasticsearch/src/mapping.ts` — pattern naming (`indexPatternBase`)

## Possible approaches

- Only add the unqualified last-segment key when the name actually looks schema-qualified for a SQL dialect (e.g. skip the split for names containing `*`/`?`, or when the connection's `dbType` is `elasticsearch`).
- Or have the ES profiler not emit dotted-base patterns (less desirable — `filebeat-7.10.0-*` is a correct ES pattern).
- Whichever path: add a round-trip test for a dotted-base pattern name so the whitelist key set is asserted.

## Acceptance criteria

- A `filebeat-7.10.0-*` entity registers only the full-name whitelist key (no bogus `0-*` fragment), on all three load paths.
- The full-name key still validates `FROM "filebeat-7.10.0-*"` and the DSL `index: "filebeat-7.10.0-*"`.
- Regression test covering a dotted-base ES pattern entity's whitelist keys.

## Dependencies

- Follow-up from #3316 / #3269.

msywulak:
Fixes #3317.

## Problem

The semantic whitelist loader treated every entity `table:` value as a SQL `schema.table` identifier and split it on `.` to derive an unqualified last-segment key. For a dotted Elasticsearch/OpenSearch name, that injects a meaningless fragment alongside the real full name — e.g. `filebeat-7.10.0-*` → bogus `0-*`, the concrete date index `metrics-2024.01.01` → `01`, the data stream `logs-nginx.access-default` → `access-default`.

Low severity (the full-name key is also added, so the real query path validates fine), but the bogus key is a latent over-grant: it's a member of the DSL whitelist, so `validateIndexAccess`'s wildcard-membership check would pass it. It widened the allow-list with noise.

## Fix

Two complementary tiers, so the over-grant is closed for **every** ES name — not just wildcard patterns.

**1. Engine-aware marker (authoritative).** Add an optional `identifier_style: "sql" | "opaque"` to the shared `EntityShape`. `"opaque"` means `table` is a literal datasource identifier where `.` is an ordinary character; the loader registers the full name only and never dot-splits it. ES entities **always** carry `identifier_style: "opaque"`, stamped in both `mapping.ts` builders (`mappingToEntity` + `buildLogicalEntity`), so it serializes to disk via the CLI profiler's `yaml.dump` and flows uniformly through all three load paths (disk, plugin-registration, org-DB) — each parses the YAML with `EntityShape` before deriving keys. `EntityShape` is `.passthrough()` and the CLI entity reader is tolerant, so no validator rejects the field.

**2. Name heuristic (fallback)** for any entity without the marker: extract the key derivation into a single `tableWhitelistKeys()` helper (also de-duplicates previously copy-pasted logic across `loadEntitiesFromDir`, `registerPluginEntities`, `loadOrgWhitelist`) and derive the unqualified key only when the name is a genuine dotted SQL identifier — ≥2 segments, each a bare SQL identifier (`[A-Za-z_][A-Za-z0-9_$]*`). A `-`/`*`/`?` or a digit-leading segment can never appear in a bare SQL identifier, so the segment shape safely discriminates ES names from SQL `schema.table` for every dialect's real table refs (BigQuery `dataset.table` with the project as connection config; Snowflake/ClickHouse/Postgres/MySQL unquoted).

The full (normalized, lowercased) name is always registered, so SQL `FROM "filebeat-7.10.0-*"` and DSL `index: "filebeat-7.10.0-*"` still validate.

Together: the marker closes 100% for profiled ES entities (**including** pure word-dotted names like an alias literally `logs.app`, which the heuristic alone can't distinguish from `schema.table`); the heuristic remains a safety net for any unmarked opaque name.

### Evolution of this PR

1. Initial commit: wildcard-only (`*`/`?`) guard for the exact `filebeat-7.10.0-*` case.
2. Review pass: found non-wildcard dotted ES names (date indices, data streams) with the same defect → generalized to the SQL-identifier-segment heuristic.
3. Final: added the engine-aware `identifier_style` marker to close the name-undecidable residual for good.

## Tests

- `tableWhitelistKeys` unit tests — SQL `schema.table` and `db.schema.table`; bare names; quoted-identifier stripping; wildcard patterns; dotted date indices; dotted data-stream/alias names; digit-leading segments; the heuristic-fallback residual; and the `opaque` marker overriding both a word-dotted and a SQL-looking name.
- Disk- and plugin-path regression tests for wildcard and non-wildcard dotted ES names, plus an end-to-end disk-path test that `identifier_style: opaque` suppresses the dot-split for a pure word-dotted name.
- `mapping.test.ts`: every emitted ES entity (standalone + logical) carries `identifier_style: "opaque"`.

## Acceptance criteria

- [x] A `filebeat-7.10.0-*` entity registers only the full-name whitelist key (no bogus `0-*`), on all three load paths
- [x] The full-name key still validates `FROM "filebeat-7.10.0-*"` and the DSL `index: "filebeat-7.10.0-*"`
- [x] Regression test covering a dotted-base ES pattern entity's whitelist keys
- [x] Extended beyond the issue to cover non-wildcard dotted ES names (data streams, date indices) and pure word-dotted names — the whole over-grant class is closed

## Verification

- `semantic.test.ts` → 35 pass; ES plugin suite → 356 pass; API affected isolated runner → 88 files pass
- Dependent suites green: `sql.test.ts` (103), `sql-connection-whitelist`, `sql-org-whitelist`, `semantic-org`, `semantic-multisource`, `group-scoped-loader`, `cli/elasticsearch-profiler`
- `bun run lint` clean; `bun run type` clean

Follow-up from #3316 / #3269.

https://claude.ai/code/session_011wKbAvWsW5bjZvuryMtjsr

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented incorrect dot-splitting of dotted Elasticsearch index/data-stream/alias names so full identifiers are recognized correctly; ES-derived entities are now marked to avoid SQL-style splitting.
* **Tests**
  * Added regression and unit tests covering dotted patterns, wildcard/dashed/digit segments, quote-stripping/lowercasing, and opaque identifier behavior.

### DIFF
--- a/packages/api/src/lib/__tests__/semantic.test.ts
+++ b/packages/api/src/lib/__tests__/semantic.test.ts
@@ -226,6 +226,22 @@ describe("per-connection whitelists", () => {
     expect(tables.has("access-default")).toBe(false);
     expect(tables.size).toBe(1);
   });
+
+  // #3317: an `identifier_style: opaque` marker (which ES entities always carry)
+  // closes the name-undecidable residual — a pure word-dotted ES name that the
+  // name heuristic alone would split is registered as the full name only.
+  it("identifier_style: opaque suppresses the dot-split end-to-end (disk path)", () => {
+    const dir = ensureEntitiesDir(`es-opaque-${testCounter}`);
+    writeFileSync(
+      resolve(dir, "stream.yml"),
+      `table: logs.app\nidentifier_style: opaque\ncolumns:\n  message:\n    type: text\n`,
+    );
+
+    const tables = getWhitelistedTables("default", dir);
+    expect(tables.has("logs.app")).toBe(true);
+    expect(tables.has("app")).toBe(false);
+    expect(tables.size).toBe(1);
+  });
 });
 
 describe("tableWhitelistKeys", () => {
@@ -276,11 +292,22 @@ describe("tableWhitelistKeys", () => {
     expect(tableWhitelistKeys("metrics.2024.01")).toEqual(["metrics.2024.01"]);
   });
 
-  // Documented residual: a pure word-dotted name (every segment a valid SQL
-  // identifier) is indistinguishable from `schema.table` by name alone, so it
-  // still contributes the unqualified key. Captures current behavior.
-  it("pure word-dotted name → still splits (name-undecidable residual)", () => {
+  // Heuristic fallback (no marker): a pure word-dotted name is name-undecidable
+  // — indistinguishable from `schema.table` — so it still splits. The opaque
+  // marker (below) is what closes this for ES entities.
+  it("pure word-dotted name without marker → still splits (heuristic fallback)", () => {
     expect(tableWhitelistKeys("logs.app").sort()).toEqual(["app", "logs.app"]);
+  });
+
+  // #3317: the `identifier_style: opaque` marker is authoritative — it forces
+  // full-name-only even for a name the heuristic would treat as SQL, closing
+  // the pure word-dotted residual.
+  it("opaque marker → full name only, even for a word-dotted name", () => {
+    expect(tableWhitelistKeys("logs.app", { opaque: true })).toEqual(["logs.app"]);
+  });
+
+  it("opaque marker → suppresses the unqualified key for a SQL-looking name", () => {
+    expect(tableWhitelistKeys("public.orders", { opaque: true })).toEqual(["public.orders"]);
   });
 });
 
--- a/packages/api/src/lib/__tests__/semantic.test.ts
+++ b/packages/api/src/lib/__tests__/semantic.test.ts
@@ -253,6 +253,12 @@ describe("tableWhitelistKeys", () => {
     expect(tableWhitelistKeys("orders")).toEqual(["orders"]);
   });
 
+  // #3317 review: a malformed empty `table:` must register no whitelist key.
+  it("empty table name → no keys (no bogus empty-string entry)", () => {
+    expect(tableWhitelistKeys("")).toEqual([]);
+    expect(tableWhitelistKeys("", { opaque: true })).toEqual([]);
+  });
+
   it("strips identifier quotes and lowercases", () => {
     expect(tableWhitelistKeys(`"User"`)).toEqual(["user"]);
     expect(tableWhitelistKeys('analytics."Events"').sort()).toEqual(["analytics.events", "events"]);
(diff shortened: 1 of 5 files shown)


msywulak:
@coderabbitai review

---

Issue #3552, opened 2026-06-14
feat(mcp): profile plugin-managed datasources over MCP — depends on the #3303 plugin-profiler seam


msywulak:
**Parent:** #3303 (the plugin-profiler PRD) · Split out of #3547 (the profiling half) · Grandparent #3483

> ⚠️ Not committed v0.0.15 work — this is a slice of the parked PRD #3303 and should be sequenced when #3303 is `/kickoff`-sliced. (The GitHub milestone field still shows v0.0.15 because it was mis-filed there; clear it manually.)

## Context

#3547 was re-scoped to its **provisioning** half (capability-derived check + plugin-aware validate-before-persist + per-type credential elicitation for ClickHouse/Snowflake/ES/OpenSearch/REST). This issue tracks the **profiling** half that #3547 deferred.

`profile_datasource` (#3512, shipped in #3545) is still postgres/mysql-only. The blocker is not the MCP seam — `SemanticGenerator` already exposes the `profileFn` injection point (`DatasourceProfiler`) designed exactly for plugin dbTypes, and `loadDatasourceProfileTarget` already returns a clean `unsupported` envelope for non-pg/mysql. The blocker is that **there is no API-reachable profiler implementation** for these types:

- ClickHouse / Snowflake profilers exist only in `packages/cli/lib/profilers/`, which `@atlas/api` must not import (ADR-0013 core→plugin decoupling).
- ES/OpenSearch index profiling and REST/OpenAPI endpoint discovery have their own, different "profile" semantics.

That API-reachable, registry-resolved profiler is precisely the **reusable plugin-profiler seam** of parked PRD **#3303** (`listObjects` / `profile` contract per datasource plugin). This issue should start once that seam lands.

## Key files
- `packages/api/src/lib/effect/semantic-generator.ts` — `DatasourceProfiler` / `profileFn` injection seam (already present)
- `packages/api/src/lib/datasources/mcp-lifecycle.ts` — `loadDatasourceProfileTarget` (lift the pg/mysql-only discriminant), `runSemanticProfile` (inject the resolved `profileFn`)
- `packages/mcp/src/datasource-tools.ts` — `profile_datasource` (drop the `unsupported` short-circuit for newly-supported types)
- the #3303 plugin-profiler seam (registry-resolved `profileFn` source)

## Acceptance criteria
- [ ] `profile_datasource` profiles each provisioning-supported plugin type via a registry-resolved injected `profileFn`, populating entities + the table whitelist
[...]

Issue #3547, opened 2026-06-14
MCP datasource PROVISIONING: extend create_datasource beyond Postgres/MySQL (ClickHouse / Snowflake / Elasticsearch / OpenSearch / REST)


msywulak:
**Parent:** #3483 · Follow-up to #3511 / #3512 (landed in #3545) · Profiling half split to #3552

## Scope (re-scoped 2026-06-14)

This issue now covers the **provisioning** half only. The **profiling** half — injecting a `profileFn` so `profile_datasource` populates entities + the whitelist for these types — is split to **#3552**, because it is a hard blocker on the #3303 reusable plugin-profiler seam (the only ClickHouse/Snowflake profilers today live in `packages/cli/lib/profilers/`, which `@atlas/api` must not import; ES/OpenSearch/REST have their own profile semantics).

## Context

`create_datasource` (#3511), as shipped in #3545, is scoped to **postgres / mysql** — the native dbTypes where the validate-before-persist pre-flight (`connections.register` ephemeral pool → `SELECT 1` → unregister) genuinely tests connectivity. Other types are rejected with an actionable `unsupported` envelope (`MCP_NATIVE_DB_TYPES` in `lib/datasources/provisionable-types.ts`).

Plugin-managed pools (clickhouse / snowflake / bigquery / …) are built via `registerPluginDatasourceInstall` → the plugin's `createFromConfig`, not the core `connections.register` switch (see `datasource-registry-bridge.ts`). So the **test-connect** path needs a plugin-aware seam before these types can flow through `create_datasource`.

## Key files
- `packages/api/src/lib/datasources/provisionable-types.ts` — `MCP_NATIVE_DB_TYPES` / `MCP_PROVISIONABLE_CATALOG_SLUGS` (capability-derived check)
- `packages/api/src/lib/datasources/mcp-lifecycle.ts` — `provisionDatasource` (plugin-aware ephemeral probe)
- `packages/api/src/lib/db/datasource-registry-bridge.ts` — `registerPluginDatasourceInstall` (the plugin createFromConfig path to generalize the probe out of)
- `packages/mcp/src/datasource-tools.ts` — `create_datasource` `db_type` enum + per-type masked credential elicitation

## Acceptance criteria
[...]

Issue #3303, opened 2026-06-07
PRD: in-product onboarding for all datasource types — reusable plugin profiler seam


msywulak:
> **PRD — re-scoped 2026-06-14 (post-MCP-V2).** The core thesis still holds and is still unbuilt as a *registry-resolved* seam, but the v0.0.15 MCP V2 wave shipped a substantial chunk of the foundation this PRD assumed it would have to build. See the **Revision** section below for what's now done, what changed shape, and what genuinely remains. No longer parked; needs a fresh milestone anchor.

## Revision (2026-06-14) — post-MCP-V2 re-scope

A PRD-vs-code audit after v0.0.15 (MCP V2) found three shifts under this PRD. Summary:

**Already shipped (this PRD assumed it had to be built):**
- **Blocker #1 (#3506) is DONE.** `SemanticGenerator` (`packages/api/src/lib/effect/semantic-generator.ts`) is the shared profile→generate→persist engine, and it already exposes the exact **injection point** the seam needs: the `DatasourceProfiler` type (`:75-82`), a `profileFn?` option (`:108`), and `resolveProfiler` (`:260-280`) — pg/mysql in-core, injected otherwise, with an explicit `unsupported_db_type` failure (never a silent skip, matching this PRD's AC). Introspection types (`ProfilingResult`/`DatabaseObject`/…), group-scoped persist, and the credential-scrub security AC are all satisfied here.
- **The MCP provisioning half shipped** via #3547 → #3553 (`resolveProvisionCapability`, `provisionDatasource`, `provisionRestDatasource` in `mcp-lifecycle.ts`; `MCP_PROVISIONABLE_CATALOG_SLUGS` in `provisionable-types.ts` covers clickhouse/snowflake/bigquery/elasticsearch). The old "#3300 in-flight form-install handler" dependency is therefore resolved.

[...]

PR #3553, opened 2026-06-14
feat(mcp): capability-derived datasource provisioning + plugin-aware test-connect (#3547)

msywulak:
## What & why

Implements the **provisioning** half of **#3547** — extends MCP datasource provisioning beyond native postgres/mysql to **every provisionable datasource type**. The profiling half is split to **#3552** (child of PRD #3303, blocked on the reusable plugin-profiler seam).

Three commits, by mechanism:

### 1. Capability seam + plugin-aware test-connect (SQL types via `createFromConfig`)
- **AC #1** — `resolveProvisionCapability` replaces the hardcoded `MCP_NATIVE_DB_TYPES` gate: provisionable iff native pg/mysql **OR** a datasource plugin with a registered `createFromConfig`. Provisioning and future profiling read the same predicate, so support stays in lockstep.
- **AC #2** — extracted `findDatasourcePluginConnection` + added `probePluginDatasourceConnection` (`createFromConfig → SELECT 1 → close`, always closing the probe) to `datasource-registry-bridge` — one seam shared by install and pre-flight. `provisionDatasource` validate-before-persists plugin connections, scrubbing credentials and persisting nothing on failure. Lands **ClickHouse + Snowflake**.

### 2. Schema-driven multi-field masked elicitation (non-`url` credential shapes)
- **AC #4** — `create_datasource` now collects credentials via a **config_schema-driven masked form** (`elicitMaskedForm`, the multi-field sibling of `elicitMaskedField`): one principal-bound, TTL'd, single-use requestState authorizes the whole form; values travel out-of-band, never into agent/LLM context. `ProvisionDatasourceInput` generalized from a single `url` to `{ config, secretKeys }`; the error-scrub redacts **every** secret field value. Adds **Elasticsearch / OpenSearch** (`apiKey` / basic / SigV4, one catalog slug) + **BigQuery** (`service_account_json` + `project_id`).

### 3. REST / OpenAPI provisioning
- Adds `create_rest_datasource` — REST uses a different install path (the `openapi-generic` form handler, which probes the spec on install). Routes through the **same lib seam the admin `/install-form` route calls** (`getInstallHandler().validateConfig`, ADR-0016), maps a failed probe / bad auth to a typed validation outcome with the credential scrubbed, read-only by default.

## Result
`create_datasource` (postgres · mysql · clickhouse · snowflake · bigquery · elasticsearch/opensearch) + `create_rest_datasource` (OpenAPI) — the full #3547 provisioning set, all with masked, schema-driven, out-of-band credential collection and validate-before-persist.

## Out of scope (tracked)
- **AC #3 — profiling** these types → **#3552** (depends on #3303's plugin-profiler seam; CLI profilers must not be imported into `@atlas/api`).

## Testing
- New lib tests: capability resolution, plugin probe (success / failure-rolls-back-nothing / scrub), `loadProvisionConfigFields` mapping, REST install / validation-scrub / rethrow.
- New MCP tool tests: schema-driven single + multi-field (ES) elicitation, capability-unsupported-before-elicitation, REST form install / validation / decline; `elicitMaskedForm` round-trip (multi-field, empty-drop, decline).
- Existing provision tests migrated to the `{ config, secretKeys }` shape; server/smoke/sse tool-list + count assertions updated (13 → 14 tools).
- Green locally: lint, `tsgo` (api + mcp, 0 errors), full MCP suite (27/27), api affected, syncpack, template-drift, railway-watch.

Closes the provisioning scope of #3547; profiling tracked in #3552.

https://claude.ai/code/session_01PXSoKuosXCuocF4LcbEeSM

### DIFF
--- a/packages/api/src/lib/db/datasource-registry-bridge.ts
+++ b/packages/api/src/lib/db/datasource-registry-bridge.ts
@@ -102,19 +102,34 @@ export type PluginProbeOutcome =
 const PLUGIN_PROBE_SQL = "SELECT 1";
 const PLUGIN_PROBE_TIMEOUT_MS = 10_000;
 
+/** A built plugin connection, plus the OPTIONAL liveness surface some adapters expose. */
+type ProbeableConnection = {
+  query(sql: string, timeoutMs?: number): Promise<unknown>;
+  close(): Promise<void>;
+  /**
+   * A non-SQL liveness round-trip (e.g. Elasticsearch/OpenSearch `ping()` → cluster
+   * info). Preferred over `SELECT 1` when present: the ES connection's `query()`
+   * routes to the cluster SQL API (`/_sql`, optional on OpenSearch), so probing it
+   * with `SELECT 1` would test the wrong surface — `ping()` is what the plugin's own
+   * health check uses.
+   */
+  ping?(timeoutMs?: number): Promise<unknown>;
+};
+
 /**
  * Plugin-aware ephemeral test-connect for a CANDIDATE plugin-managed datasource
  * (#3547). Builds a throwaway connection via the registered plugin's
- * `createFromConfig(decryptedConfig)`, runs a `SELECT 1` probe under a short
- * timeout, and ALWAYS closes it — registering nothing in the
- * `ConnectionRegistry` and persisting nothing. This is the plugin counterpart
- * to the native `connections.register(probeId) → healthCheck` pre-flight, so
- * the MCP `create_datasource` validate-before-persist path is uniform across
- * native and plugin dbTypes.
+ * `createFromConfig(decryptedConfig)`, runs a liveness probe under a short
+ * timeout — the connection's own `ping()` when it exposes one (ES/OpenSearch),
+ * else a `SELECT 1` (ClickHouse/Snowflake/BigQuery) — and ALWAYS closes it,
+ * registering nothing in the `ConnectionRegistry` and persisting nothing. The
+ * plugin counterpart to the native `connections.register(probeId) → healthCheck`
+ * pre-flight, so the MCP `create_datasource` validate-before-persist path is
+ * uniform across native and plugin dbTypes.
  *
  * `message` carries the RAW driver error (which may echo the credential) — the
  * caller (`mcp-lifecycle.provisionDatasource`) runs it through
- * `scrubSecretFromMessage` before it ever reaches a client/agent/log. Returning
+ * `scrubSecretsFromMessage` before it ever reaches a client/agent/log. Returning
  * the raw string keeps the secret-scrub seam in one place.
  */
 export async function probePluginDatasourceConnection(
@@ -132,10 +147,16 @@ export async function probePluginDatasourceConnection(
     };
   }
 
-  let built: { query(sql: string, timeoutMs?: number): Promise<unknown>; close(): Promise<void> } | undefined;
+  let built: ProbeableConnection | undefined;
   try {
-    built = await conn.createFromConfig(decryptedConfig);
-    await built.query(PLUGIN_PROBE_SQL, PLUGIN_PROBE_TIMEOUT_MS);
+    built = (await conn.createFromConfig(decryptedConfig)) as ProbeableConnection;
+    // Prefer the connection's native liveness check (ES/OpenSearch `ping`); fall
+    // back to `SELECT 1` for SQL-only adapters that expose no `ping`.
+    if (typeof built.ping === "function") {
+      await built.ping(PLUGIN_PROBE_TIMEOUT_MS);
+    } else {
+      await built.query(PLUGIN_PROBE_SQL, PLUGIN_PROBE_TIMEOUT_MS);
+    }
     return { ok: true };
   } catch (err) {
     return {
--- a/packages/api/src/lib/db/datasource-registry-bridge.ts
+++ b/packages/api/src/lib/db/datasource-registry-bridge.ts
@@ -70,6 +70,93 @@ function getDatasourceConnection(plugin: unknown): DatasourceConnectionShape | u
 }
 
 /**
+ * Find the registered datasource plugin `connection` for a `dbType`, or
+ * `undefined` when none is registered. The single home for the
+ * `PluginRegistry.getAll()` → structural-shape → `dbType` match that both the
+ * boot/install registration ({@link registerPluginDatasourceInstall}) and the
+ * pre-flight probe ({@link probePluginDatasourceConnection}) consult — so
+ * "which plugin builds this type" lives in exactly one place (#3547).
+ *
+ * Uses `getAll()` (not `getByType`, which filters to `status==="healthy"`): the
+ * plugin is consulted purely as an ADAPTER via `createFromConfig`, independent
+ * of whether a static config-defined connection of its dbType is healthy. Lazy-
+ * imports the registry to keep it out of this module's static graph (several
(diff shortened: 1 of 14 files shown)


packages/mcp/src/elicitation.ts
```diff
@@ -284,34 +284,123 @@ export async function elicitMaskedField(
...
 
+  const properties: Record<string, { type: "string"; title: string; description?: string }> = {};
+  const required: string[] = [];
+  for (const field of args.fields) {
+    properties[field.name] = {
+      type: "string",
+      title: field.title,
+      ...(field.description ? { description: field.description } : {}),
+    };
+    if (field.required) required.push(field.name);
```
msywulak, line 415: **Open finding (not fixed in 71575e7): `select`/`options`/`default` collapse to free-text in the masked form.**

Every catalog field is rendered as `{ type: "string" }` here, so `select`-typed fields lose their enum and default. Two concrete cases:

- The openapi-generic `auth_kind` is a **required** `select` with a `bearer` default. In the masked form it becomes required free text with no default, so a user who leaves it blank now gets "Missing required field: Authentication" instead of the old zero-effort default — and a near-miss like `Bearer`/`api-key` fails handler validation downstream.
- ES `engine` is also a `select` (`elasticsearch`/`opensearch`).

I left this one for you rather than fixing it blind: rendering enum/default means threading `field.type`/`options`/`default` through `ElicitMaskedFormArgs` into `requestedSchema` (MCP elicitation supports enum schemas), which touches the security-critical mint/verify path. Worth deciding whether to (a) thread enum+default through, or (b) keep credentials masked-string-only and collect non-secret `select`s (`auth_kind`, `engine`) as agent tool args like `schema`/`display_name`. Option (b) is consistent with the `schema` fix in 71575e7.

---

packages/api/src/lib/db/datasource-registry-bridge.ts
```diff
@@ -69,6 +69,164 @@ function getDatasourceConnection(plugin: unknown): DatasourceConnectionShape | u
...
+  // connection the build later yields (below) so a slow-but-eventual resolve
+  // can't leak a pool.
+  const probeRun = (async () => {
+    built = (await createFromConfig(decryptedConfig)) as ProbeableConnection;
+    // Prefer the connection's native liveness check (ES/OpenSearch `ping`); fall
+    // back to `SELECT 1` for SQL-only adapters that expose no `ping`.
+    if (typeof built.ping === "function") {
+      await built.ping(timeoutMs);
+    } else {
+      await built.query(PLUGIN_PROBE_SQL, timeoutMs);
```
msywulak, line 170: **Open finding (not fixed in 71575e7): BigQuery pre-flight assumes a literal `SELECT 1` contract.**

For an adapter with no `ping()` (BigQuery), the probe falls back to `query("SELECT 1")`. BigQuery's SQL dialect may not answer an unqualified `SELECT 1` the way pg/clickhouse do (standard-SQL parsing / no `FROM` quirks), so a **valid** credential could fail pre-flight and surface a spurious `health_error`, blocking provisioning of a working datasource.

I didn't fix this because the right fix is plugin-side and I couldn't verify BigQuery's adapter behavior from here: either confirm BigQuery's `createFromConfig` connection answers a bare `SELECT 1`, or give it a `ping()` (the branch above already prefers `ping`). Note there's **no end-to-end BigQuery provision test** (ties to the AC4 coverage gap below), so this path is currently unverified for the one type whose probe shape is most likely to differ.

---

msywulak (commented):
## Review — capability-derived datasource provisioning

Ran four passes: **Standards**, **Spec** (vs #3547), **Security**, and a high-effort **correctness/cleanup** review. Overall this is disciplined and well-tested — the credential-handling core holds up. No security vulnerabilities found (verified out-of-band masking, scrub-on-every-error-path, requestState anti-replay, authz-before-probe, validate-before-persist).

I pushed fixes for three correctness findings in **71575e7**, and left two as inline comments above for your call.

### ✅ Fixed in 71575e7
1. **`schema`/search_path was leaking into the masked credential prompt.** The postgres/mysql/clickhouse catalog `config_schema` define a non-secret `schema` field that wasn't in `NON_CREDENTIAL_CONFIG_KEYS`, so it was elicited in the "never shared with the agent" form, and the agent could no longer set a search_path (the `schema` tool arg was dropped). Added it to the skip-set and re-exposed it as an optional `create_datasource` arg, merged into config like `description`. (This contradicted the stated intent of commit 89f306b.)
2. **Plugin pre-flight could hang the MCP call.** `createFromConfig` had no timeout, and the `query`/`ping` timeoutMs is only honored if the adapter cooperates — so an adapter that eagerly connects to an unreachable host could stall `create_datasource`. Now the whole build+probe runs under one deadline, with a salvage-close so a late-resolving build can't leak a pool.
3. **Whitespace-only required field was probed as a blank credential.** `elicitMaskedForm` kept whitespace (`raw.length > 0`); now blank-after-trim is dropped at the source so the required-field guard catches it as missing. Internal whitespace is preserved.

### 🔎 Open (inline above)
4. **`select`/`options`/`default` collapse to free-text** in the masked form (`auth_kind` required-select loses its `bearer` default; ES `engine`).
5. **BigQuery probe assumes a literal `SELECT 1`** — a valid credential may fail pre-flight; no end-to-end BigQuery test exists.

### Spec (#3547)
AC1/AC2/AC3 met and in-scope; `create_rest_datasource` correctly in scope; profiling correctly deferred to #3552 (no CLI-profiler import into `@atlas/api`). **AC4 coverage is partial** — only ClickHouse gets the full end-to-end provision-success + rollback + scrub trio; Snowflake and BigQuery (the multi-field `service_account_json` shape) have no end-to-end provision test. Worth adding before merge.

### Standards
No hard violations.

All checks green locally after the fix commit: lint, `tsgo`, full MCP suite (27/27 → tools now 14), api affected, and the new regression tests (schema-arg merge, schema exclusion, probe deadline + salvage-close, whitespace drop).

---

packages/mcp/src/elicitation.ts
msywulak: Resolved in 63c5bbb — went with **option (b)+**: the masked form now mirrors the catalog `config_schema` field types. `type`/`options`/`default` are threaded through `ProvisionConfigField` → the form's `requestedSchema`, so a `select` (`auth_kind`, ES `engine`) renders as an enum with its default; the catalog default is also injected server-side when the client returns a defaulted field empty/omitted. This does **not** touch the mint/verify/anti-replay seam — only the `requestedSchema` properties.

Standardized rule: the masked form carries every connection/auth field per its catalog type (string→masked text, select→enum+default); only non-connection metadata (labels, agent-owned `schema`, admin governance) stays out as tool args. Tests cover enum+default rendering and default-injection.

---

packages/api/src/lib/db/datasource-registry-bridge.ts
msywulak: Verified in 63c5bbb — **not a bug**, no change needed. BigQuery's adapter (`plugins/bigquery/src/connection.ts`) runs `query()` with `useLegacySql: false` (Standard SQL), where a bare `SELECT 1` is valid and returns one row — so the probe's SQL fallback is correct for BigQuery. Also closed the coverage gap: added an end-to-end BigQuery provision test (multi-field `service_account_json` + `project_id`: success + probe-failure rollback + credential scrub), alongside Snowflake, so the AC4 per-type matrix is now literal.

---

Issue #3490, opened 2026-06-13
flaky-ci: Bun 1.3.13 segfault crashes api-tests shards (false-negative failures)


msywulak:
## Symptom

A CI `api-tests (N/4)` shard reports `failure` not because a test failed, but because **Bun itself segfaults** mid-run, crashing the test subprocess. The runner then reaps the hung process and marks the shard failed — a false negative.

## Evidence

Observed on PR #3487, run [27468021015](https://github.com/AtlasDevHQ/atlas/actions/runs/27468021015) (`api-tests (2/4)`, head `e1aa6933`). The log for `src/api/__tests__/admin-roles.test.ts`:

```
Elapsed: 788ms | User: 503ms | Sys: 98ms
panic(main thread): Segmentation fault at address 0x70000002F
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
https://bun.report/1.3.13/...
```

- The test body completed in **788ms** with **zero `(fail)` markers and zero failed assertions**. The ~29s "elapsed" the shard reported is the hung-subprocess wall time after the segfault, before reap.
- `admin-roles.test.ts` passed **10/10** raw + **3/3** isolated-runner runs locally (Bun **1.3.11**), and the full isolated api suite is **612/612 green**.
- CI runs **Bun 1.3.13**; local was 1.3.11. Segfaults of this kind are non-deterministic across Bun patch versions.

## Impact

Spurious red shards force needless re-runs / branch updates and erode trust in the gate. Distinct from the timing-sensitive parallel-runner flake in #3455 — this is a native Bun crash, not a `waitFor` timeout.

## Options

- Pin/upgrade the Bun version in the CI image to a build without this segfault (durable fix).
- Add a runner-level retry for a shard that exits via signal/segfault (vs. a real assertion failure).
- File the `bun.report` link upstream if it recurs.

## Acceptance criteria
- [ ] CI api-tests shards no longer false-fail on a Bun segfault (version pin/upgrade, or signal-aware retry)

## Notes
Surfaced while landing #3428 (PR #3487). No product-code defect.

PR #3717, opened 2026-06-16
fix(billing): route dunning emails through the durable email_outbox (#3680)

### LINKED ISSUE #3680: prod: dunning (payment-failure) emails bypass the durable email_outbox and are dropped on transport failure
## Summary

Found during the v0.0.13–v0.0.16 production-readiness audit (HIGH).

The v0.0.14 payment-failure recovery sequence sends dunning emails via raw `sendEmail`, **not** the durable `sendTransactionalEmail` (`email_outbox`) path that auth flows use. After the 3 in-process retries are exhausted the send is **permanently dropped** — there is no outbox row, and the Stripe webhook handler returns 2xx regardless, so there is no redelivery either.

Blast radius: a customer whose card failed never learns it, then gets auto-suspended at the end of the grace period with no prior warning — a hidden P1 billing problem.

## Key files
- `packages/api/src/lib/email/dunning.ts:109` (+ `:121-126`)
- Call chain: `packages/api/src/lib/auth/server.ts:1565-1576`, `:2165-2172`

## Proposed fix
Route dunning through `sendTransactionalEmail` (the durable `email_outbox` already exists and is used by auth flows). Move the `recordDunningEmail` dedup bookkeeping to enqueue/dispatch time so a deferred send is still recorded exactly once. Contained refactor within `dunning.ts` + its callers.

## Acceptance criteria
- [ ] Dunning emails enqueue to `email_outbox` and survive a transient transport failure (retried by the dispatcher, not dropped)
- [ ] Dedup bookkeeping (`recordDunningEmail`) is correct under deferred send — no double-send, no missed record
- [ ] Tests cover: transport-failure-then-redeliver, and the dedup guard across a deferred send

## Dependencies
None.

msywulak:
## Summary

Fixes #3680. Dunning (payment-failure) emails were sent via raw `sendEmail`, **not** the durable `sendTransactionalEmail` (`email_outbox`) path that auth flows use. After the 3 in-process `fetchWithRetry` attempts were exhausted the send was **permanently dropped** — no outbox row, and the Stripe webhook returns 2xx regardless (`sendDunning` swallows), so there was no redelivery either. A customer whose card failed never learned it, then got auto-suspended at the end of the grace period with no warning.

## What changed

`packages/api/src/lib/email/dunning.ts` — `dispatchDunningEmail`:

- **Route through the durable path.** Send via `sendTransactionalEmail({ emailType: step, orgId })` instead of raw `sendEmail`. On a *real* transport failure it enqueues the rendered notice to `email_outbox`, so the Scheduler-backed flusher re-sends it on a later tick — a transient provider outage no longer drops the email.
- **Move dedup bookkeeping to dispatch time.** `recordDunningEmail` now fires as soon as the send was **delivered OR durably enqueued** (`result.success || shouldEnqueueFailedSend(result)`). This records a deferred send exactly once, so Stripe's webhook redelivery neither double-sends nor re-enqueues. A no-transport (`log` provider) failure is neither delivered nor queued, so it stays unrecorded and retryable — preserving the existing "don't record when there's nowhere to deliver" semantics.

Contained to `dunning.ts`; the callers in `auth/server.ts` (`sendDunning`, `handlePaymentFailure`) are thin wrappers and need no change.

## Tests

`packages/api/src/lib/email/__tests__/dunning.test.ts`:

- **transport-failure-then-redeliver** — a real transport (`resend`) failure now enqueues to the outbox and records the deferred send (not dropped).
- **dedup across a deferred send** — Stripe redelivery of the same event neither re-sends nor re-enqueues (the dispatch-time record makes the `(user_id, step)` guard fire).
- Existing tests updated for the durable wrapper; the no-transport-failure case is clarified (not recorded, not enqueued).

The mock is a faithful stand-in for `sendTransactionalEmail` (mirrors `shouldEnqueueFailedSend` to track outbox enqueues); the real outbox enqueue machinery stays covered by `delivery.test.ts`.

## Acceptance criteria (#3680)

- [x] Dunning emails enqueue to `email_outbox` and survive a transient transport failure (retried by the dispatcher, not dropped)
- [x] Dedup bookkeeping correct under deferred send — no double-send, no missed record
- [x] Tests cover transport-failure-then-redeliver and dedup across a deferred send

## CI

Local gates green: lint, type, full test (dunning suite passes; 5 unrelated files flagged by the isolated runner all pass individually — known WSL-runner flakiness #3490), syncpack, template/security-header/railway/schema/openapi/oauth drift, test-discipline, settings-readers, published-symbols.

### DIFF
--- a/packages/api/src/lib/email/__tests__/delivery.test.ts
+++ b/packages/api/src/lib/email/__tests__/delivery.test.ts
@@ -427,6 +427,7 @@ describe("sendTransactionalEmail", () => {
         send: async () => ({ success: false, provider: "resend", error: "503" }),
         enqueueFailed: async (_m, o) => {
           captured.push({ emailType: o.emailType, ttlMs: o.ttlMs });
+          return true;
         },
       },
     );
@@ -441,11 +442,13 @@ describe("sendTransactionalEmail", () => {
         send: async () => {
           throw new Error("sendEmail blew up unexpectedly");
         },
-        enqueueFailed: async () => {},
+        enqueueFailed: async () => false,
       },
     );
     expect(result.success).toBe(false);
     expect(result.provider).toBe("log");
+    // A thrown send becomes a log-provider failure — nowhere to queue, so lost.
+    expect(result.durable).toBe(false);
   });
 
   it("returns the send result and does NOT enqueue on success", async () => {
@@ -457,10 +460,12 @@ describe("sendTransactionalEmail", () => {
         send: async () => ({ success: true, provider: "resend", messageId: "m1" }),
         enqueueFailed: async () => {
           enqueued++;
+          return true;
         },
       },
     );
     expect(result.success).toBe(true);
+    expect(result.durable).toBe(true);
     expect(enqueued).toBe(0);
   });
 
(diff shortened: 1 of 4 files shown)


msywulak:
Follow-up from review (commit `00e6edfa`): the original change recorded the once-per-customer dedup row whenever `shouldEnqueueFailedSend(result)` was true — an *intent* predicate. Because `sendTransactionalEmail` swallows a failed/throwing outbox enqueue and returns the original `DeliveryResult`, a real-transport failure whose `email_outbox` INSERT then threw (pg blip / pool exhaustion / schema drift) would be marked sent **and** permanently suppress future Stripe redeliveries — the send lost, masked by a "durably enqueued" log. That's a narrower, silent recurrence of the exact #3680 failure mode.

Fix: `sendTransactionalEmail` now returns a real `durable` outcome (delivered now OR row confirmed committed); `enqueueFailedTransactionalEmail` returns whether the row landed. Dunning records + counts **only** when `result.durable`, so a lost send stays unrecorded and retryable. Added tests for the enqueue-failure path and per-recipient isolation; `delivery.test.ts` now asserts `durable`.

Local gates: type ✅, lint ✅, affected isolated test run ✅ (13/13).

Issue #3943, opened 2026-06-24
test(prod): verify 3-region (us/eu/apac) signup + residency routing end-to-end


msywulak:
Pre-launch (v0.1.0 / #2919) verification: drive a **real, end-to-end signup through each of the three data-residency regions in PRODUCTION** — us / eu / apac — to confirm the region step and region routing actually work for a brand-new user. Split out from the cold-start activation audit (#3925) and the F3 decision (#3934).

## Why prod (can't be done locally or on staging)

The region step only renders on a **multi-region SaaS** deployment (EE `ResidencyResolver` with `residency.regions` configured). Local dev and staging are **single-region** — staging's residency arm is keyed `"local"` and never claims a real region. **Only prod has us/eu/apac**, so the three-region routing path can only be exercised against prod.

This is a one-time, human-in-the-loop verification (a "is anything messed up before launch" pass), **not** a CI test — it can't run unattended because prod signup requires a real email OTP (see Prerequisites).

## Current state (verified 2026-06-24, read-only)

- All three region API services are live: `api.useatlas.dev` (200), `api-eu.useatlas.dev` (200), `api-apac.useatlas.dev` (200).
- `deploy/api/atlas.config.ts` residency block declares `us` (default) / `eu` / `apac` correctly, `strictRouting: false`.
- The infra is in place — what's unverified is the **end-to-end user path**: does picking `eu`/`apac` at signup actually land the new workspace's metadata in that region's DB, and does the post-signup app work there.

## Method

Drive Playwright (or manual) through the full funnel once per region:
`/signup` (account) → email **OTP** → `/signup/workspace` → `/signup/region` (**pick the target region**) → `/signup/connect` (explore demo data) → `/signup/success`.

After each, confirm the workspace's `organization.region` is the picked region and that `resolveRegionDatabaseUrl` routes it to that region's DB.

Realistic execution: an agent/operator drives the browser; a human reads the OTP from the inbox and supplies it at the OTP step (3 hand-offs). Capture a screenshot at each step per region.

## Prerequisites / gotchas

- **Email OTP, no bypass in prod.** The production env-profile sets `requireEmailVerification: true` — signup sends an 8-char OTP via Resend (10-min expiry) with no magic-code/test-domain bypass. A human must read it from a real inbox.
[...]

Issue #3925, opened 2026-06-24
feat(web,api): cold-start activation audit + polish (demo → trial → first answer)


msywulak:
## Parent

#2919 — v0.1.0 public launch (milestone #74)

## What to build

A polish pass over the **SaaS cold-start activation funnel** — the path a brand-new visitor takes from landing to their first useful answer — removing friction at every step. The conversion metric is **time-to-first-answer from a cold click**: an HN visitor who hits a dead end, a confusing empty state, or an unclear gate bounces. Walk the whole funnel with Playwright as a new user, then fix the rough edges and instrument first-answer-time.

**Funnel to walk end-to-end:**
1. **Landing** — is the primary CTA obvious, and does it lead somewhere sensible?
2. **Demo** (the demo surface / demo.useatlas.dev) — can a visitor ask a question and get a real answer with zero signup? How long to first answer?
3. **Signup / start trial** — business-email gate, OTP/claim interstitial, any confusing or slow step.
4. **Authed first-run** — connect a first datasource → generate the semantic layer → first query → first answer. Empty states, the connection wizard, dead ends, missing guidance.
5. **The "aha"** — the first real answer from data.

**Friction classes to hunt:** confusing/empty empty-states, dead-end links, missing loading/skeleton states, unclear CTAs, error messages with no next step, connection-wizard friction, and any step that gives no feedback while it works.

**Scope of fixes:** make the **safe, conventional polish fixes** directly (copy, empty states, loading/feedback states, dead-end links, obvious CTA/guidance gaps). Where a change needs a genuine **product/UX decision**, implement the most standard / least-surprising option and **call it out explicitly in the PR description** for human review — do not block or halt on it.

## Setup & verification (the `/dev` hook)

- Bring up local dev via the **`/dev` skill in SaaS mode** (`ATLAS_DEPLOY_MODE=saas` + `ATLAS_DEPLOY_ENV=development`, which relaxes the SaaS fail-closed boot guards for local) so the **trial/signup funnel actually exists** locally — plain self-hosted dev has no trial path. Drive Playwright against `localhost:3000`.
  - Staging (`app.staging.useatlas.dev`) is an acceptable alternative for cross-checking the *real* deployed funnel, via the throwaway business-email + OTP account flow — but the fix→verify loop is tighter on local SaaS dev.
- Use the **demo dataset / dev seed** as the test datasource for the connect→query→answer leg so the funnel reaches a real first answer (no external DB needed).
[...]

PR #3945, opened 2026-06-24
fix(web,api): F3 region-step resolution — retry copy + regions-error alert

### LINKED ISSUE #3934: feat(web): region step — residency policy on persistent regions-API error (F3, P1)
**P1 — residency-policy decision** (the safe interim fix already shipped in #3928). Split out from the #3925 cold-start audit (`docs/development/cold-start-activation-audit.md` §F3).

The signup region step (`/signup/region`) intentionally does **not** auto-skip when the regions API errors — only on an explicit "not configured" 200 `[]` — so a transient failure can't push a user past a required residency pick. #3928 added an in-place **Retry** button (safe).

**Open decision (needs a human call):** should a *persistent* regions-API error eventually allow the user to "continue without a region", or must they always be blocked until a region resolves? This is a data-residency policy call, not a polish change — hence `ready-for-human`.

## Key files
- `packages/web/src/app/signup/region/page.tsx` — region step (Retry already added in #3928)
- `/api/v1/onboarding/regions` — the regions source
- residency / region config — where a "required region" is determined (and whether it differs self-hosted vs SaaS)

## Acceptance criteria
- [ ] Decision recorded: persistent error → block vs. allow continue-without-region (and under what conditions / deploy modes)
- [ ] Behavior implemented to match the decision, consistent with residency requirements
- [ ] No path lets a user skip a *required* residency pick due to a transient error

## Dependencies
- From #3925 (audit doc §F3). Builds on #3928 (Retry button). Milestone #74 — v0.1.0 Public Launch.

msywulak:
## Summary

Resolves **F3** (#3934) from the cold-start activation audit (#3925) as part of v0.1.0 launch readiness.

**Decision: block + Retry is correct — no "continue without a region" escape hatch.** Grounded in how residency actually works: by the region step the workspace already physically lives in its home region (`organization.region` NULL → `resolveRegionDatabaseUrl` falls through to that same region), the only realistic error is a transient client-side fetch (which Retry handles), and a *persistent* failure is a bug to fix/alert on — not to route users around (if the API is down, the next steps are too).

- **web** — the region step swaps to honest copy + a **Contact support** path after ≥2 consecutive failed loads, instead of a generic disabled-Continue dead-end.
- **api** — `GET /onboarding/regions` emits a named `onboarding.regions_error` log event on any non-`not_configured` failure, so a real prod occurrence is caught by monitoring rather than discovered through a quietly-stuck signup.
- **comments** — corrected stale "region is immutable once set" comments (`services.ts`, `ee/platform/residency.ts`, `db/internal.ts`, `admin-residency.ts`) that predated the admin **cross-region migration** flow; they now say "one-way via the assign path; changeable via migration." Left the 409-response descriptions, which are contract-accurate.
- **docs** — decision + full rationale recorded in `docs/development/cold-start-activation-audit.md` §F3.

The end-to-end prod verification of all three regions is tracked separately in **#3943** (needs a prod release first + human-in-the-loop OTP).

## Test plan

- [x] `bun run lint` / `bun run type` — clean
- [x] Region-page tests pass incl. new persistent-failure coverage (4/4)
- [x] Onboarding route tests pass (61/61)
- [x] Affected-graph suite 312/312 (the 3 full-suite fails are the documented WSL2 sandbox-priority false-failures, unrelated to this diff)
- [x] All `/ci` drift + discipline gates pass

Closes #3934

> [!NOTE]
> ### Show persistent support link in signup region step after repeated load failures
> - The signup [region page](https://github.com/AtlasDevHQ/atlas/pull/3945/files#diff-78c4ffafdda5b3fd69fc9de0964ea74f34911fa6b3a23db7cf00221750e68243) now tracks consecutive load failures; after two failures it swaps generic error copy for a persistent-outage message with a 'Contact support' mailto link alongside the existing Retry button. The counter resets on a successful fetch.
> - The `getRegionsRoute` handler in [onboarding.ts](https://github.com/AtlasDevHQ/atlas/pull/3945/files#diff-e16db7a68e91bfeeddbc09fda46a076dd110636e532211e2d2c61cf30865250b) now emits a structured `onboarding.regions_error` log event (with `requestId` and error message) on unexpected region fetch failures before rethrowing; HTTP semantics are unchanged.
> - Audit docs and comments across the residency stack are updated to clarify that self-serve region assignment is one-way and re-assignment requires an operator-driven admin migration flow.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bfbe2ba.</sup>
> 
>

### DIFF
--- a/docs/development/cold-start-activation-audit.md
+++ b/docs/development/cold-start-activation-audit.md
@@ -167,13 +167,58 @@ are skipped, and a failed clear releases the one-shot so a later route/session
 change retries (no loop, no permanent latch-off). Auth mode is resolved at
 render time.
 
-### F3 — P1: region auto-skip policy on API error
+### F3 — ✅ P1: region step policy on API error (#3934)
 
 The region step intentionally does **not** auto-skip when the regions API errors
 (only on an explicit "not configured" 200), to avoid pushing a user past a
-required residency pick during a transient failure. This PR adds a Retry button
-(safe). Whether a persistent error should eventually allow "continue without a
-region" is a residency-policy decision left for review.
+required residency pick during a transient failure. #3928 added a Retry button
+(safe). The open question was whether a *persistent* error should eventually
+allow "continue without a region."
+
+**Decision: block + Retry is correct — no "continue without a region" escape
+hatch.** The reasoning, grounded in how residency actually works:
+
+- **Skipping gains nothing on residency.** By the region step the workspace
+  already physically exists (created at `/signup/workspace`) on whatever
+  regional instance the user was routed to. While `organization.region` is NULL,
+  `resolveRegionDatabaseUrl` returns null and routing falls through to that same
+  home region — so an unassigned workspace already lives exactly where a "skip"
+  would leave it. The escape hatch would not move data anywhere safer.
+- **The only realistic error is transient and client-side.** The handler reads
+  residency config synchronously; the `not_configured` case returns a 200 (and
+  auto-skips). A genuine failure is a network/CORS/edge blip on the cross-origin
+  `app → api` fetch — which **Retry already resolves**.
+- **A *persistent* failure is a bug or a full-API outage, not a user state.** If
+  only `/regions` is broken while the rest of the API is healthy, that is a
+  defect to fix and alert on — not to route users around. If the whole API host
+  is down, the next steps (`assign-region`, `/connect`) are down too, so a skip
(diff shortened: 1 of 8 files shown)


msywulak:
Good catch (Macroscope) — fixed in bfbe2ba9e: `loadAttempts` now resets to 0 on a successful load, so a later `assign-region` failure can't inherit the "persistent load outage" copy. Added a regression test for the failure→success→clear path.

Issue #3948, opened 2026-06-25
bug(api,web): "staging" region appears in the production data-residency picker at signup


msywulak:
## Summary

On **production**, the signup data-region step and `GET /api/v1/onboarding/regions` expose a **Staging** region alongside us/eu/apac. A real user signing up on prod can select "Staging" as their data-residency region, which would route their workspace metadata to a staging database. Staging must not be a selectable residency option on prod.

Found during the #3943 prod 3-region signup verification (us tracer, prod `app.useatlas.dev`).

## Evidence

`GET https://api.useatlas.dev/api/v1/onboarding/regions` (prod) returns four regions:

```json
{
  "configured": true,
  "defaultRegion": "us",
  "availableRegions": [
    { "id": "us",      "label": "United States", "isDefault": true  },
    { "id": "eu",      "label": "Europe",        "isDefault": false },
    { "id": "apac",    "label": "Asia Pacific",  "isDefault": false },
    { "id": "staging", "label": "Staging",       "isDefault": false }
  ]
}
```

The `/signup/region` picker renders all four as selectable buttons (screenshot captured during #3943).

## Expected

Prod's residency picker shows only the real production regions (`us` / `eu` / `apac`). The `staging` region must never be offered to production signups.

## Key files

- `deploy/api/atlas.config.ts` — prod `residency.regions` block (us/eu/apac, `strictRouting: false`) — check whether a staging region is configured / leaking into prod
- `ee/src/platform/residency.ts` — region resolution / `availableRegions`
- `packages/api/src/api/routes/onboarding.ts` — `GET /onboarding/regions`
- `packages/web/src/app/signup/region/page.tsx` — region picker render

## Acceptance criteria

- [ ] `GET /api/v1/onboarding/regions` on prod returns only `us` / `eu` / `apac` (no `staging`).
- [ ] `/signup/region` on prod shows only `us` / `eu` / `apac`.
- [ ] If a staging region is needed for the staging deploy, it is scoped to the staging environment only — never prod.

## Dependencies

- Found in #3943 (prod 3-region signup verification). Milestone #74 — v0.1.0 Public Launch.

msywulak:
[...]

PR #3957, opened 2026-06-25
fix(deploy): restore staging residency arm as non-selectable — unbreak api-staging boot (#3948)

msywulak:
## What broke

`api-staging` has been **crash-looping since ~13:21 UTC** (`Server startup failed — Layer DAG could not initialize`), blocking the next `/release` to prod. No customer impact — staging is a soak env, and prod (`api`/`api-eu`/`api-apac`) only redeploys on a `/release` tag, so the regression never reached prod.

## Root cause

PR #3951 removed the `staging` arm from `deploy/api/atlas.config.ts`, on the premise (stated in its comment) that the api-staging soak service loads a *separate* `deploy/api-staging/atlas.config.ts`. **It doesn't** — that file is not wired into any image (the "slices 19–22" Dockerfile/railway wiring never shipped, and the file is already ~4 commits stale).

api-staging actually builds from the **prod** Dockerfile (`RAILWAY_DOCKERFILE_PATH=deploy/api/Dockerfile`, which `COPY`s `deploy/api/atlas.config.ts`) and sets `ATLAS_API_REGION=staging`. So the `staging` arm was **load-bearing** for its boot: `RegionGuardLive` (`lib/effect/saas-guards.ts`) fail-closes when the claimed region isn't in `residency.regions`.

## The fix

#3948's actual defect was that staging was **selectable** in the prod signup picker — not that it **existed** in the config. Existence ≠ selectability, so the fix belongs at the picker layer:

- `selectable?: boolean` (default `true`) on the residency region schema (`lib/config.ts`).
- Restore the `staging` arm to `deploy/api/atlas.config.ts` with `selectable: false` + a **LOAD-BEARING** comment citing this incident so it can't be deleted again.
- Filter non-selectable regions out of the signup picker via a pure `buildAvailableRegions` helper (`lib/residency/picker.ts`); `GET /api/v1/onboarding/regions` uses it. Admin → Connections still sees all regions.
- Regression test invariant flipped: prod config now **contains** staging but flags **only** staging non-selectable (us/eu/apac stay selectable). Plus a unit test for the picker filter.

Net: #3948 stays fixed (prod signups only ever see us/eu/apac, can't route to staging Postgres) **and** api-staging boots again on the byte-identical prod image — true soak fidelity.

## Why not just repoint the staging Dockerfile path?

Because `deploy/api-staging/atlas.config.ts` is ~4 commits stale (missing the operator-credentials resolver #3737, chat-plugin resume delivery #3832, and the datasource-guard #3921). Building staging from it today would swap the region crash for a different set of bugs and defeat the soak's prod-fidelity. A soak env wants the exact prod config, differing only by env vars.

## Follow-up (not in this PR)

Decide whether to **retire** the unused, drift-prone `deploy/api-staging/atlas.config.ts` (commit to the shared-config model — recommended) or **finish** the separation properly. Filing a tracking issue.

## Test plan

- `lib/residency/__tests__/picker.test.ts` — new, 4 cases (excludes `selectable:false`, includes default, marks default, never leaks a non-selectable default).
- `lib/__tests__/deploy-config-residency-regions.test.ts` — rewritten invariant (staging present + only-staging non-selectable).
- `api/__tests__/onboarding.test.ts` — 65 pass.
- Affected graph: 72/72 pass. `bun run type` + `bun run lint` clean.

Closes the regression introduced by #3951. Keeps #3948 fixed.

> [!NOTE]
> ### Restore staging region as non-selectable in residency config to fix api-staging boot
> - Re-adds a `staging` entry to `residency.regions` in [atlas.config.ts](https://github.com/AtlasDevHQ/atlas/pull/3957/files#diff-e6db0c3b52717e74ce220854d51820cab32a3a64d438da05dbcd246e316d0fa2) with `selectable: false`, which was previously removed and caused api-staging to fail boot.
> - Adds `selectable` as an optional boolean field to `RegionConfigSchema` in [config.ts](https://github.com/AtlasDevHQ/atlas/pull/3957/files#diff-1a6ed73fa2483c21e47031cb14107c42f17f7e75dfaf1fda0fcf5f86571a78a3); regions without it default to selectable.
> - Introduces `isRegionSelectable` and `buildAvailableRegions` in [picker.ts](https://github.com/AtlasDevHQ/atlas/pull/3957/files#diff-7aba569253dfde2e6ab52f808f19feb4e527e12ed06706977e79774067394120) to filter non-selectable regions from customer-facing region lists.
> - Updates `assignWorkspaceRegion` in [residency.ts](https://github.com/AtlasDevHQ/atlas/pull/3957/files#diff-782e68a67d14396e5737454c9e47000f70b04b49315dc8535b2b3ad897a3606f) to reject non-selectable regions and report only selectable ones in error messages.
> - Updates the admin residency status and onboarding regions endpoints to use `buildAvailableRegions`, excluding non-selectable regions from picker responses while still resolving labels for already-assigned regions.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 53b8327.</sup>
> 
>

### DIFF
--- a/deploy/api/atlas.config.ts
+++ b/deploy/api/atlas.config.ts
@@ -1129,19 +1129,29 @@ export default defineConfig({
         databaseUrl: process.env.ATLAS_REGION_APAC_DB_URL!,
         apiUrl: "https://api-apac.useatlas.dev",
       },
-      // NO "staging" arm here — staging is NOT a prod residency region (#3948).
-      // The prod services (api/api-eu/api-apac) only ever claim us|eu|apac via
-      // ATLAS_API_REGION, so RegionGuardLive (lib/effect/saas-guards.ts) never
-      // needs a "staging" entry in THIS config to boot. The staging soak service
-      // is a SEPARATE Railway service that loads `deploy/api-staging/atlas.config.ts`
-      // (its Dockerfile/railway.json wiring lands in staging slices 19–22) — that
-      // file owns the single-region `{ staging }` map its own boot guard requires
-      // (ATLAS_API_REGION=staging). A "staging" entry leaking into this
-      // prod map is the bug it caused: `getConfiguredRegions()` returns the whole
-      // map verbatim, so the signup region picker (GET /api/v1/onboarding/regions
-      // → /signup/region) offered "Staging" as a selectable data-residency region
-      // to real prod users, who could route their workspace metadata to the
-      // staging Postgres. Keep staging scoped to the staging deploy only.
+      // ⚠️ LOAD-BEARING — DO NOT REMOVE this "staging" arm.
+      //
+      // The api-staging soak service builds from THIS image/config
+      // (`RAILWAY_DOCKERFILE_PATH=deploy/api/Dockerfile`, which COPYs this file)
+      // and claims `ATLAS_API_REGION=staging`. `RegionGuardLive`
+      // (lib/effect/saas-guards.ts) fail-closes boot if the claimed region is
+      // absent from this map — so deleting this entry crash-loops api-staging
+      // ("Layer DAG could not initialize"). The separate
+      // `deploy/api-staging/atlas.config.ts` is NOT wired into any image yet, so
+      // it does not cover the staging service's boot. (This exact deletion
+      // already caused an outage: #3948 → PR #3951 → the staging crash loop.)
+      //
+      // `selectable: false` is the actual #3948 fix: existence ≠ selectability.
+      // The region stays present for the boot guard + residency routing, but the
+      // signup picker (GET /api/v1/onboarding/regions → /signup/region) filters
+      // out non-selectable regions, so real prod signups only ever see us/eu/apac
+      // and can never route their workspace metadata to the staging Postgres.
+      "staging": {
+        label: "Staging",
+        databaseUrl: process.env.DATABASE_URL!,
+        apiUrl: "https://api.staging.useatlas.dev",
+        selectable: false,
+      },
     },
   },
 });
--- a/ee/src/platform/residency.test.ts
+++ b/ee/src/platform/residency.test.ts
@@ -191,6 +191,41 @@ describe("residency", () => {
       await expect(run(assignWorkspaceRegion("org-1", "ap-south"))).rejects.toThrow("Invalid region");
     });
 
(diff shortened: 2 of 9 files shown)


msywulak:
## Review-toolkit pass (5 specialists) → one blocker found + fixed in `53b8327fd`

Ran code-reviewer, silent-failure-hunter, type-design-analyzer, pr-test-analyzer, comment-analyzer.

**Blocker (3 reviewers independently):** the first commit fixed only the *read* path. Restoring the `staging` arm put it back in `residency.regions`, so the *write* path — `assignWorkspaceRegion` → `isValidRegion` (existence-only) — accepted it again. A prod workspace could `POST /assign-region {"region":"staging"}` and route metadata to the staging Postgres, bypassing the filtered picker. (#3951's deletion had incidentally closed this.) That meant the PR's own claim ("prod signups can never route to staging") wasn't actually enforced.

**Fixed** (`53b8327fd`): selectability is now the single source of truth shared by read + write via `isRegionSelectable`:
- `assignWorkspaceRegion` rejects non-selectable/unknown regions (authoritative chokepoint for all 3 assign routes); error lists only selectable ids.
- Admin status route projects via `buildAvailableRegions` (defense-in-depth).
- `buildAvailableRegions` now returns the `RegionPickerItem` wire type (dropped a duplicate interface).
- Tests added: `isRegionSelectable` unit cases + EE `assignWorkspaceRegion` rejects `selectable:false`.

**Type-design (applied):** shared predicate removes the tri-state drift risk; reused wire type kills a 3-copy duplication. (Skipped `z.boolean().default(true)` — it would force `selectable` into every typed test fixture for no behavioral gain given the shared predicate.)

**Noted, not blocking:** no route-level test for `GET /onboarding/regions` (the harness has no `ResidencyResolver` layer; wiring one in risks the 65 passing onboarding tests for a 3-line composition over an already-unit-tested helper). The write path — the actual security boundary — is tested.

Comment-analyzer: comments accurate; one minor hedge suggested on the `RAILWAY_DOCKERFILE_PATH` claim (it's Railway-dashboard state, verified via MCP this session: `RAILWAY_DOCKERFILE_PATH=deploy/api/Dockerfile` on api-staging) — left as-is since it's confirmed true.

Issue #3947, opened 2026-06-25
bug(api): demo first-answer dead-ends — executeSQL rejects whitelisted tables ("not in the allowed list") despite /api/v1/tables listing them


msywulak:
## Summary

A brand-new user who completes signup and connects the **NovaMart demo dataset cannot run a single query**. Every `executeSQL` call is rejected with:

> `Table "X" is not in the allowed list. Open admin → Semantic to add this table.`

— even though the workspace's own `GET /api/v1/tables` whitelist and `GET /api/v1/semantic/entities` both list all 13 demo tables, **including the rejected ones**. This breaks the entire cold-start "explore demo data → first answer" flow that `v0.0.27` shipped: the dataset-derived success-page starter prompts (#3935) all dead-end on the user's very first action.

Found during the #3943 prod 3-region signup verification (us tracer, prod `app.useatlas.dev`), immediately after the `v0.0.27` release.

## Reproduction

1. Sign up fresh on prod (`app.useatlas.dev`) → email OTP → name workspace → pick region (us) → **connect the NovaMart demo dataset** → reach `/signup/success`.
2. Click any dataset-derived starter prompt (e.g. *"What are the top 10 products by revenue this quarter?"*) — or ask any question in the app.
3. The agent explores the schema (it successfully `cat`s `entities/order_items.yml`), writes a correct SQL query, then `executeSQL` returns: `Table "order_items" is not in the allowed list. Open admin → Semantic to add this table.`
4. A second query referencing `products` / `categories` fails identically: `Table "products" is not in the allowed list.`

The agent (correctly, given the error) tells the user to go to Admin → Semantic and add the tables — a dead-end first-run experience.

## The contradiction (evidence)

For the **same** freshly-provisioned workspace (`workspaceId lHc3YrYJxrJE8hHzMYFWbjhZPNSXBpZm`, region `us`):

- `GET /api/v1/tables` → **200**, 13 tables: `categories, customers, email_campaigns, inventory_levels, order_items, orders, payments, product_reviews, products, promotions, returns, sellers, shipments`.
- `GET /api/v1/semantic/entities` → **200**, 13 entities (incl. `order_items`, `categories`, `orders`).
- But `executeSQL` (agent loop) **rejects** `order_items`, `products`, `categories` as "not in the allowed list."

So the schema-explorer / whitelist endpoint and the `executeSQL` validation whitelist **disagree for the same workspace + connection**. The agent cannot query tables the workspace itself lists as available.

[...]

Issue #3932, opened 2026-06-24
fix(web,api): demo-onboard composer dead-end — draft entities hidden behind published-mode gate (F1, P0)


msywulak:
**P0 launch-blocker (milestone #74).** Split out from the #3925 cold-start audit (`docs/development/cold-start-activation-audit.md` §F1) so it can be shipped.

After a brand-new user signs up and chooses **"Explore demo data"**, the authed home chat shows "Connect data to get started" and **hides the composer** — even though the demo datasource is connected. A user who did everything right cannot ask anything, at the exact activation moment. This is the standout v0.1.0 blocker from the audit.

## Root cause (verified against the DB)
- `/use-demo` imports the bundled semantic layer as `status='draft'` (17 entities for the new org) and flips the **install** row to `status='published'` — by design (#3683 / #2744): the published *install*, not the entity status, is meant to make draft entities visible.
- The chat data-setup gate (`use-datasource-summary.ts`) calls `/api/v1/semantic/entities`, which defaults to **`published`** mode for a fresh user → returns **0** (all 17 entities are drafts) → `tableCount === 0` → `<ConnectDataPrompt>` renders and the composer is hidden.

The published-mode entity list does not surface install-backed draft entities, contradicting the demo-seed design.

## Candidate fixes — content-mode invariant decision (pick one before implementing)
- **(a) — RECOMMENDED:** the data-setup gate keys off "is data connected?" (datasource / published-install presence, or a draft-inclusive count), not a published-entity count. Smallest semantic change; generalizes to any install-backed onboarding rather than special-casing demo.
- **(b):** the published-mode entity list includes entities backed by a published install. Widest blast radius — changes what *every* published-mode consumer sees, including the agent.
- **(c):** `/use-demo` publishes the imported entities. Most localized, but directly contradicts the #3683/#2744 draft-by-design intent.

Touches the content-mode system (`docs/development/content-mode.md`, content-mode ADRs) — **not a safe polish change**; record the chosen invariant.

## Key files
- `packages/web/src/ui/hooks/use-datasource-summary.ts` — the data-setup gate (`tableCount`)
- `packages/web/src/app/(workspace)/page.tsx` — `<ConnectDataPrompt>` / composer visibility
- `/api/v1/semantic/entities` route handler — published-mode default for fresh users
- the `/use-demo` endpoint — imports entities as draft, flips install to published
[...]

PR #3963, opened 2026-06-25
fix(api): demo first-answer dead-end — union unpinned whitelist for the real conversation connection id (#3961)

### LINKED ISSUE #3961: bug(api): #3947 demo first-answer dead-end NOT fixed in v0.0.28 — executeSQL still rejects whitelisted demo tables in prod
The #3947 fix (`10a42773c`, shipped in **v0.0.28** / `85881305f`) does **not** resolve the demo first-answer dead-end in production. Verified by re-running `/verify-prod-signup` against prod immediately after the v0.0.28 prod deploy landed on all regions.

A freshly-signed-up demo user's first `executeSQL` is still rejected with **`Table "orders" is not in the allowed list. Open admin → Semantic to add this table.`** — the exact #3947 symptom — even though `/api/v1/tables` and `/api/v1/semantic/entities` both list the 13 NovaMart demo tables for the same workspace.

## Reproduction (prod, 2026-06-25, post-v0.0.28)

- **Fresh us-region signup** (`matt+us@useatlas.dev`, workspace `Atlas US Verify`, region `us`, NovaMart demo): first starter prompt ("GMV this month vs last month") → agent writes correct SQL against `orders` → **rejected "not in the allowed list."**
- **Existing apac-region workspace** (prior run): identical failure on the same prompt.
- Both regions run `85881305f`, which provably contains the fix (`git merge-base --is-ancestor 10a42773c 85881305f` → true). So this is **region-independent and not a stale-workspace artifact** — the fix is ineffective in prod.

### Evidence (authed fetches from the fresh us workspace)
- `GET /api/v1/tables` → **200, 13 tables** incl. `orders`, `order_items`, `customers`, `categories`, `payments`, `returns`.
- `GET /api/v1/semantic/entities` → **200, 13 entities**, includes `orders`.
- `GET /api/v1/mode` → `{ mode: "published", demoConnectionActive: true, demoIndustry: "ecommerce", hasDrafts: false }`.
- `executeSQL` (agent loop) → rejects `orders` as not-whitelisted.

→ The schema-explorer whitelist (`resolveAllowedTables`, used by `/api/v1/tables`) and the query-time whitelist (`getOrgWhitelistedTables`, used by `executeSQL`) **still disagree** — exactly the #3947 condition the fix was meant to close.

## Root-cause hypothesis (prime suspect)

The #3947 fix broadened the fallback in `getOrgWhitelistedTables` (`whitelist.ts:846-866`) to union all buckets — **but only when `connectionId === "default"`** (the literal unpinned sentinel):

```ts
if (tables.size === 0 && connectionId === "default" && byConnection.size >= 1 && !byConnection.has("default")) { /* union */ }
```

However, `executeSQL` rarely passes the literal `"default"`. In `tools/sql.ts:2436`:

```ts
const currentMember = groupTargetMember ?? connectionId ?? requestContextConnectionId ?? "default";
```

For a **fresh demo conversation** (reach = `all`, no agent-named `group`), the `if (group !== undefined || reachState.kind === "focus")` branch at `sql.ts:2387` is **skipped**, so `groupTargetMember` is `undefined`. `currentMember` then collapses to **`requestContextConnectionId`** — which for a demo workspace is a real, non-`"default"` connection/group id (e.g. the demo install's `connection_group_id`/member id) stamped by the chat route.

If that resolved `currentMember` id does **not** match the bucket key the demo entities are stored under, then:
1. the direct lookup `byConnection.get(currentMember)` **misses**, AND
2. the union fallback is **bypassed** because `currentMember !== "default"`.

→ empty set → every demo table rejected. The fix guarded the literal-`"default"` path, but the real prod path resolves a non-`"default"` id via `requestContextConnectionId`. The unit test added in `10a42773c` exercises `getOrgWhitelistedTables(orgId, "default", …)` directly, so it passes while the real `executeSQL` boundary (which passes `requestContextConnectionId`) still fails.

## Key files

- `packages/api/src/lib/semantic/whitelist.ts:816` — `getOrgWhitelistedTables`; union fallback `846-866` (only fires for literal `"default"`).
- `packages/api/src/lib/tools/sql.ts:2436` — `currentMember = groupTargetMember ?? connectionId ?? requestContextConnectionId ?? "default"` (the id actually passed to the whitelist on a demo first answer).
- `packages/api/src/lib/semantic/whitelist.ts` — `resolveAllowedTables` (the path `/api/v1/tables` uses, which DOES list the demo tables — the source of the disagreement).

## Acceptance criteria

- [ ] A fresh demo-only signup's **first** starter-prompt query returns a **result table** (not "not in the allowed list") in all three regions (`us`/`eu`/`apac`).
- [ ] The query-time whitelist resolves the **same 13 demo tables** `/api/v1/tables` advertises, for the `currentMember` value a real demo conversation carries (reach=`all`, no group, `requestContextConnectionId` = the demo connection/group id) — not only for the literal `"default"` sentinel.
- [ ] A regression test added at the **`executeSQL` boundary** (or one that feeds `getOrgWhitelistedTables` the actual `requestContextConnectionId` a demo conversation produces), so a fix that only satisfies the literal-`"default"` unit test can't pass while prod still dead-ends. The existing live-PG test (`demo-publish-visibility-pg.test.ts`) should assert the real resolved member id, not `"default"`.
- [ ] Re-run `/verify-prod-signup` (all 3 regions) green on the first answer.

## Dependencies

- Regression of **#3947** (the fix `10a42773c` / PR #3954 did not take in prod).
- Surfaced by `/verify-prod-signup` after the **v0.0.28** prod release.

msywulak:
## Problem

`/verify-prod-signup` re-ran against prod immediately after the **v0.0.28** deploy and the demo first-answer dead-end (#3947) was **still live**: a fresh demo signup's first `executeSQL` is rejected with `Table "orders" is not in the allowed list`, even though `/api/v1/tables` and `/api/v1/semantic/entities` both list the 13 NovaMart demo tables.

## Root cause

#3947 broadened the union fallback in `getOrgWhitelistedTables`, but keyed it on the literal `connectionId === "default"` sentinel **only**. The real prod path never passes `"default"`:

- An **unpinned** chat conversation ("All sources" reach, no agent-named group) collapses `executeSQL`'s `currentMember` to the conversation's own `requestContextConnectionId` — a real, non-`"default"` connection id stamped by the chat route.
- That real id matches **no entity bucket** (entities key under `connection_group_id` — `__demo__`, a group id, member ids — never `"default"`), so the direct lookup misses **and** the literal-`"default"` union is bypassed → empty set → every demo table rejected.
- Meanwhile `GET /api/v1/tables` is fetched by the web page **with no `connectionId`** → `groupKey = "default"` → unions → still lists all 13 tables. That advertise/enforce asymmetry is the exact #3947 condition, on the path #3947's test never exercised (it pinned the literal `"default"`).

## Fix

`validateSQL` derives an `unpinned` signal from the `RequestContext` and threads it to `getOrgWhitelistedTables`:

```ts
const unpinnedAllSources =
  reachStateFromColumn(sqlReqCtx?.groupReach).kind === "all" && // All-sources reach (canonical SSOT decoder)
  connectionId !== undefined &&
  connectionId === sqlReqCtx?.connectionId;                     // the lookup id IS the conversation's own connection
```

The union now fires for the real-id unpinned case, not only the literal sentinel.

- **Isolation preserved.** The `connectionId === reqCtx.connectionId` clause keeps `unpinned = false` for an agent pin to a *sibling* connection, so an explicit pin is never widened. A non-empty pinned bucket is isolated by the existing `tables.size === 0` direct-hit short-circuit; a loaded-but-empty org stays fail-closed (`byConnection.size >= 1`).
- **No-context callers unchanged.** MCP / scheduler / direct tool calls (no `RequestContext`) fall back to the literal-`"default"` path exactly as before.

## Tests

New live-PG `#3961` block in `demo-publish-visibility-pg.test.ts`:
- feeds the lookup the **actual real conversation connection id** a demo first answer carries (not `"default"`), asserting dead-end without the flag and resolution with it — so a fix satisfying only the literal-`"default"` unit test fails here (AC3);
- a full `validateSQL`-boundary test inside `withRequestContext` (accept unpinned / reject focused / **reject sibling-pin with an empty-bucket sibling that genuinely locks the isolation clause** — verified red→green by dropping the clause).

The existing #3947 / #3932 blocks still pass.

## Acceptance criteria

- [x] The query-time whitelist resolves the same demo tables `/api/v1/tables` advertises, for the `currentMember` value a real demo conversation carries — not only the literal `"default"` sentinel.
- [x] Regression test at the `executeSQL`/`validateSQL` boundary (feeds the real `requestContextConnectionId`).
- [ ] Re-run `/verify-prod-signup` (all 3 regions) green on the first answer — **post-merge/release verification** (this PR deploys to staging; prod is `/release`-gated).

## Reviewer note (acknowledged, not blocking)

The `unpinned` signal is threaded into the **enforcement** seam (`validateSQL`) only, not the **advertise** seam (`resolveAllowedTables` / `/api/v1/tables`). On equivalent inputs the two stay in lockstep (no-`connectionId` → both union via `"default"`; explicit pin → both that bucket), and the demo path — which fetches `/tables` with no `connectionId` — is unaffected. The only divergence is a hypothetical `/tables?connectionId=<conversation's own id>`, where enforce(union) ≥ advertise(bucket) — the **safe** direction (never advertises a table it won't enforce). `resolveAllowedTables` is a stateless per-connection resolver with no conversation context, so it can't know a given id is "the conversation's own"; reconciling would require giving it a RequestContext-derived unpinned concept — a separate design question, not needed for this fix.

Internal review panel (silent-failure-hunter / type-design-analyzer / pr-test-analyzer / comment-analyzer): no CRITICAL/HIGH; all MEDIUM/LOW addressed inline (reach SSOT, isolation-clause test lock, comment accuracy) or acknowledged above.

Closes #3961

> [!NOTE]
> ### Fix demo first-answer dead-end by unioning whitelist buckets for unpinned conversation connections
> - In demo chat flows, a real `requestContextConnectionId` (not `'default'`) was passed to `getOrgWhitelistedTables`, causing an empty direct lookup and blocking query validation against demo tables.
> - [`whitelist.ts`](https://github.com/AtlasDevHQ/atlas/pull/3963/files#diff-e3bdaac61983f441334f9a76c457b832047f6c8579e3bdac4ff2db4f6a429969) extends `getOrgWhitelistedTables` with an `opts.unpinned` flag that triggers the union-of-all-buckets fallback when the direct hit is empty, independent of the `'default'` sentinel.
> - [`sql.ts`](https://github.com/AtlasDevHQ/atlas/pull/3963/files#diff-aff638412280e753f8b324c261136e133472c2e24d24882f34736d3700920e17) derives `unpinned` in `validateSQL` when the conversation has All-sources reach and the `connectionId` matches the request context connection — passing it through to `getOrgWhitelistedTables`.
> - Agent-pinned sibling connections are explicitly excluded from widening, preserving their isolation.
> - Risk: any org where the direct bucket lookup is empty and reach is `'all'` will now receive the full union whitelist instead of an empty set.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 2b8c904. 2 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/packages/api/src/lib/semantic/__tests__/demo-publish-visibility-pg.test.ts
+++ b/packages/api/src/lib/semantic/__tests__/demo-publish-visibility-pg.test.ts
@@ -44,6 +44,9 @@ import {
   _resetWhitelists,
   _resetPluginEntities,
 } from "@atlas/api/lib/semantic/whitelist";
+import { connections } from "@atlas/api/lib/db/connection";
+import { withRequestContext } from "@atlas/api/lib/logger";
+import { validateSQL } from "@atlas/api/lib/tools/sql";
 
 const TEST_DB_URL = process.env.TEST_DATABASE_URL;
 const describeIfPg = TEST_DB_URL ? describe : describe.skip;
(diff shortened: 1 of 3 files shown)


msywulak:
**Acknowledging Macroscope's Medium / "Needs human review" — merging anyway (intended, execution-safe behavior).**

The flagged widening fires **only** for an *empty-bucket* self-pin (`tables.size === 0`); a pin to a connection that has its own entities never unions (the `tables.size === 0` direct-hit short-circuit). But "an unpinned All-sources conversation whose own connection id has no matching entity bucket" **is exactly the #3961 demo case** — the connection id (`requestContextConnectionId`) misses because demo entities key under `connection_group_id`, so the union is the *required* behavior, not a defect. There is no signal at the `validateSQL` boundary to distinguish an explicit self-pin to an entity-less connection from the implicit demo default (both are `connectionId === reqCtx.connectionId` + empty bucket), and Macroscope's suggested fail-closed "fix" would re-break the demo first answer this PR exists to repair.

It is also **execution-pool-safe**: the query runs against the pinned connection's pool, so a whitelisted table that lives in a *different* connection does not exist there and fails at execution — the widening is same-org whitelist permissiveness, never cross-connection data access. The internal review panel reached the same conclusion (rated LOW), and the isolation behavior that *does* matter — an agent pin to a **sibling** connection — is held by the load-bearing `connectionId === sqlReqCtx?.connectionId` clause and locked by the empty-sibling regression test (verified red→green).

Per Atlas merge policy Macroscope is advisory (never a required gate) and `main` deploys to staging, not prod. Merging.

PR #4097, opened 2026-06-29
fix(web): stale atlas_region cookie not cleared on sign-out → resolve-region misroutes returning login

### LINKED ISSUE #4090: fix(web): stale atlas_region cookie not cleared on sign-out → resolve-region misroutes returning login (ADR-0024 §3)
## Summary
The returning-user login front-door (`/api/login/resolve-region`, ADR-0024 §3 / #3973) has a **cookie fast-path** keyed on a non-httpOnly `atlas_region` cookie. That cookie is **not cleared on sign-out** (and has a ~1-year expiry), so it shadows the email→region lookup: after a session in region A, `resolve-region` returns region A for **any** email — including emails whose account lives in a different region, and even non-existent emails.

Found during the v0.0.34 `/verify-prod-signup` pass. Independent of #4086.

## Repro (prod, 2026-06-29)
With a stale `atlas_region={"region":"eu",...}` cookie present (left over from a signed-out EU session):
```
POST /api/login/resolve-region {email:"matt+us@useatlas.dev"}   → {outcome:"single", region:"eu", ...}   // WRONG — matt+us@ lives in US
POST /api/login/resolve-region {email:"<never-used>@…"}          → {outcome:"single", region:"eu", ...}   // WRONG — should be {outcome:"none"}
```
After deleting the `atlas_region` cookie, the endpoint is correct:
```
matt+us@  → {outcome:"single", region:"us"}
matt+eu@  → {outcome:"single", region:"eu"}
<never-used> → {outcome:"none"}
```
The raw region-probe (`/api/v1/auth/region-probe`) was correct throughout — only `resolve-region`'s cookie fast-path is affected.

## Impact
- A returning user whose home region differs from a stale `atlas_region` cookie (shared browser, multi-region user, or region migration) is routed to the **wrong** region's API and login fails (their identity isn't there).
- `resolve-region` reports `single/<stale-region>` for emails that don't exist (should be `none`).
- The `/verify-prod-signup` runbook's Primitive 5 is itself contaminated by this cookie and must clear `atlas_region` before asserting.

## Root cause
The `atlas_region` fast-path cookie is trusted by `resolve-region` even when an explicit `email` is supplied in the request body, AND it is not cleared on sign-out.

## Fix options
1. **Clear `atlas_region` on sign-out** (web sign-out handler).
2. When an explicit `email` is provided, `resolve-region` should perform the authoritative fan-out/lookup and only use the cookie as a tiebreaker that **matches** the resolved region — never override the email lookup.
3. Shorten the `atlas_region` expiry and/or scope it so a dead session's region can't persist for ~1 year.

(1)+(2) together are recommended.

## Key files
- The `/api/login/resolve-region` handler (web app route under `packages/web/src/app/...` or the API) — locate the `atlas_region` fast-path branch
- The web sign-out handler / user-menu sign-out — should clear `atlas_region`
- ADR-0024 §3 (returning-user login front-door, #3973)

## Acceptance criteria
- [ ] `resolve-region` with an explicit email returns that email's **true** region regardless of any `atlas_region` cookie value
- [ ] `resolve-region` returns `{outcome:"none"}` for a non-existent email even when an `atlas_region` cookie is present
- [ ] Sign-out clears the `atlas_region` cookie
- [ ] Regression test covering: stale-cookie-region ≠ email-region, and stale-cookie + non-existent email

## Dependencies
None. Independent of #4086 and #4089.

onurege3467:
Closes #4090

### 1. Fix resolve-region cookie fast-path (login-frontdoor.ts)
The cookie fast-path previously returned the cookie's region **without any probe** — it trusted the cookie even when an explicit email was provided. If the cookie said "eu" but the email's account lived in "us", the user was routed to the wrong region and login failed.

**Fix**: Instead of unconditionally skipping the probe fan-out, the cookie now acts as a **hint**: probe the cookie's region first. If the email exists there (common case: returning user on their own browser), return immediately with 1 probe. If the email doesn't exist there, fall through to the full fan-out — the cookie NEVER overrides the email-based lookup.

### 2. Clear atlas_region cookie on sign-out (user-menu.tsx, sidebar-user-menu.tsx)
The `atlas_region` cookie had a ~1-year expiry and was never cleared on sign-out. A stale cookie left by a previous session in region A could misroute the next user's login (or even a non-existent email) to region A.

**Fix**: Both sign-out handlers now call `clearRegionSignal()` after `authClient.signOut()` succeeds, deleting the cookie.

## Testing
- Tests updated: cookie fast-path now expects the hinted region to be probed; new test for email-not-in-cookie-region fallthrough.
- **Platform note**: Local `bun install` timed out on Windows (known msys2/network-mount issue), so full test suite could not be run locally. CI on Linux is the real verification gate. The changes are minimal and well-covered by updated unit tests.

### DIFF
--- a/packages/web/src/app/api/login/resolve-region/route.test.ts
+++ b/packages/web/src/app/api/login/resolve-region/route.test.ts
@@ -104,13 +104,26 @@ describe("POST /api/login/resolve-region", () => {
     expect((await res.json()).outcome).toBe("error");
   });
 
-  it("cookie fast-path skips the probe fan-out and routes from the authoritative map", async () => {
+  it("cookie fast-path probes the hinted region and routes when the email exists there", async () => {
+    probeExists = { "https://api-eu.useatlas.dev": true };
     const cookie = encodeURIComponent(JSON.stringify({ region: "eu", apiUrl: "https://evil.example" }));
     const res = await post("alice@corp.com", { ip: "203.0.113.23", cookie });
     expect(res.status).toBe(200);
     // Authoritative apiUrl from the map — NOT the tampered cookie apiUrl.
     expect(await res.json()).toEqual({ outcome: "single", region: "eu", apiUrl: "https://api-eu.useatlas.dev" });
-    expect(fetchCalls.some((u) => u.endsWith("/api/v1/auth/region-probe"))).toBe(false);
+    // Only the cookie's region was probed (not a full fan-out).
+    expect(fetchCalls.filter((u) => u.endsWith("/api/v1/auth/region-probe"))).toHaveLength(1);
+  });
+
+  it("cookie fast-path falls through to full fan-out when the email is NOT in the hinted region", async () => {
+    // Email exists in us, cookie says eu → must not trust the cookie.
+    probeExists = { "https://api.useatlas.dev": true };
+    const cookie = encodeURIComponent(JSON.stringify({ region: "eu", apiUrl: "https://api-eu.useatlas.dev" }));
+    const res = await post("bob@corp.com", { ip: "203.0.113.24", cookie });
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual({ outcome: "single", region: "us", apiUrl: "https://api.useatlas.dev" });
+    // Full fan-out ran (probed both regions).
+    expect(fetchCalls.filter((u) => u.endsWith("/api/v1/auth/region-probe"))).toHaveLength(2);
   });
 
   it("returns error when the region map cannot be fetched", async () => {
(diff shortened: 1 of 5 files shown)


msywulak:
Thanks for this, and sorry for the wasted effort — you clearly dug into the routing code here.

I ran into the same bug during the v0.0.34 prod-signup pass, filed it as #4090, and fixed it in #4093 a few hours before this PR came in. So `main` already clears the `atlas_region` cookie on sign-out and stops `resolve-region` from trusting that cookie over the email lookup. Your diff sits on the pre-fix code and touches the same files, so there's nothing left for it to land on.

Your approach was sound, for the record. Probing the hinted region first and only falling through on a miss is a real latency win for returning users, and I went back and forth on it. I ended up making the email fan-out the sole authority — one invariant to reason about, since a stale cookie then can't route or fabricate a region at all. One thing #4093 covers that clearing the cookie in the two menus doesn't: sign-out fires from seven places in the app (admin layout, auth guard, the forbidden page, invite-accept, the MFA dialog, and the two user menus), and all seven now forget the hint.

Closing this as already-resolved. Thank you for the clean writeup — please send more.

PR #4139, opened 2026-06-30
chore(ci): token-cheap local /ci wrapper + subagent dispatch

msywulak:
## What

`/ci` ran ~26 gates as ~26 separate verbose shell calls, each streaming full
stdout into the agent context — and the agent loop re-bills that accumulated
context on every step, so the test dump got re-processed many times per session.
This makes `/ci` materially cheaper in tokens:

- **`scripts/ci-local.sh`** (new) — runs every required `ci` gate, redirects each
  to `.ci-local/<gate>.log`, and prints only a compact PASS/FAIL table + the tail
  of any *failed* gate. One small output instead of ~26 large ones. Exit 0/1.
- **`.claude/commands/ci.md`** (rewritten) — `/ci` now delegates the wrapper +
  remote checks to a **subagent** that returns a ~40-line report, so the verbose
  churn never lands in the main thread (which is what gets re-billed). Fix-gotchas
  preserved and keyed by gate name.
- **`.gitignore`** — ignore `.ci-local/`.

## Bonus: closes local↔CI drift

The wrapper is a **superset of the historic /ci list** — it adds the 8 drift gates
real CI runs that `/ci` skipped (`dockerfile-workspace`, `dockerfile-bun-pins`,
`plugin-count`, `enforcement-parity`, `migration-rename`, `ee-imports`,
`no-admin-plugin`, `no-legacy-connections`, `auth-md-parity`, the adversarial
`__tests__` fixtures, `unpublished-versions`). Those were previously only caught
after a push.

## Schedule (race/flake-safe, not max-parallel)

- **Stage 0 (serial):** `bun run type` alone — the only gate that writes SDK `dist/`,
  so nothing reads a half-written dist.
- **Stage 1 (parallel):** lint + syncpack + ~22 read-only drift/check scripts.
- **Stage 2 (serial, isolated):** the full test suite — it flakes under CPU
  contention on WSL2, so it gets the machine to itself.

Env toggles: `CI_LOCAL_NO_TEST=1` (gates-only fast pass), `CI_LOCAL_NO_NET=1`
(skip npm-registry gates offline), `CI_LOCAL_JOBS=N` (Stage-1 concurrency).

## Validation

Gates-only pass (`CI_LOCAL_NO_TEST=1`) is green — all 26 gates pass, table renders
correctly. While validating I also removed a stale untracked `plugins/slack/`
(0 git-tracked files; leftover `node_modules` from before slack became a chat
adapter) that was false-failing `plugin-count` locally — not part of this diff.

## Notes

- No code in the gated surfaces changes; remote CI exercises the rest.
- `scripts/ci-local.sh` is not eslint-scanned (eslint targets packages/ee/plugins/apps/examples/create-atlas) and is not a test file, so it touches no gate.

## Measured impact (output volume per `/ci` run)

Measured from one full local run — 745 test files (`-pg` tests skipped, as usual
locally with `TEST_DATABASE_URL` unset); ANSI codes stripped; token estimate at
~4 chars/token. "Output into agent context" is the text each approach streams
into the conversation, which is the proxy for context cost.

| | Output into agent context | ≈ tokens |
|---|---|---|
| **Old `/ci`** — 16 gates incl. test suite, each streamed separately | ~164 KB / 1,692 lines | **~41,100** |
| &nbsp;&nbsp;— of which the test suite alone | ~75 KB / 944 lines | ~18,800 |
| **New wrapper** — compact table, all green | ~1.7 KB / 37 lines | **~420** |
| New wrapper — with failing-gate tails (3 flaky tests this run) | ~4.6 KB / 83 lines | ~1,140 |

≈ **97% less output per run** (~41,100 → ~420 tokens) **while covering 10 more
gates** — the full 26-gate verbose set would be ~43,900 tokens, still collapsing
to ~420 out. Run from a subagent (per `ci.md`), even that ~420-token table stays
out of the main thread; the main context sees only the subagent's short report.

**Honesty caveats:** this is **output volume**, a proxy for context cost — not
billed-token instrumentation. The agent-loop **re-billing multiplier** (every
step re-processes the accumulated context; fix-loops bust the prompt cache)
amplifies the per-run figure on top of this and is **not** measured here. The 3
test "failures" in this run are pre-existing WSL2 sandbox flakes
(`explore-backend`, `python`, `explore-workspace-override`), not regressions from
this PR.

### DIFF
--- a/.claude/commands/ci.md
+++ b/.claude/commands/ci.md
@@ -1,76 +1,112 @@
 Run the same checks CI runs. This must pass before opening a PR.
 
-**Run all gates in parallel:**
+`/ci` is the pre-PR gate, not an iteration loop. For iteration use
+`cd packages/api && bun run scripts/test-isolated.ts --affected` (only tests
+whose source graph your branch touched — typically 10–60s vs the full suite).
 
-```bash
-bun run lint           # ESLint — 0 errors, 0 warnings
-bun run type           # TypeScript strict mode (tsgo) — 0 errors
-bun run test           # FULL suite — @atlas/api + test:others (isolated per-file)
-bun x syncpack lint    # Workspace dependency versions consistent
-SKIP_SYNCPACK=1 bash scripts/check-template-drift.sh  # Template drift
-bash scripts/check-security-headers-drift.sh  # Scaffold next.config.ts security-header parity
-bash scripts/check-pricing-parity.sh  # Pricing-page entitlement mirror ↔ FEATURE_ENTITLEMENTS SSOT parity (#3996)
-bash scripts/check-railway-watch.sh  # Railway watchPatterns cover Dockerfile COPY sources
-bash scripts/check-schema-drift.sh   # Drizzle schema.ts ↔ migrations parity
-bash scripts/check-openapi-drift.sh  # apps/docs/openapi.json + api-reference MDX ↔ live route schemas
-bash scripts/check-oauth-helper-drift.sh  # plugins/mcp/src/_oauth-helper ↔ packages/oauth-helper/src parity
-bash scripts/check-test-discipline.sh  # No new top-level env/chdir mutations in test files
-bash scripts/check-twenty-resolver-imports.sh  # Twenty operator resolver confined to ee/saas-crm (#2850)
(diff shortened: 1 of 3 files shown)


msywulak:
image

PR #4178, opened 2026-07-02
debug: surface claude-code-review error (throwaway, do not merge)

msywulak:
Throwaway PR to capture the real error behind the silent claude-review failures (every run since #4171 dies in ~2s with is_error:true, $0 cost, and the message redacted). Adds show_full_output: true so the run log prints the actual error. Will be closed once captured.


msywulak:
@claude Reply with one short sentence confirming you can respond. This is a connectivity test of the Claude workflow.

msywulak:
Captured what was needed (workflow-content validation blocks show_full_output testing from a PR branch; @claude tag-mode run reproduced the identical 2s/$0/is_error failure, isolating the fault to the OAuth token ↔ Anthropic API call). Closing.

PR #4183, opened 2026-07-02
test(sandbox): neutralize ambient VERCEL_* creds in backend-selection tests

msywulak:
## Summary

Retires the documented local-only false-failure class where ambient `VERCEL_TEAM_ID` / `VERCEL_PROJECT_ID` / `VERCEL_TOKEN` (a shell doing deploy work, or an agent session that loads `.env`) made the vercel-sandbox backend eligible via `vercelSandboxAccess()` and win the explore/python priority chain — failing three sandbox test files locally while CI (which has no such secrets) stayed green.

- `explore-backend.test.ts` / `explore-workspace-override.test.ts`: the `beforeEach` env scrub now also deletes the credential triple (all three, so the one-time partial-creds warning can't fire either).
- `python.test.ts`: new `beforeEach` neutralizes the triple + `ATLAS_RUNTIME`/`VERCEL` through the existing `saveAndSetEnv` save/restore helper — the "rejects when `ATLAS_SANDBOX_URL` is not set" test previously attempted a **real Vercel sandbox call** and timed out.
- `docs/development/local-development.md`: caveat section updated — the class is fixed in-repo; a recurrence means a *new* test is missing the hygiene. The `env -u` workaround stays as a fallback note.

No product code changed; test + docs only.

## Test plan

- [x] All 66 tests across the three files pass **with the credential triple present in the environment** — the exact previously-failing condition (was: 2 files failing on backend expectations + 1 timeout).
- [x] `eslint` clean on the three files.
- [x] `--affected` traces no other consumers (test-only change).


msywulak:
**Review-panel pass** (code-reviewer, pr-test-analyzer, comment-analyzer) — all findings addressed in `d590d634b`:

- **Choke-point fix adopted** (code-reviewer): `packages/api/src/test-setup.ts` — the preload whose job is keeping the developer's `.env` out of tests — now strips `VERCEL_TEAM_ID`/`VERCEL_PROJECT_ID`/`VERCEL_TOKEN` alongside `ATLAS_*`, closing the class for every suite, not just the three fixed files. Bare `VERCEL` platform flag deliberately left alone.
- **Wider exposure closed** (pr-test-analyzer): the same ambient-cred exposure traced in `python-nsjail`, `agent-integration` (eager real `Sandbox.create()` from a unit test), `explore-plugin`, and `explore-byoc` tests is covered by the preload strip — verified empirically with the credential triple present (40 tests across those 4 files + 118 across sandbox-adjacent suites, 0 fail).
- **Comment accuracy** (comment-analyzer): `python.test.ts` comment now attributes `ATLAS_RUNTIME`/`VERCEL` to platform detection rather than `vercelSandboxAccess`; docs point recurrences at the preload strip list; CLAUDE.md pointer reworded.

Issue #4223, opened 2026-07-02
Investigate region pinning for Vercel Sandbox — semantic files leave the residency region


msywulak:
Neither explore nor Python passes a region when creating a Vercel Sandbox, and the SaaS deploy uses one Vercel team/project across all regions. Semantic files copied into the sandbox include sample values derived from customer data, so EU/APAC workspaces ship customer-derived data into a sandbox in an unspecified (likely US) region — at odds with the ADR-0024 regional residency guarantees.

**Investigate:**
1. Does `@vercel/sandbox` v2 `Sandbox.create()` support a region/placement primitive (param, project-level setting, or team-level pinning)? Check current SDK + Vercel Sandbox API docs — our local type stub (`packages/api/src/types/vercel-sandbox.d.ts`) mirrors only the surface we use, so absence there proves nothing.
2. If a primitive exists: plumb it through `sandbox.vercel` config (`config.ts` schema) and derive the value from region identity / `ATLAS_DEPLOY_ENV` env-profile rather than stamping per-service.
3. If none exists: decide between (a) documenting a residency carve-out for sandbox execution (DPA/docs), (b) making BYOC the residency answer for EU/APAC customers, (c) evaluating an alternative per-region sandbox provider behind the existing plugin seam.

## Key files
- `packages/api/src/lib/tools/explore-sandbox.ts` — `Sandbox.create({ runtime, networkPolicy, persistent })`, no region
- `packages/api/src/lib/tools/python-sandbox.ts` — same
- `packages/api/src/lib/tools/backends/detect.ts` — `vercelSandboxAccess()` (teamId/projectId/token only)
- `packages/api/src/lib/config.ts` — `sandbox.vercel` schema (`teamId` + `projectId`, no region)
- `deploy/api/atlas.config.ts` — SaaS pins `sandbox.priority: ["vercel-sandbox"]`, comment says team/project are "constant across regions"

## Acceptance criteria
- [ ] Determination documented: does a region primitive exist for Vercel Sandbox, yes/no, with source
- [ ] If yes: follow-up issue (or PR) plumbing region through config + env-profile, per-region verification plan
- [ ] If no: decision recorded (ADR-0024 amendment or docs carve-out) on how sandbox execution relates to the residency guarantee

## Dependencies
- ADR-0024 (regional residency routing)

msywulak:
Evidence from the live Vercel API (project sandbox listing, 2026-07-02):

[...]

PR #4224, opened 2026-07-02
fix(sandbox): evict stale explore backends on semantic writes; cap + timeout explore output

msywulak:
## Summary

Fixes three gaps found while comparing Atlas's sandbox layer against its genesis project (vercel-labs/oss-data-analyst). Sibling residency gap filed separately as #4223.

### 1. Stale semantic files in pooled file-snapshot explore backends (live on SaaS)
Semantic writes (`/admin/publish`, entity CRUD via `whitelist.ts`, knowledge via `admin-knowledge.ts`) call `invalidateOrgModeRoots`, which rebuilds the on-disk mode roots in place — but never dropped the org's cached explore backends. The Vercel/BYOC microVMs copy semantic files in **once at creation** and never re-read host disk, so a pooled sandbox served stale YAMLs after a publish until an infra error forced a rebuild. SaaS pins `vercel-sandbox`, so this was live.

`invalidateOrgModeRoots` now also fires `invalidateOrgExploreBackends(orgId)` (fire-and-forget dynamic import — breaks the `tools/explore` → `semantic/sync` module cycle; the eviction only has to land before the next agent turn).

### 2. Unbounded output + no wall-clock on exactly the SaaS-pinned backend
The vercel-sandbox explore backend returned `stdout()` untruncated and ran commands with no timeout; just-bash was also unbounded (nsjail/sidecar already cap 1 MB / 10 s). Now:
- **`capOutput`** (new, `backends/shared.ts`) caps stdout/stderr at 1 MB **at the tool seam** in `explore.ts`, with an explicit truncation notice for the model. Seam placement covers vercel, just-bash, sidecar, plugin, and BYOC backends — and all three explore surfaces (agent loop, MCP `explore`, `POST /api/v1/explore`) since they all route through `explore.execute`.
- The Vercel backend passes `timeoutMs: 10_000` to `runCommand` (sandbox-side SIGKILL), matching the nsjail/sidecar defaults. Added `timeoutMs` to our local `@vercel/sandbox` type stub (present in the installed SDK's own types).

### 3. Docs overstated explore enforcement
CLAUDE.md and `cli.mdx` claimed an `ls`/`cat`/`grep`/`find` allowlist with server-side rejection of writes and shell escapes. No such validator exists (`explore.ts` states this explicitly — enforcement is structural isolation per backend). Reworded both to describe the actual model: no command allowlist; read-only scoping is structural (ephemeral microVM / read-only binds / OverlayFs).

## Tests
- `explore-plugin.test.ts`: 3 new tests — oversized stdout truncated with notice, oversized stderr truncated on non-zero exit, small output passes through byte-identical.
- `mode-semantic-root.test.ts`: new test asserting `invalidateOrgModeRoots` evicts the org's cached explore backends (explore module fully mocked per mock-all-exports discipline).

## Verification
- `bun test` on both touched test files: green (12 + 9 pass).
- `bun run scripts/test-isolated.ts --affected`: 73/74 green; the one failure (`description-rubric.test.ts`) is the documented WSL2 load-flake and passes standalone.
- `bun run type`: green (after the type-stub addition).
- eslint on all touched files: clean.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,7 +75,7 @@ Guidance for Claude Code when working in this repository.
 
 ### Agent Tools
 - [ ] **Tools return structured data** — `executeSQL` returns `{ columns, rows }`
-- [ ] **Explore is read-only** — Only `ls`, `cat`, `grep`, `find` on `semantic/`. No writes, no shell escapes. Sandbox priority documented under **Security (General)** above
+- [ ] **Explore is read-only by isolation, not command validation** — There is no command allowlist; the agent may run arbitrary shell (`awk`/`sed`/pipes included). Read-only scoping to `semantic/` is enforced structurally by each backend (ephemeral microVM / read-only bind mounts / OverlayFs): writes land in ephemeral or in-memory layers and never touch host files. Output is capped at 1 MB at the tool seam. Sandbox priority documented under **Security (General)** above
 - [ ] **Agent max steps** — `stopWhen: stepCountIs(getAgentMaxSteps())`. Default 25, via `ATLAS_AGENT_MAX_STEPS` (1–100)
 - [ ] **Semantic layer drives the agent** — Read entity YAMLs before writing SQL
 
--- a/apps/docs/content/docs/reference/cli.mdx
+++ b/apps/docs/content/docs/reference/cli.mdx
@@ -302,7 +302,7 @@ bun run atlas -- metric run total_gmv --json
 
 ## explore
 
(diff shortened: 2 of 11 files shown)


msywulak:
Added in `c81b8392f`: attribution `tags` ({ `app`, `source`, `env` }) on every `Sandbox.create` — api explore (`source=explore`), api python (`source=python`), and the BYOC plugin (`explore-plugin` / `health-check`). Prompted by the 1,263 untagged sandboxes that accumulated from local dev/e2e runs carrying real `VERCEL_*` creds (all deleted; local `.env` now pins `ATLAS_SANDBOX_PRIORITY`). Type stub mirrors the SDK's `tags` param. Plugin tests 27 ✓, api sandbox tests 44 ✓, full `bun run type` ✓.

Issue #4734, opened 2026-07-19
Hosted MCP query tool drops transport on >120s runs — POST SSE stream idle, no progress heartbeat


msywulak:
The hosted MCP `query` tool (server-side analyst agent) drops the transport on any run longer than ~120s — the POST SSE stream goes idle for the whole agent run and an edge idle-timeout kills it.

## Symptom (observed live against hosted prod, Streamable-HTTP client)
- Calling `query` (NL question → second-LLM analyst → `{answer, sql, data}`) ran past 120s then failed: `MCP server "atlas" transport dropped mid-call; response for tool "query" was lost`.
- Fast deterministic tools (`executeSQL`, `listEntities`, `list_datasources`) return in ~1s and are unaffected — they simply finish under the edge idle window, not because they keep the stream warm.

## Root cause
The `query` POST response is a Streamable-HTTP `text/event-stream`, but **during the agent run the stream emits zero bytes**, so an intermediary idle-timeout (Railway edge/LB, ~120s) closes it → client sees `transport dropped`. The keepalive machinery exists but is never driven on this path:
- `packages/mcp/src/query-tool.ts:162-175` wraps the run in `withProgressAndCancellation` but discards the reporter — the work callback is `async (_reporter, _signal) => executeAgentQuery(...)`. Only `progress(0)` at start and `progress(1)` at end are emitted (`progress.ts:98`, `:114-117`); nothing between.
- `executeAgentQuery` (`packages/api/src/lib/agent-query.ts:165`) resolves only at completion and exposes no step/progress hook.
- The MCP SDK's `WebStandardStreamableHTTPServerTransport` sends no keepalive/ping on POST request streams; progress notifications *would* keep the stream warm (they route to the request stream via `relatedRequestId`) but none are emitted.

The drop is **not** enforced by our code — which rules our layer out: the API server sets `idleTimeout: 0` (`packages/api/src/api/server.ts:210`, comment cites 60-90s agent gaps) and the session sweep refreshes `lastSeenAt` per-chunk (`session-store.ts:401-411`). That leaves the intermediary; corroborated by `docs/guides/openstatus-mcp-monitor.md:34` (Railway per-request LB routing, 30s keepalive gaps).

Secondary: progress emission is a no-op unless the client sent a `progressToken` in `_meta` (`progress.ts:74,77`).

## Fix
[...]

Issue #4733, opened 2026-07-19
describeEntity (MCP) misses group-scoped entities that listEntities returns — disk-only lookup vs DB path


msywulak:
`describeEntity` (MCP) returns `unknown_entity` for group-scoped entities that `listEntities` surfaces — the two tools read different backends.

## Symptom (observed live against hosted prod)
- `listEntities` (no filter) returns 23 entities, every one `source: "g_prod"` (a connection group), e.g. `{name:"Conversation", table:"conversations", source:"g_prod"}`.
- `describeEntity({name:"Conversation"})` → `{code:"unknown_entity"}`. Same for the table name `{name:"conversations"}`.
- So an entity that lists fine cannot be described by its `name` **or** its `table` — the tools disagree.

## Root cause
The two MCP tools resolve entities through different backends, and only `listEntities` was migrated to the org-scoped DB path:
- `listEntities` (`packages/mcp/src/semantic-tools.ts:156`) calls the API `listEntities({ orgId: workspaceId, mode: "published" })`, which on SaaS hits the DB-backed branch (`packages/api/src/lib/semantic/entities.ts:686-696`) reading group-scoped rows from `semantic_entities` (hence every row's `source: g_prod` = `connection_group_id`).
- `describeEntity` (`packages/mcp/src/semantic-tools.ts:222` single, `:257` batch) calls `getEntityByName(name)` (`packages/api/src/lib/semantic/lookups.ts:37-62`), which is **disk-only**: no `orgId`, never checks `hasInternalDB()`, scans the on-disk semantic root. On a SaaS pod the group-scoped entities live only in `semantic_entities`, not on disk → every lookup returns `null` → `unknown_entity` (`semantic-tools.ts:229-235`).

This is the same `listEntities`/`executeSQL` divergence class that the DB-backed `listEntities` was built to close (#2142); `describeEntity` was never brought onto the DB path.

## Fix
[...]

PR #4737, opened 2026-07-19
fix(deploy): watch packages/mcp in api services so hosted-MCP fixes deploy

msywulak:
## What & why

The api service (all 3 regions: `api`, `api-eu`, `api-apac`) **serves the hosted MCP** — it depends on `@atlas/mcp` (`workspace:*`) and bundles `packages/mcp` into its image via the monorepo Docker build. But `deploy/api*/railway.json` `watchPatterns` **omitted `packages/mcp/**`**, so a change confined to `packages/mcp/**` produces `skippedReason: "No changes to watched files"` on the prod-branch push and **never rebuilds/deploys**.

Caught while shipping **v0.0.63** (#4733 + #4734, both `packages/mcp`-only): all 4 tag-gated services (`api`/`api-eu`/`api-apac`/`web`) `SKIPPED` the `prod` push, leaving v0.0.62's build (`227eb0db8`) live. The hosted-MCP fixes were merged, tagged, and released — but not running.

## The fix

Add `packages/mcp/**` to the `watchPatterns` of `api`, `api-eu`, `api-apac`. This commit is itself a `deploy/api*/**` change (already watched), so **releasing it rebuilds the api services and carries the pending MCP fixes (v0.0.63) to prod** in the same deploy.

## Why the `railway-watch` gate didn't catch it

`scripts/check-railway-watch.sh` verifies every explicit Dockerfile `COPY <src>` is watched — but `packages/mcp/package.json` is covered by `packages/**/package.json`, and the full mcp **source** arrives via the broad `COPY . .` the gate deliberately skips. So the gate can't see that a bundled workspace-dep's source must be watched.

## Follow-up (not in this PR)

Audit the api `watchPatterns` against its full workspace-dependency closure — `@atlas/okf-bundle`, `@useatlas/chat`, `@useatlas/twenty`, `@atlas/oauth-helper` may have the same latent gap — and consider extending `check-railway-watch.sh` to assert the closure. Filing as an issue.

## Scope
Deploy config only. No code change. Prod-affecting: the next release rebuilds the api services.

### DIFF
--- a/deploy/api-apac/railway.json
+++ b/deploy/api-apac/railway.json
@@ -9,6 +9,7 @@
       "package.json",
       "patches/**",
       "packages/api/**",
+      "packages/mcp/**",
       "packages/react/**",
       "packages/types/**",
       "packages/schemas/**",
--- a/deploy/api-eu/railway.json
+++ b/deploy/api-eu/railway.json
@@ -9,6 +9,7 @@
       "package.json",
       "patches/**",
       "packages/api/**",
+      "packages/mcp/**",
       "packages/react/**",
       "packages/types/**",
       "packages/schemas/**",
--- a/deploy/api/railway.json
+++ b/deploy/api/railway.json
@@ -9,6 +9,7 @@
       "package.json",
       "patches/**",
       "packages/api/**",
+      "packages/mcp/**",
       "packages/react/**",
       "packages/types/**",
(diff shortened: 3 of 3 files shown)


msywulak:
Follow-up filed: #4738 (audit api watchPatterns against full workspace-dependency closure + extend the gate).

Issue #4820, opened 2026-07-26
docs(security): write the incident-response runbook /dpa Annex II already claims exists


msywulak:
> *This was generated by AI during triage.*

Split out of #1928 Track 4 — the one item in that compliance program that costs nothing and is delegable today.

## Problem

`/dpa` Annex II asserts **"documented incident-response runbook"** as a present-tense Technical and Organizational Measure. No such document exists. `apps/docs/content/docs/security/` contains passkeys, row-level-security, security-headers, sql-validation, and two-factor-auth — nothing on incident response. The only `*incident*` match in the tree is a UI banner component.

This matters more than a typical docs gap because of where the claim sits. Every other unsupported item in that same Annex II sentence is deliberately hedged — "formal certification **on the roadmap**", "ISO 27001-**aligned** ISMS", "security program **aligned with** SOC 2 Type II controls", "A third-party penetration-testing program is **on the public roadmap**". The 2026-04-26 public-page audit (finding H9) softened four claims to exactly that shape. The incident-response runbook was not among the four and is still stated flatly, in a legal instrument, with no hedge.

So the options are to hedge it or to make it true. Making it true is cheap.

## The binding constraint the runbook must satisfy

`/dpa` §"Personal Data Breach" commits Atlas to notifying Customer **without undue delay and in any event within 48 hours** of becoming aware of a breach affecting Customer Data, with the information Art. 33(3) GDPR requires, plus ongoing updates as facts develop. That 48-hour clock is tighter than GDPR's own 72 and is the runbook's organizing deadline — every phase needs to fit inside it.

## Approach

Draft the runbook **grounded in mechanisms that already exist**, not in aspirational process. The detection and containment sections should describe what the system actually does today:

- `/api/health` component model — which components degrade vs 503, and the backups tripwire's cadence-window semantics
- SaaS boot guards as a fail-closed class, and what a guard failure looks like operationally
- Audit logging of administrative operations (what is captured, what is not)
- Regional isolation — three residency regions where the process is the region, and what that means for blast radius and for which regions a given incident touches
- Internal-DB scheduled backups + the verification path, as the recovery mechanism
- The status page and the in-app incident banner component as the customer-communication surfaces
- Credential rotation paths for the encrypted integration-credential store

[...]

PR #4466, opened 2026-07-10
fix(www): correct audit-found drift on pricing, legal, and landing copy

msywulak:
## Summary

`/www-audit` sweep (4 parallel agents, ~96 checks across Parts A–J). Zero CRITICAL findings; this PR fixes every repo-fixable finding:

**HIGH**
- **Pricing default-model drift** — `/pricing` said "Sonnet 4.6" for Pro/Business; `plans.ts` ships `anthropic/claude-sonnet-5`. Page now says "Sonnet 5" (card ×2 + comparison row).
- **Privacy region-selection understatement** — privacy policy said regions are Customer-selectable "on the Business plan"; the billing SSOT (`feature-entitlement.ts` residency floor = `trial`) grants it to every paid tier, and `/pricing` + docs already say so. Privacy now matches.

**MEDIUM**
- **Terms "write mode"** — ToS described a grantable write mode that doesn't exist (SQL validation is SELECT-only; sessions are read-only). Parenthetical removed; now consistent with `/security`.
- **Landing deploy card** — the "from $39" Cloud card listed Business-only features (SSO · SAML · SCIM, audit export) and Pro+ support; checklist now tier-accurate ("SSO · SCIM · audit on Business"). Also removed the hardcoded always-green "live status" dot (static decoration implying health) and softened the unverifiable "Weekly updates" to "Managed updates".
- **DPA subscribe card** — claimed "We email account admins automatically", but the sub-processor publisher is feed+webhook only. Copy now matches delivery (email list on request, feed, signed webhook). "RSS" button relabeled "Feed" (it serves Atom).
- **Sub-processor Annex I completeness** — Vercel's purpose only covered the AI Gateway; on SaaS, Vercel Sandbox also executes agent explore/Python workloads (`deploy/api/atlas.config.ts` pins `["vercel-sandbox"]`). Purpose broadened; `changed_at` bumped to 2026-07-10 so the change feed + webhooks announce the clarification.

**LOW**
- Comparison table: "stdio + SSE" → "stdio + Streamable HTTP" (canonical MCP transport description).
- Three quick-start links now point at `/self-hosted/getting-started/quick-start` directly instead of riding the Caddy 308.
- Legal stamps: `lastUpdated` set to 2026-07-10 for privacy/terms/dpa (also fixes terms `lastUpdated` predating its `effective` date). Versions intentionally not bumped — these are errata, not policy changes; bump if you consider any of them a formal revision.

[...]

Issue #1928, opened 2026-04-27
Compliance program: SOC 2 Type II + ISO 27001 + annual third-party pen test + IR drills


msywulak:
## Context

Surfaced by the 2026-04-26 public-page audit (`.claude/research/public-page-audit-2026-04-26/report.md`, finding **H9**).

The legal pages originally asserted four claims that have no evidence in the codebase or operational history:

1. **SOC 2 Type II** "reported annually" / "current SOC 2 Type II report"
2. **ISO 27001** certificate (vs. ISMS-aligned)
3. **Annual third-party penetration testing**
4. **Quarterly incident-response drills**

The same audit pass softened the prose on `/dpa` Annex II and `/privacy` §9 to "aligned with SOC 2 Type II controls / ISO 27001-aligned ISMS / on the roadmap." That's defensible for now — but procurement-grade enterprise sales will require at least two of the four to actually exist. This issue tracks the work to make each claim true so the prose can be strengthened back.

### Track 1 — SOC 2 Type II
- [ ] Pick an auditor (Drata / Vanta / Secureframe → audit firm; common picks: Insight Assurance, Prescient Assurance, AssuranceLab)
- [ ] Choose Trust Services Criteria scope (at minimum Security; consider Availability + Confidentiality)
- [ ] 6-month observation window (Type II requires controls *operating effectively over time*; a Type I report is faster but enterprise procurement usually wants Type II)
- [ ] Once issued: enable on-request distribution under NDA; revert /dpa §7 + /privacy §9 + /dpa Annex II prose to the stronger claim

### Track 2 — ISO 27001 certification
- [ ] Build the ISMS scope statement
- [ ] Stage 1 + Stage 2 audits (typically 12-18 months from kickoff for a startup)
- [ ] Optional — ISO 27017 / 27018 add-ons for cloud + privacy
- [ ] Once issued: revert /dpa Annex II prose to "ISO 27001-certified ISMS"

### Track 3 — Annual third-party penetration testing
- [ ] Pick a vendor (Cobalt, NCC Group, Bishop Fox, Trail of Bits — choice depends on budget + targeted depth)
- [ ] Scope: web app (app.useatlas.dev), API (api.useatlas.dev + regional endpoints), agent loop SQL-injection surfaces, plugin sandbox isolation
- [ ] Annual cadence; remediation tracked in this repo as `area: security` issues
[...]

Issue #1922, opened 2026-04-27
Pre-signed DPA-v2.4-pre-signed.pdf — host static under apps/www/public/dpa/


msywulak:
**Context**

PR #1921 (1.3.0 Bucket 8 /dpa redesign) shipped a \"Request countersigned PDF\" CTA on /dpa that opens \`mailto:legal@useatlas.dev\` with a pre-filled subject. The intended end-state is a static, pre-signed PDF hosted directly on the marketing site so procurement teams can self-serve.

**Acceptance**

- [ ] Legal counsel reviews + countersigns the DPA-v2.4 contents currently on /dpa.
- [ ] Drop the resulting PDF at \`apps/www/public/dpa/DPA-v2.4-pre-signed.pdf\`.
- [ ] Update \`apps/www/src/app/dpa/page.tsx\` pre-signed card: change CTA from \`mailto:\` to direct \`<a href=\"/dpa/DPA-v2.4-pre-signed.pdf\" download>Download PDF</a>\`.
- [ ] Verify the PDF metadata (\"countersigned by Atlas\", \"valid through 2027-01-15\", page count) matches the card copy on /dpa.
- [ ] If counterparty edits land in the PDF that differ from the on-page text, sync the on-page text first so the rendered HTML and the downloadable PDF say the same things.

**Scope notes**

- Don't ship a draft PDF as \"pre-signed\" — wait for actual sign-off.
- The expiry date in the on-page meta line (currently 2027-01-15) tracks the SCC reference date; refresh together if either changes.

msywulak:
Heads-up from the `/www-audit` pass (#2896): `apps/www/src/app/dpa/page.tsx` has since moved to **v2.5** (effective 2026-05-02) — it now names `DPA-v2.5-pre-signed.pdf` (`:191`), not the v2.4 filename in this issue's title. No file exists under `apps/www/public/dpa/` yet, and the "Request countersigned PDF" CTA is a `mailto:legal@useatlas.dev` (request-by-email) rather than a hard link, so there's **no broken link** today — this issue is still the right tracker for actually hosting the static signed PDF. Suggest updating the title/scope to `DPA-v2.5-pre-signed.pdf` when picked up.

msywulak:
[...]

PR #4821, opened 2026-07-26
fix(www): remove unsupported security claims from /dpa + /privacy

msywulak:
## What

Audited every technical and organizational measure asserted on `/dpa` Annex II and `/privacy` §9 against the codebase, and corrected the ones that were never true.

## The four overclaims

| Claim | Evidence it was false |
|---|---|
| `secure SDLC with mandatory code review` | `gh api .../branches/main/protection` → `required_pull_request_reviews: null`, `enforce_admins: false`. CLAUDE.md: *"Required reviews are intentionally off — solo dev + parallel-claude workflow."* |
| `automated vulnerability scanning of container images and dependencies` | Dependencies ✅ — Dependabot security updates enabled, CodeQL default setup configured. **Container images ✗** — no Trivy/Grype/Snyk in any workflow. |
| `ISO 27001-aligned ISMS` | No ISMS exists. #1928 Track 2 still lists *"Build the ISMS scope statement"* as an open to-do. |
| `documented incident-response runbook` | No such file. `apps/docs/content/docs/security/` has five pages, none of them incident response. Tracked as #4820. |

**The pattern worth noting:** the 2026-04-26 public-page audit (finding H9) softened four claims to *"on the roadmap"* / *"aligned with"*. Every claim it touched stayed honest; every claim it missed stayed flatly asserted. In `/dpa` Annex II these four sat unhedged in a list of otherwise-hedged neighbours — the unhedged item in a hedged list is the tell.

Rather than delete them, each moves into an explicit **"on the public roadmap and are not represented as in place today"** clause, so a procurement reviewer sees the posture instead of a silent gap.

## Also removed

`least-privilege IAM` and `segregated production access` — operational claims about the cloud setup that can't be verified from the repo. Maintainer confirmed removal.

## Added — real controls that were previously unclaimed

Static application security testing on every change · secret scanning with push protection · fail-closed startup guards · audit-log retention with data-subject erasure tooling (`admin-action-retention.ts` implements GDPR/CCPA right-to-erasure with a `dsr_request` origination path and blast-radius preview).

2FA corrected: **TOTP or passkey**, scoped to the roles the gate actually enforces (`admin`, `owner`, `platform_admin`) on managed-mode sessions.

## The most substantive change isn't a removal

Annex II now discloses the **sandbox execution carve-out** — US `iad1` tool execution for all regions, because the sandbox sub-processor exposes no region primitive. #4223 required this on *every* residency-promise surface. The DPA is one, and carried it nowhere. That gap predates this PR.

Read-only is also now stated accurately: SELECT-only validation applies to every generated query, but **session-level** read-only enforcement is PostgreSQL, MySQL, and ClickHouse only — the first draft of this PR overstated it as universal.

## Pre-signed PDF

`/dpa` asserted *"DPA-v2.5-pre-signed.pdf · Countersigned by Atlas · valid through 2027-05-02 · available on request"*. **No such PDF exists** — not in the repo, not offline (confirmed with the maintainer). The hero's *"Pre-signed for standard deals"* said the same thing.

The card now says the page text is the agreement and a countersigned PDF is produced on request, which is what actually happens. #1922 still tracks hosting a static signed copy if that's wanted later.

## Stamps

`lastUpdated` → 2026-07-26 on both pages; versions held at v2.5 / v3.2, matching the #4466 errata precedent. **No countersigned DPA has ever been issued, so no counterparty has relied on the prior Annex II text** — this corrects a public misstatement rather than amending an executed agreement.

## Review notes

- ⚠️ **This is a legal instrument — counsel should review before merge.** This PR verified claims against *code*; it does not assess them against GDPR Art. 28/32.
- `www` deploys direct-from-main, so merging publishes immediately. Launch week is Jul 27–31 — the maintainer has asked to merge after.
- `tsgo --noEmit` clean. Lint warnings in the run are pre-existing `packages/api` test warnings (the permanent `warn` set per ADR-0031), untouched by this change.

Refs #1928, #4820, #1922, #4223


msywulak:
## Second pass — the remaining public surfaces

Audited `/terms`, `/aup`, `/sub-processors`, `/security`, and the `.well-known` files. Two more fixes pushed; most of it came back clean.

### `/terms` — fixed

| Claim | Reality |
|---|---|
| `configurable PII detection on result sets`, listed among measures Atlas implements generally | Masking is entitlement-gated: `FEATURE_ENTITLEMENTS.masking = "business"`. A Starter or Pro customer does not get it. Now scoped to **(Business plan)** and renamed to *PII masking*, which is what the code calls it. |
| `least-privilege access` | Same unverifiable operational class removed from `/dpa` and `/privacy` in the first pass. Dropped for consistency. |

### `/terms` — verified accurate, unchanged

TLS 1.2+/AES-256 · logged admin operations · **data residency on all paid plans** (`plans.ts` records the entitlement floor as `trial`, i.e. every active paid tier — the claim is exactly right) · 14-day trial (`TRIAL_DAYS`) · business-email-only signup (`lib/auth/business-email.ts`).

The **7-layer validation enumeration** is precisely correct. `/terms` lists *empty check, regex guard, AST parse, table whitelist, RLS injection, auto-LIMIT, statement timeout*; the canonical pipeline is `0. Empty check → 1. Regex mutation guard → 2. AST shape guards → 3. Table whitelist`, then at execution `RLS injection → Auto LIMIT → Statement timeout`. Seven, in order, no drift.

### Clean — no changes

- **`/aup`** — no unbacked factual claims. Conspicuously honest in the right places: *"Atlas does not pre-screen Customer Data"*, *"Atlas does not currently operate a paid bug-bounty program."* The 5-business-day research response and 30-day change notice are forward commitments, not assertions of present fact.
- **`/security`** — "seven-layer" matches; the network-denied microVM claim matches the SaaS `deny-all` sandbox pin; read-only is already correctly scoped to PostgreSQL/MySQL/ClickHouse at the session level.
- **`/sub-processors`** — all nine entries map to real usage. Bedrock, Ollama, and openai-compatible are correctly **absent**: they're self-host provider options, and the DPA governs Atlas Cloud only, where `ATLAS_PROVIDER` is immutable.
- **`.well-known`** — `security.txt` present and unexpired (2027-04-26), with Canonical/Policy/Encryption set. The PGP fingerprint cited on `/privacy` (`B00E 2A64 …92C7`) **matches the actual key byte for byte**, and the key is valid to 2028-05-01.

### One open question for counsel — not a factual error

`/terms` and `/dpa` state **different deletion timelines**:

- **`/terms` §Term & Termination** — data available for export 30 days → deleted from production within 30 days → from backups within 90 days
- **`/dpa` §Return & Deletion** — return or delete within 90 days → backups within an *additional* 90 days

Both are satisfiable (actual `ATLAS_BACKUP_RETENTION_DAYS` default is 30), and `/dpa` §1 says the DPA controls on conflict — so this isn't a breach risk. But procurement reviewers read these side by side, and the difference invites a question. Worth picking one set of numbers deliberately rather than leaving it to the conflict clause.

Also worth a deliberate decision: **Slack is not in the sub-processor list.** Defensible — the customer enables the integration into their own workspace, so Slack is arguably their processor, not Atlas's. But it's a judgment call that should be made on purpose, not by omission.

msywulak:
## Third pass — Slack added, deletion timelines unified on the DPA

### Slack → Annex I

The Atlas Slack app posts answers and query results into the Customer's workspace and receives questions from it, with per-workspace bot tokens stored encrypted. That is sub-processing, and it was disclosed nowhere.

**Region** is stated as *"Customer's own Slack workspace — data location determined by Customer's agreement with Slack"* rather than "United States". Slack's data location follows the Customer's own Slack plan (Enterprise Grid offers EU residency), so asserting a region Atlas doesn't control would be a new overclaim of exactly the kind this PR exists to remove.

**`changed_at` is today**, so the Atom feed's `<updated>` advances and anyone subscribed to sub-processor notices is alerted. Worth being precise about what this is: a **disclosure of an existing relationship, not a new engagement** — it does not start the DPA's 30-day objection clock against a newly-added vendor.

⚠️ **One field I inferred: `since: "2026-01"`.** The repo's git history was squash-seeded on 2026-04-25, so the true date Slack became live on Atlas Cloud isn't recoverable from it. I matched the founding-vendor cohort because the Slack adapter predates the seed. **If you know the actual month, correct it** — a `since` date in a legal annex is a factual claim, and guessing at one is the specific failure mode this whole PR is about.

### `/privacy` reconciled

`/privacy` §"Who we share with" enumerates sub-processors *by category* — cloud infrastructure, uptime monitoring, payments, bot mitigation, CRM. Chat delivery wasn't among them, so adding Slack to Annex I alone would have left the two documents disagreeing. Category added in both the legal text and the plain-English summary.

### Deletion timelines unified

| Document | Before | Now |
|---|---|---|
| `/dpa` | return or delete within 90 days; backups an additional 90 | unchanged — this is the canon |
| `/terms` | production deletion within 30 days; backups within 90 | **matches the DPA**, and says so |
| `/privacy` | "encrypted backups: up to 90 days" (rolling-window framing) | **matches the DPA's additional-90 framing** |

The 30-day export window in `/terms` is retained — the DPA doesn't mention it, and it's a commitment in the Customer's favour.

All three now state the same numbers, so a procurement reviewer diffing them finds agreement instead of a conflict clause.

`tsgo` clean; sub-processor JSON parses to 10 entries.

msywulak:

**Slack `since` corrected to `2026-03`** — maintainer-confirmed. The inferred `2026-01` is gone; that was the one guessed field flagged in the previous comment, so Annex I now contains no inferred values.

The entry also moved after OpenStatus to keep the list chronological by `since` (2026-03 sits between the 2026-01 founding cohort and the 2026-06 additions). `changed_at` stays `2026-07-26`, so the Atom feed still advances and the disclosure still notifies.

Issue #4457, opened 2026-07-10
prod: automated backup scheduler is dead code — startScheduler has no production caller (and no volume for ./backups)


msywulak:
Found by `/prod-audit` (2026-07-10), Part D.

## Problem

The automated backup cron never runs. `startScheduler` in `ee/src/backups/scheduler.ts:133` has zero production callers — repo-wide grep finds only its definition and the re-export at `ee/src/backups/index.ts:48`. It is not in `BackupsManagerShape`, not wired in `packages/api/src/lib/effect/enterprise-layer.ts`, and not among the ~25 `registerPeriodicFiber` names in `packages/api/src/lib/effect/layers.ts`.

The docs promise "Scheduler: cron-based, default daily at 03:00 UTC" (`apps/docs/content/docs/platform-ops/backups.mdx`). The restore-and-verify pipeline itself is solid (`ee/src/backups/verify.ts` does a full restore into a scratch DB with a source-table-count assertion, and `scheduler.ts:102-109` verifies every automated backup) — but it is unreachable via automation. Only manual API-triggered backups can occur.

Compounding: backups write to local `./backups` and `deploy/api/railway.json` declares no volume, so even manual backups do not survive a redeploy.

Until fixed, the only backstop is Railway's managed-Postgres backups on the three regional internal DBs — the operator should verify those are enabled.

## Key files
- `ee/src/backups/scheduler.ts:133` — `startScheduler`, never called in production
- `packages/api/src/lib/effect/enterprise-layer.ts` — where the wiring belongs (or a `registerPeriodicFiber` in `layers.ts`)
- `deploy/api/railway.json` — no volume for the backup output dir
- `apps/docs/content/docs/platform-ops/backups.mdx` — the promise ("default daily at 03:00 UTC")

## Acceptance criteria
- [ ] The backup scheduler actually starts on SaaS boot (via the enterprise layer / a registered periodic fiber, with the per-tick span the fiber registry requires)
- [ ] Backup output survives redeploy (Railway volume or object storage) — or the docs are corrected to match reality
- [ ] `ATLAS_BACKUP_VERIFY_SCRATCH_URL` presence on prod services confirmed so verification runs in full-restore mode, not header-only
- [ ] A boot-green-but-broken tripwire: health/metrics surface when the last successful verified backup is older than the cadence

## Dependencies
- None (the backup + verify code all exists; this is trigger wiring + deploy config)

msywulak:
[...]

Issue #3958, opened 2026-06-25
refactor(deploy): retire or finish the separate api-staging config — commit to shared-config soak fidelity


msywulak:
## Context

The `api-staging` soak service builds from the **prod** image/config (`RAILWAY_DOCKERFILE_PATH=deploy/api/Dockerfile` → `COPY deploy/api/atlas.config.ts`) and differs from prod only by env vars (`ATLAS_API_REGION=staging`, `ATLAS_DEPLOY_ENV=staging`, staging DB URLs/keys). A half-built alternative — `deploy/api-staging/atlas.config.ts` — exists but is **not wired into any image** and is **~4 commits stale** (missing operator-credentials resolver #3737, chat-plugin resume delivery #3832, datasource-guard #3921).

This drift directly caused an outage: PR #3951 deleted the load-bearing `staging` residency arm from the prod config believing the separate config covered staging boot. It didn't → api-staging crash-looped. Fixed in #3957 by keeping `staging` in the shared config flagged `selectable: false`.

## Decision needed

Pick one and execute:

- **(A) Retire the separate config (recommended).** Delete `deploy/api-staging/atlas.config.ts`, commit to the shared-config-+-env-vars model (maximizes soak fidelity — staging runs the byte-identical prod config). Update references in PRD `docs/prd/staging-environment.md`, ADR-0014, `docs/development/staging-environment.md`, and the `deploy-config-residency-regions.test.ts` staging-config assertion. Document the model in the staging README/PRD.
- **(B) Finish the separation.** Sync the staging config up to current prod, add `deploy/api-staging/Dockerfile` (COPY the staging config), repoint the api-staging Railway service, and add a CI drift-check that keeps the two configs in sync. More moving parts + a standing maintenance tax.

## Key files
- `deploy/api/atlas.config.ts` (shared prod+staging config; `staging` arm is load-bearing)
- `deploy/api-staging/atlas.config.ts` (unused, stale)
- `deploy/api/Dockerfile`
- `packages/api/src/lib/__tests__/deploy-config-residency-regions.test.ts`
- `docs/prd/staging-environment.md`, `docs/adr/0014-salesforce-datasource-stays-on-oauth.md`, `docs/development/staging-environment.md`

## Acceptance criteria
- One model chosen and the dead/duplicate config removed or kept-in-sync (no silent drift possible).
- If (A): no remaining references imply a separate staging config is loaded at runtime.
[...]

Issue #1981, opened 2026-05-02
prod: SaaS /health must return 503 when internal DB is down (currently 200 degraded)


msywulak:
**Surfaced by:** /prod-audit pass on 2026-05-02 (Part B — Graceful Degradation).

## Problem

\`packages/api/src/api/routes/health.ts:225-234\`. Status logic:

\`\`\`
hasDsError && !dsNotConfigured → "error" (503)
otherwise internal-DB failure → "degraded" (200)
\`\`\`

For SaaS where internal DB is the source of truth for auth, org, billing, settings, audit, scheduled tasks — a downed internal DB should fail the load balancer probe so traffic stops routing to the affected pod. Today the LB sees 200 (degraded) and keeps sending requests that will all fail at the auth layer.

## Acceptance

- [ ] When \`getConfig().deployMode === "saas"\` AND \`hasInternalDbError === true\` → status \`"error"\`, return 503
- [ ] Self-hosted deployments retain current behavior (degraded → 200) — internal DB is optional in self-hosted (\`hasInternalDB()\` returns false acceptably)
- [ ] Test covers SaaS mode + internal DB unreachable → 503

## Related

- Auth fail-closed semantics already correct (`auth/middleware.ts:351-362, 433-449` — verified clean by Part B)
- This is purely about the LB probe response so traffic routing reflects reality

Issue #4124, opened 2026-06-29
atlas query (NL agent) is unusable against a workspace datasource — CLI sends API key as Bearer + /api/v1/query ignores connectionId


msywulak:
Found while functionally testing the CLI/MCP actions against a real datasource (staging, 2026-06-29). `atlas sql` and MCP `executeSQL` returned real rows, but `atlas query` (the flagship NL text-to-SQL action) is broken **two** independent ways:

### Bug A — CLI auth (`packages/cli/src/commands/query.ts`)
`query.ts` reads **only** `process.env.ATLAS_API_KEY` and sends it as `Authorization: Bearer <key>` (lines 54, 138, 144). It **never** reads the `atlas login` device-flow session (`readSession`), unlike `sql`/`entities`/`datasource`/`explore`. And API keys must ride the `x-api-key` header (the Better Auth `apiKey()` plugin), **not** `Authorization: Bearer` — so the server tries to validate the key as a session JWT and returns **"Not signed in"**.

Net: `atlas query` fails for a logged-in user (session ignored) **and** for an API-key user (wrong header). Observed: `Error: Authentication failed -- Not signed in. Set ATLAS_API_KEY...` even with `ATLAS_API_KEY` exported and called via the bin directly.

### Bug B — server connection routing (`/api/v1/query`)
Hitting `/api/v1/query` directly with valid auth (session Bearer **and** `x-api-key` both clear auth) and `{"question":"...","connectionId":"ds_actions"}` returns:
```
{"error":"configuration_error","message":"DATABASE_URL is set but ATLAS_DATASOURCE_URL is not. ...",
 "diagnostics":[{"code":"MISSING_DATASOURCE_URL", ...}]}
```
So the NL `/query` path **ignores `connectionId`** and falls back to the **env-level** `ATLAS_DATASOURCE_URL` instead of resolving the workspace's registered connection — unlike `executeSQL`/`/api/v1/execute-sql`, which routes to the workspace ConnectionRegistry correctly (proven: same query returned real rows via `executeSQL`). On a multi-tenant SaaS workspace there is no env `ATLAS_DATASOURCE_URL`, so the NL agent can never run.

## Key files
- `packages/cli/src/commands/query.ts` (auth: add `readSession` fallback; send API keys via `x-api-key`, mirror `lib/credential.ts` like `sql`/`datasource` do)
[...]

PR #4855, opened 2026-07-27
fix(api): let a deployment declare it expects no datasource

### LINKED ISSUE #4854: fix(api): /health degrades forever on a deployment that intentionally has no datasource
`GET /api/health` on staging reports `status: "degraded"` permanently, because staging is a multi-tenant SaaS deploy with no process-level analytics datasource. That burns the signal: the top-level status can no longer be used to detect a real regression on the soak environment, which is exactly what staging exists for.

Verified live (2026-07-27):

```
$ curl -s https://api.staging.useatlas.dev/api/health
{"status":"degraded","region":"staging",
 "components":{"datasource":{"status":"disabled","message":"MISSING_DATASOURCE_URL"},
               "internalDb":{"status":"healthy"},"provider":{"status":"healthy"},
               "scheduler":{"status":"healthy"},"sandbox":{"status":"healthy"},
               "plugins":{"status":"healthy"},"backups":{"status":"healthy"}}, ...}   # HTTP 200

$ curl -s https://api.useatlas.dev/api/health
{"status":"ok","region":"us", ...}   # prod sets ATLAS_DATASOURCE_URL, so it is unaffected
```

Every component is healthy; the only thing dragging the rollup down is the deliberate absence of a datasource. Note the contradiction inside a single response: `components.datasource.status` is `"disabled"` — the schema's state for *intentionally absent* — while the rollup treats the same condition as degradation.

## Mechanism

`packages/api/src/api/routes/health.ts:372-382`:

```ts
if ((hasDsError && !dsNotConfigured) || internalDbBlocksProbe) status = "error";
else if (dsNotConfigured || hasKeyError || hasSemanticError || hasInternalDbError || hasAuthError)
  status = "degraded";
else status = "ok";
```

The `&& !dsNotConfigured` guard already stops `MISSING_DATASOURCE_URL` from becoming a 503 (correct — staging returns 200 and stays in the LB, #1981). But `dsNotConfigured` is still a term in the degraded arm, and there is no input anywhere in the route that answers *is a datasource expected on this deployment at all?*

That input is the actual gap. The `#4457` scheduled-backup tripwire a few lines down (`health.ts:581-599`, `lib/backups/health.ts`) is the precedent to copy: `getScheduledBackupHealth()` returns `{ expected: false }` when scheduled backups aren't expected here, and only a deployment that *does* expect them can go `degraded`. The datasource check has no `expected` equivalent.

## The trap

A self-hosted box that **should** have a datasource but is missing `ATLAS_DATASOURCE_URL` must **still** degrade. Intent must not be inferred from the absence itself — that converts a misconfiguration into a silent green, which is the "prefer errors over silent fallbacks" rule in CLAUDE.md. The expectation has to be **declared explicitly**, and the default when undeclared must be today's degrading behaviour.

## Why the declaration cannot live in `atlas.config.ts` alone

`api-staging` builds from `deploy/api/Dockerfile`, which COPYs the **prod** `deploy/api/atlas.config.ts` — the separate staging config was retired in #3958/#4368. Staging and prod differ only by env vars. A single config-file field would therefore declare "no datasource expected" for **prod too**, and prod *does* have one — suppressing exactly the real misconfiguration signal we want to keep. So the per-service seam (an env var) is load-bearing; the config field is a convenience for self-hosters authoring their own `atlas.config.ts`.

## Key files
- `packages/api/src/api/routes/health.ts:335-393` — `dsNotConfigured`, the three-arm rollup, the `exploreFailClosed`/plugins/backups promotion pattern
- `packages/api/src/api/routes/health.ts:616-628` — `components.datasource`, which already reports `"disabled"` but stamps `message: "MISSING_DATASOURCE_URL"` on it
- `packages/api/src/lib/backups/health.ts` — the `{ expected: false }` precedent (#4457) and `scheduledBackupsExpected()`
- `packages/api/src/lib/backups/cadence.ts:47` — `isScheduledBackupEnvDisabled()`, the env-declaration naming precedent (`ATLAS_BACKUP_SCHEDULED_ENABLED=false`)
- `packages/api/src/lib/startup.ts:135-165` — `checkDatasourceUrlPresence`; emits `MISSING_DATASOURCE_URL` as an **error** (not a warning) when `DATABASE_URL` is set and `ATLAS_DATASOURCE_URL` is not, which is staging's exact shape
- `packages/api/src/lib/config.ts:475-481` — `AtlasConfigSchema`, where a `datasourceExpected` field would go (+ the resolve passthrough at ~1322)
- `packages/api/src/lib/workspace-capability.ts` — prior art for the same underlying truth: #4826 established that a multi-tenant / knowledge-only deployment legitimately has no process datasource
- `deploy/api/atlas.config.ts` — SaaS config registers datasource plugins adapter-only, "no operator env var, no static datasource"
- `packages/api/src/api/__tests__/health.test.ts` — existing mocks for `validateEnvironment` + `ConnectionRegistry`

## Findings
- Staging is `degraded` with every component healthy; prod `us` is `ok` only because `ATLAS_DATASOURCE_URL` happens to be set there. The multi-tenant SaaS shape is otherwise identical, so any region that drops its demo datasource inherits the same permanent degradation.
- The response is internally inconsistent: the component says `disabled` (intentionally absent), the rollup says `degraded` (something is wrong).
- `components.datasource.message` is `"MISSING_DATASOURCE_URL"` even on the `disabled` status, because `hasDsError` is true in staging's shape — the operator dashboard shows an error code for an intentional state.
- The 503/LB-eviction path is orthogonal and must stay untouched: it can only be reached when a datasource **is** configured (`hasDsError && !dsNotConfigured`), so a declaration consulted solely in the `dsNotConfigured` branch cannot weaken it.

## Suggested fix

1. Add an explicit expectation input, e.g. `isDatasourceExpected()` in `packages/api/src/lib/db/` (or a small `lib/health/` module), resolved in this precedence:
   - `ATLAS_DATASOURCE_EXPECTED=false` (or `0`) → not expected (mirrors `isScheduledBackupEnvDisabled()`)
   - else `getConfig()?.datasourceExpected === false` → not expected
   - else **expected** (undeclared default = today's behaviour)
2. In `health.ts`, use it **only** inside the no-datasource branch: `const dsIntentionallyAbsent = dsNotConfigured && !isDatasourceExpected();` then drop `dsNotConfigured` from the degraded arm in favour of `(dsNotConfigured && !dsIntentionallyAbsent)`. Leave the `error` arm byte-for-byte unchanged.
3. When intentionally absent, replace the `MISSING_DATASOURCE_URL` message on `components.datasource` with a plain statement of intent (status stays `"disabled"`; no wire-schema change, so no OpenAPI drift).
4. Add `datasourceExpected: z.boolean().optional()` to `AtlasConfigSchema` + the resolved-config passthrough, and document the env var in `.env.example` and the deployment docs.
5. Set `ATLAS_DATASOURCE_EXPECTED=false` on the **`api-staging` service only** (a service-scoped Railway override, not a shared-scope var — shared-scope refs resolve empty here), so staging's health becomes a usable signal.

## Acceptance criteria
- [ ] A deployment that explicitly declares no datasource is expected reports `components.datasource.status = "disabled"` and does **not** degrade the top-level rollup
- [ ] A deployment that does not declare it, with no datasource, still reports `degraded` — unchanged from today
- [ ] A deployment with a datasource configured that then FAILS its `SELECT 1` still reports `status: "error"` + HTTP 503 — the #1981 LB-eviction path is untouched
- [ ] The declaration cannot green a real failure: declared-not-expected **plus** a configured datasource that fails still reports `error`/503
- [ ] `components.datasource` no longer carries `message: "MISSING_DATASOURCE_URL"` when the absence is declared
- [ ] Tests in `packages/api/src/api/__tests__/health.test.ts` cover all four cases above, explicitly including the **negative** (undeclared + absent still degrades) — a fix that greens everything would pass a naive test
- [ ] `ATLAS_DATASOURCE_EXPECTED=false` is set on the `api-staging` Railway service and `https://api.staging.useatlas.dev/api/health` returns `status: "ok"`
- [ ] The env var + config field are documented (`.env.example`, deployment docs)

msywulak:
Closes #4854

`GET /api/health` on staging has been `degraded` permanently — every component healthy, the rollup red — because staging is a multi-tenant deploy with no process-level analytics datasource. That burns the top-level status as a regression signal on the one environment that soaks `main`.

Verified live before the fix:

```
$ curl -s https://api.staging.useatlas.dev/api/health
{"status":"degraded","region":"staging",
 "components":{"datasource":{"status":"disabled","message":"MISSING_DATASOURCE_URL"},
               "internalDb":{"status":"healthy"},"provider":{"status":"healthy"},
               "scheduler":{"status":"healthy"},"sandbox":{"status":"healthy"},
               "plugins":{"status":"healthy"},"backups":{"status":"healthy"}}}   # HTTP 200
```

Note the response contradicting itself: the component says `disabled` (the schema's state for *intentionally absent*), the rollup says `degraded`.

## What was missing

The rollup had no input answering **"is a datasource expected on this deployment at all?"**, so it could not tell an intentional absence from an operator who forgot `ATLAS_DATASOURCE_URL`. This adds one, following the `#4457` scheduled-backup tripwire's `{ expected: false }` precedent a few lines down in the same file.

- `ATLAS_DATASOURCE_EXPECTED=false` (per service), or `datasourceExpected: false` in `atlas.config.ts`
- New `lib/db/datasource-expectation.ts` owns the resolution and documents why

## What it deliberately does not do

- **Intent is never inferred from the absence itself.** Undeclared still degrades — that is what keeps a self-hosted box that meant to set `ATLAS_DATASOURCE_URL` visible instead of silently green.
- **It cannot mute a real failure.** The declaration is narrowed by `dsNotConfigured`, so it is structurally unreachable on a deployment that *has* a datasource. A configured datasource that fails still reports `error` + 503 and still leaves the load balancer (#1981).
- An unrecognized env value falls back to *expected* (the noisy answer) and warns once.

The env var overrides the config field **in both directions**, because `api-staging` builds from the prod `deploy/api/atlas.config.ts` (#3958) and differs only by env vars — a shared config file must not be able to declare an expectation on another service's behalf.

Also drops the `MISSING_DATASOURCE_URL` message from the `disabled` component when the absence is declared. An undeclared absence keeps it: there, it is the finding.

## Tests

`packages/api/src/api/__tests__/health.test.ts` (+8) and a new `lib/db/__tests__/datasource-expectation.test.ts` (+8). The positives are the easy half; these are the ones that matter:

- **undeclared + no datasource still degrades** — a fix that simply stopped counting a missing datasource would pass every positive test and silently green a misconfigured self-host
- **declared not-expected + a configured datasource that FAILS still returns `error` + 503** — the declaration must not become a mute switch
- an unrecognized declaration is treated as undeclared, so a typo leaves the deployment noisy

## Rollout

After merge, set `ATLAS_DATASOURCE_EXPECTED=false` on the **`api-staging` service only** (service-scoped Railway override, not shared scope), then confirm `https://api.staging.useatlas.dev/api/health` returns `"ok"`. Prod `api`/`api-eu`/`api-apac` must **not** get it — they have real datasources and must keep degrading if they ever lose one.

## Verification

- `/ci` — 32/32 gates PASS
- `--affected` — 119 files, all pass
- No wire-schema change (`"disabled"` was already in the enum), so no OpenAPI drift

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -73,6 +73,15 @@ ATLAS_DATASOURCE_URL=postgresql://atlas:atlas@localhost:5432/atlas_demo
 # ATLAS_ES_AWS_REGION=                         # Selects AWS SigV4 (Amazon OpenSearch Service); keys from AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY (+ AWS_SESSION_TOKEN)
 # ATLAS_ES_AWS_SERVICE=                        # SigV4 service code (default "es"; "aoss" for OpenSearch Serverless)
 # ATLAS_ES_ENGINE=                             # Optional engine override: elasticsearch | opensearch (wins over the URL scheme)
+# ATLAS_DATASOURCE_EXPECTED=false              # Declare that this deployment intentionally has NO process-level analytics
+#                                              # datasource — a multi-tenant deploy whose connections live per workspace, or a
+#                                              # knowledge-only / brain-only box. /health then reports the datasource component
+#                                              # as "disabled" without degrading the top-level status. Leave it UNSET on a
+#                                              # deployment that should have one: undeclared means a missing ATLAS_DATASOURCE_URL
+#                                              # still degrades /health, which is how a misconfiguration stays visible. Set it
+#                                              # per service — it overrides `datasourceExpected` in atlas.config.ts, which several
+#                                              # services may share. It never silences a datasource that IS configured and failing
+#                                              # (that is still error + 503).
 # NOTE: ATLAS_ES_API_KEY is the canonical name. The atlas.config.ts examples wire the same var into the
 # plugin's `apiKey:` field (elasticsearchPlugin({ apiKey: process.env.ATLAS_ES_API_KEY! })), so one
 # variable covers BOTH the CLI auto-read path (above) and an explicit static config.
--- a/.env.example
+++ b/.env.example
@@ -63,6 +63,12 @@ DATABASE_URL=postgresql://atlas:atlas@localhost:5432/atlas
 # Pre-configured for the demo dataset seeded by `bun run db:up`.
 # When atlas.config.ts defines datasources, this env var is not needed.
 ATLAS_DATASOURCE_URL=postgresql://atlas:atlas@localhost:5432/atlas_demo
+# ATLAS_DATASOURCE_EXPECTED=false              # Set ONLY on a deployment that intentionally has no analytics datasource (a
+#                                             # multi-tenant deploy whose connections live per workspace, a knowledge-only box).
+#                                             # /health then stops degrading over the missing datasource. Leave it UNSET
+#                                             # otherwise — that is what keeps a forgotten ATLAS_DATASOURCE_URL visible. Set it
+#                                             # per service; it overrides `datasourceExpected` in atlas.config.ts. See
+#                                             # docs/deployment/multi-datasource for what it does NOT do.
 # ATLAS_DATASOURCE_URL=mysql://user:pass@localhost:3306/mydb
 # ATLAS_DATASOURCE_URL=elasticsearch://my-cluster.es.io:9243   # Elasticsearch (HTTPS; append ?ssl=false for a local cluster). opensearch://host:9200 selects the OpenSearch engine
 # ATLAS_ES_CLOUD_ID=                           # Elastic Cloud ID (<name>:<base64>) — endpoint alternative to the elasticsearch:// URL above
@@ -73,15 +79,6 @@ ATLAS_DATASOURCE_URL=postgresql://atlas:atlas@localhost:5432/atlas_demo
 # ATLAS_ES_AWS_REGION=                         # Selects AWS SigV4 (Amazon OpenSearch Service); keys from AWS_ACCESS_KEY_ID / AWS_SECRET_ACCESS_KEY (+ AWS_SESSION_TOKEN)
 # ATLAS_ES_AWS_SERVICE=                        # SigV4 service code (default "es"; "aoss" for OpenSearch Serverless)
 # ATLAS_ES_ENGINE=                             # Optional engine override: elasticsearch | opensearch (wins over the URL scheme)
-# ATLAS_DATASOURCE_EXPECTED=false              # Declare that this deployment intentionally has NO process-level analytics
-#                                              # datasource — a multi-tenant deploy whose connections live per workspace, or a
-#                                              # knowledge-only / brain-only box. /health then reports the datasource component
-#                                              # as "disabled" without degrading the top-level status. Leave it UNSET on a
-#                                              # deployment that should have one: undeclared means a missing ATLAS_DATASOURCE_URL
-#                                              # still degrades /health, which is how a misconfiguration stays visible. Set it
-#                                              # per service — it overrides `datasourceExpected` in atlas.config.ts, which several
-#                                              # services may share. It never silences a datasource that IS configured and failing
-#                                              # (that is still error + 503).
 # NOTE: ATLAS_ES_API_KEY is the canonical name. The atlas.config.ts examples wire the same var into the
 # plugin's `apiKey:` field (elasticsearchPlugin({ apiKey: process.env.ATLAS_ES_API_KEY! })), so one
 # variable covers BOTH the CLI auto-read path (above) and an explicit static config.
--- a/apps/docs/content/docs/deployment/multi-datasource.mdx
+++ b/apps/docs/content/docs/deployment/multi-datasource.mdx
@@ -395,6 +395,44 @@ view) but does **not** change the aggregate `status` or trip a 503. This keeps o
 workspace's misconfigured connection from pulling the whole shared region out of
 the load balancer.
 
+### Deployments with no datasource at all
+
+Some deployments legitimately have no process-level analytics datasource: a
+multi-tenant deploy where every connection is registered per workspace, or a
+knowledge-only / brain-only box that answers from the Knowledge Base and Company
+Brain. By default Atlas reports a missing datasource as `degraded` — that is what
+makes a forgotten `ATLAS_DATASOURCE_URL` visible instead of silent.
+
+Declare the absence so the aggregate status stays usable:
+
+```bash
+ATLAS_DATASOURCE_EXPECTED=false
+```
+
+or, in `atlas.config.ts`:
+
+```ts
+export default defineConfig({
+  datasourceExpected: false,
+});
+```
+
+`/api/health` then reports `components.datasource.status: "disabled"` — the state
+that means *intentionally absent* — and leaves the top-level `status` at `ok`.
+Three things this deliberately does **not** do:
+
+- It does not infer intent from the absence. Leave it unset and a missing
+  datasource still degrades, exactly as before.
+- It does not silence a datasource that **is** configured and failing. That is
+  still `status: "error"` + HTTP 503, and the region still leaves the load
+  balancer.
+- It does not disable anything. `ATLAS_DATASOURCE_EXPECTED` only describes what
+  this deployment expects; it never changes which connections are registered.
+
+Set the env var per service when several services share one `atlas.config.ts` —
+it overrides the config field in both directions, so one shared file cannot
+declare an expectation on another service's behalf.
+
 ---
 
 ## Troubleshooting
--- a/apps/docs/content/docs/deployment/multi-datasource.mdx
+++ b/apps/docs/content/docs/deployment/multi-datasource.mdx
@@ -412,14 +412,19 @@ ATLAS_DATASOURCE_EXPECTED=false
 or, in `atlas.config.ts`:
 
 ```ts
+import { defineConfig } from "@atlas/api/lib/config";
+
 export default defineConfig({
   datasourceExpected: false,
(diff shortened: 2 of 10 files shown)


msywulak:
## Review panel — 1 must-fix, applied in 6f113e3b7

Four specialist reviewers over the diff, fresh context. All four independently converged on the same defect, which was real.

### Must fix (applied)

**`datasource-expectation.ts` — the unrecognized-value branch fell through to the config file.** It warned and then kept going, so `ATLAS_DATASOURCE_EXPECTED=ture` + `datasourceExpected: false` in a shared `atlas.config.ts` resolved to `false` — a silent green — while the log line asserted "treating the deployment as expecting a datasource, so a missing one still degrades". Exactly backwards, on the operator who was already debugging a typo, in the shared-image shape (#3958) the module is built around. A failed declaration is now a declaration: warn, resolve to expected, do not consult the config. Pinned by `an unrecognized value does NOT fall through to a config that says false`.

### Should consider (applied)

- **`ATLAS_DEMO_DATA=true` was maskable.** It is itself a declaration that a datasource is expected, so a Neon integration that never provisioned would have gone from `degraded` + `MISSING_DATASOURCE_URL` to `ok` under a coarse declaration. It now outranks the declaration.
- **The whole `config.ts` half was untested end-to-end** — the reviewer mutated away both the Zod field and the resolver spread and got 179 tests green, while the documented `defineConfig({ datasourceExpected: false })` path silently stopped working (stripping `z.object`). `config.test.ts` now pins it, including `validateAndResolve({})` omitting the key — the tri-state invariant that makes `?? true` a wrong refactor, previously comment-only.
- **The `dsNotConfigured &&` narrowing was unpinned.** Dropping it kept all 47 health tests green while producing a component that reads `"down"` and claims no datasource is expected — mid-outage, on the operator's tile. One assertion closes it.
- **The registry arm of `hasDatasource` was untested** under a declaration (multi-tenant registers `default` from config rather than the env var). Added.
- **Docs were wrong:** `disabled` does not mean "intentionally absent" — an undeclared absence reports `disabled` too, only the `message` differs. Rewrote, and added the residue the declaration does *not* clear (`checks.datasource`, the startup warning) plus the unrecognized-value behaviour.
- Citation corrections: `#4124` was a CLI auth bug, not evidence for the deployment shape (now points at `deploy/api/atlas.config.ts`); noted the `#4457` precedent is deliberately *inverted* (backups derive expectation, this must not); moved the `.env.example` block out of the Elasticsearch section it had wedged itself into.

### Declined

- Returning `{ expected, source, recognized }` instead of `boolean` (the `BackupCadence` shape). `boolean` is right at the only call site; the record is speculative until something branches on `source`.
- `_setConfigForTest(config: Partial<ResolvedConfig>)` to delete 22 `as any` casts repo-wide — real, but out of scope for this PR.
- Rejecting a contradictory `{ datasources: {default}, datasourceExpected: false }` at parse time — inert in steady state; not worth widening the diff.

### Clean axes

No new catch blocks, no `any` or non-null assertions in production code, no secrets in comments, no test-discipline violations. `/ci` 32/32 PASS after the fixes.

msywulak:
## Remaining panel findings — applied in 094540534

Everything the panel raised is now either fixed or has a recorded reason. Second pass:

**`resolveDatasourceExpectation()` returns `{ expected, source, recognized }`**, with `isDatasourceExpected()` as the boolean projection. This is the `BackupCadence` shape the module already claimed to mirror, and it collapses three findings at once: the warn message now *derives* from the accepted-vocabulary const instead of restating it (so it cannot drift), `recognized` makes an unparseable declaration observable as a value, and the config is injectable so tests resolve against a fixture rather than the process singleton.

**`/health` surfaces the unrecognized declaration in `warnings[]`.** Once-per-process logging is the right volume for a public polled endpoint but the wrong surface — by the time an operator wonders why their declaration did nothing, that line is long gone. The offending value stays out of the response body (`/health` is unauthenticated; the value is operator-supplied) and goes to the log instead.

**`_setConfigForTest(config: Partial<ResolvedConfig> | null)`.** Every caller drives one field and casts away the rest; the cast is now told once, in the helper, where it is visible and explained. Cleared 5 `as any` + 4 oxlint-disables across 3 files — the remaining `_setConfigForTest` call sites don't use that pattern, so the "~100 casts" estimate was high.

**`validateAndResolve` warns on a contradictory `datasourceExpected: false` + `datasources`.** A warning, not a throw: a registered datasource makes the declaration inert, so refusing to boot over a contradiction that changes no behaviour is the worse trade.

**Tests added:** expectation resolution pinned per-input with `source` + `recognized` (labelled per value rather than looping inside one assertion), the `/health` warnings surface in both directions, and — from the panel's finding 5 — that a declared-absent deployment does **not** suppress the startup warning. That last one matters: suppressing warnings to make a dashboard look clean is #4854 pointed the other way.

**Citations:** `#1981` → `#1981` (internal-DB half) + `#3907` (primary-datasource isolation).

### Deliberately not pinned

The warn-**once** latch. Observing it needs `mock.module("@atlas/api/lib/logger")`, and mocking `createLogger` process-wide hangs the file — pino loggers carry prototype methods a spread drops, and every module in the `lib/config` graph resolves its logger at import time. The latch governs log volume, not behaviour; `recognized` is returned on every call regardless and is pinned. The reason is recorded in the test file so the next person doesn't rediscover the hang.

`/ci` 32/32 PASS.

Issue #4878, opened 2026-07-28
sec(deps): npm advisory remediation — the Dependabot handoff #4822 assumed has never worked


msywulak:
Follow-up from #4822. The image scan reports npm advisories to code scanning and hands remediation to Dependabot by design — `image-scan.yml` states it in the header:

> Library findings (npm, Go modules) inside the images, because Dependabot already owns them and can open the fix PR.

**That handoff has never worked.** Dependabot shows 0 open alerts (20 lifetime, all fixed, most recent 2026-06-20 — a month before image scanning landed), while Trivy has ~34 fixable npm advisories open on `main`.

## Root cause — three independent failures

**1. The dependency graph does not parse `bun.lock`.**

| Source | Packages |
|---|---|
| `bun.lock` | **2146** |
| GitHub `dependency-graph/sbom` | **579** |

Querying the SBOM for `undici`, `shell-quote`, `brace-expansion`, `form-data`, `fast-uri`, `protobufjs` returns `[]`. The SBOM holds only direct deps read off `package.json` — **1567 packages are invisible to Dependabot.**

**2. No `.github/dependabot.yml` exists.** `dependabot_security_updates` is `enabled`, so it *would* open fix PRs — but only for alerts, and the graph produces none for these packages.

**3. Caret ranges mask the direct deps too.** `next: ^16.2.9` already permits the fixed `16.2.11`, so the manifest Dependabot *can* read looks satisfiable while `bun.lock` pins the vulnerable build. This is why even in-SBOM packages stayed silent.

## Scope — most of this is not a version bump

**13 of 15 flagged packages are transitive-only:** `undici`, `tar`, `axios`, `sharp`, `form-data`, `shell-quote`, `postcss`, `brace-expansion`, `uuid`, `protobufjs`, `immutable`, `fast-uri`. Vulnerable `hono@4.12.23` arrives via `@modelcontextprotocol/sdk` (the direct pin is already `^4.12.27`); vulnerable `js-yaml@4.2.0` arrives via `cosmiconfig` and three `fumadocs-*` packages.

There is no `overrides`/`resolutions` block in the root `package.json`. Clearing these requires authoring overrides or bumping parents — not `bun update`.

`next` and `js-yaml` are direct and were bumped separately (see linked PR).

## ⚠️ Freeze constraint — the bumps deploy to production

[...]

PR #4881, opened 2026-07-29
fix(deploy): scope the api runtime image to @atlas/api's production closure

### LINKED ISSUE #4880: sec(deploy): api runtime image ships the whole monorepo — 63% unused, 68% of npm findings unreachable
The api runtime image ships the **entire monorepo's** dependency tree — web, docs, www, examples, and all 20+ plugins — because `deploy/api/Dockerfile:63` runs `bun ci` at the workspace root and line 168 copies `/app/node_modules` wholesale into the runtime stage.

| | Packages |
|---|---|
| The api runtime actually needs (`packages/api` prod closure) | **773** |
| The api image ships | **2094** |
| Unnecessary | **1321 (63%)** |

This is not primarily a "dev dependencies leak" problem — only 5 flagged packages are strictly dev-only. It is a **workspace-scoping** problem, and it dominates the npm findings from #4822.

## Impact on the scan surface

All 199 npm-path Trivy alerts live in `app/` — **the api image only**. `deploy/web` and `deploy/docs` never copy `node_modules` into their runtime stages (standalone build / static export), so they carry none of this.

**137 of 199 (68%) are inside packages the api never loads:**

| Severity | Total | In never-loaded packages |
|---|---|---|
| critical | 3 | **3** |
| high | 71 | **50** |
| medium | 111 | 74 |
| low | 12 | 8 |

Top contributors, none reachable from `packages/api`:

| Alerts | Host package | Why it's there |
|---|---|---|
| **106** | `@esbuild/linux-x64` | Go binary; Trivy reports its vendored Go stdlib |
| 9 | `next` | root `dependencies` |
| 4 | `postcss` | root `dependencies` |
| 3 | `@typescript/native-preview-linux-x64` (tsgo) | other workspace devDep |
| 3 | `@oxlint-tsgolint/linux-x64` | other workspace devDep |

**All three criticals are Go stdlib inside `@esbuild/linux-x64`** — build tooling in a production runtime image.

## What `--filter` actually fixes (measured, not assumed)

`deploy/docs/Dockerfile:5-13` documents that a **single-package** install over one workspace cannot resolve the `workspace:` protocol. That blocker does **not** apply to `--filter`, which still resolves from the root. Verified empirically on bun 1.3.13:

```
bun install --frozen-lockfile --filter='@atlas/api' --production --ignore-scripts
```

| | Before | After |
|---|---|---|
| packages installed | 3218 | **1909** |
| `node_modules` | 2.2 GB | **1.5 GB** |

`bun.lock` is untouched (`--frozen-lockfile` holds). It correctly drops other workspaces — `fumadocs-mdx`, `tsup`, `obsidian` all gone, clearing tsgo and oxlint-tsgolint (**6 alerts**).

**But it does not go far enough.** Two gaps, both verified:

**1. `--production` / `--omit=dev` do NOT drop the filtered workspace's own devDependencies.** `drizzle-kit@0.31.10` is declared *only* in `packages/api/package.json` `devDependencies`, yet survives both flags (identical 1909-package result). It drags in `esbuild@0.18.20` → **66 alerts, including 2 of the 3 criticals**.

`drizzle-kit` is genuinely dev-only: it appears at runtime only in SQL comments and `__tests__/migrate.test.ts`. The runtime migration path is the custom `runMigrations` in `packages/api/src/lib/db/migrate.ts`, loaded via `internal.ts:1056`. Nothing in the runtime image needs it.

**2. Root `package.json` declares 59 `dependencies`, including app-level packages** — `next`, `react-dom`, `tailwindcss`, `fumadocs-core`, `fumadocs-openapi`, `fumadocs-ui`. These belong to web/docs, not the api, but `--production` won't drop them (they're `dependencies`, not `devDependencies`) and `--filter` won't either (the root workspace is always installed). This is what keeps `next` (9 alerts), `sharp`, and `postcss` in the api image.

## Key files
- `deploy/api/Dockerfile` — `:63` root `bun ci`, `:168` wholesale `node_modules` COPY
- `package.json` — root; 59 `dependencies` incl. app-level packages
- `packages/api/package.json` — `drizzle-kit` devDep
- `deploy/docs/Dockerfile:5-13` — prior art on why single-package install fails
- `deploy/web/Dockerfile`, `deploy/docs/Dockerfile` — unaffected; no runtime `node_modules` COPY

## Acceptance criteria
- [ ] api Dockerfile installs scoped to `@atlas/api` rather than the full workspace
- [ ] `drizzle-kit`/`esbuild@0.18.20` absent from the runtime image — clears 2 of 3 criticals
- [ ] App-level deps (`next`, `react-dom`, `tailwindcss`, `fumadocs-*`) moved out of root `dependencies` into the workspaces that use them, or into root `devDependencies`
- [ ] Boot Smoke + Deploy Validation green — the runtime resolves every dynamic import (`@atlas/mcp/hosted`, `@useatlas/webhook-publisher`, `@atlas/okf-bundle/wire`, plugin `peerDep` symlinks). The Dockerfile's defensive symlink-rebuild comments (`:146`, `:175`) show this COPY chain is fragile — **test the negative**: confirm the image still boots and serves `/mcp/{workspace_id}/sse`, don't just confirm it builds
- [ ] Re-scan shows the api image's npm alert count dropping from 199

## Dependencies
- Held until after the `v0.2.0` tag — touches `bun.lock`/`package.json`, which deploy docs+www direct from `main` (see #4878)
- Complements #4878 (overrides). **Do this first** — it clears ~3× what the override work does, with no version risk
- Related: #4822 (image scanning), #4738 (watchPattern closure audit)

## Not in scope
- The 62 alerts in packages the api genuinely loads — those need #4878's overrides
- `usr/bin` Go findings (39) — base-image binaries, covered by the `.trivyignore` baseline expiring 2026-09-01

msywulak:
Closes #4880.

## ⛔ DO NOT MERGE before the `v0.2.0` tag (held to 2026-08-03)

This touches `bun.lock` and three `package.json` files, and the `docs` + `www` Railway services deploy **direct from `main`** with no staging twin. Green and ready — leave it for a human to merge after the tag.

## Measured result

Both images built locally from the same context; baseline is clean `origin/main` @ `9cd40cbba`.

| | baseline | this PR | |
|---|---|---|---|
| image size | 2.26 GB | **1.63 GB** | −28% |
| `/app/node_modules` | 2241 MB | **1490 MB** | −34% |
| isolated-store entries | 1613 | **958** | −41% |
| `esbuild` `package.json` files | 12 | **0** | |
| `@esbuild/linux-x64` dirs | 5 | **0** | |
| `drizzle-kit` `package.json` | 1 | **0** | |

`@esbuild/linux-x64` alone was 106 of the 199 npm findings and **all 3 criticals**.

## What changed

The Dockerfile's `deps` stage split into three. `manifests` holds the ~40 `package.json` COPYs; `deps` builds the full tree the **builder** needs (tsup, bundlers); `prod-deps` installs only what **ships**:

```
bun install --frozen-lockfile --ignore-scripts --production \
  --filter='@atlas/api' \
  --filter='@useatlas/{clickhouse,snowflake,bigquery,elasticsearch}'
```

`--filter` still resolves `workspace:` from the root, so the single-package blocker documented in `deploy/docs/Dockerfile:5-13` doesn't apply. The runner now takes `node_modules` from `prod-deps`. That's a one-line swap rather than a re-plumbing because the isolated-store slugs — peer-resolution hash suffix included — are **identical** between the two installs (verified: every slug in the scoped store is present verbatim in the full one), so the workspace symlinks copied from `builder` stay valid and links to filtered-out packages dangle by design.

The four adapter plugins are named explicitly because they declare **no dependencies at all**, only peers. Measured: adding them costs zero extra packages. Keep that list in lockstep with `plugins[]` in `atlas.config.ts` — same lockstep the symlink-rebuild block already demands.

Then a targeted prune of build tooling that survives the production filter (see below): all `esbuild`, `@esbuild/*`, `@esbuild-kit/*`, `drizzle-kit`, `@drizzle-team/brocli`, `tsx`. These are left as **dangling symlinks** — no `package.json`, no binaries, nothing for a scanner to find. `drizzle-ORM` stays; it's a real dependency of `schema.ts`.

## Three of the issue's premises didn't survive measurement

**1. `--production` *does* drop the filtered workspace's own devDependencies.** The issue concluded otherwise from `drizzle-kit` surviving. It survives because **better-auth declares it as an optional `peerDependency`** (`drizzle-kit: ">=0.31.4"`) — removing it from `packages/api` devDeps changes nothing (byte-identical 1909-package result). So `--production` was silently dropping `@effect/sql` and `@effect/sql-pg`, both imported by `lib/db/internal.ts` — the internal-DB spine that loads `runMigrations` — plus `drizzle-orm`. All three moved to `dependencies`.

**2. Root `dependencies` contribute nothing to the api image.** Under `--filter`, root `node_modules` has **0** top-level entries. Tested directly: moved 20 app-level packages (`next`, `react-dom`, `tailwindcss`, `fumadocs-*`, …) out of root `dependencies`, regenerated the lock, re-ran the scoped install → **byte-identical store, 944 entries**. `next` and `sharp` enter as better-auth optional peers; `postcss` via `@tailwindcss/postcss`/`shadcn`/`tsup` peers.

**So part 3 of the deliverable is deliberately not done** — it cannot move this issue's numbers. It's still reasonable hygiene for the *builder* stage and dev installs, but that's a different issue and it fights syncpack over the duplicated `fumadocs ~16.9.3` pin for no image-side gain.

**3. `--omit=peer` is not usable.** It's tempting (1806 pkgs / 939 MB, drops all three esbuild versions declaratively) but it also strips `graphql` out from under `@linear/sdk` and every plugin's `@useatlas/plugin-sdk`, `zod`, `hono` and `ai` peers. Rejected in favour of the explicit prune.

## Two latent bugs the scoping exposed

Neither is caused by this change — both were pre-existing reliance on root hoisting, made visible by scoping.

- `@effect/sql` + `@effect/sql-pg` + `drizzle-orm` were devDeps of `@atlas/api` but imported from runtime source → moved to `dependencies`.
- **`ee/package.json` declared neither `ai` nor `zod`** despite importing both across 100+ files, resolving them through the root-hoisted install → both added.

Both were caught by the new assertion, not by reading code.

## Verification — testing the negative

A green build proves nothing here: bun resolves lazily, so a missing package isn't a boot failure, it's a module-not-found thrown whenever some route or scheduler tick first reaches the import. Two build-time assertions:

- **Static** — `deploy/api/scripts/assert-runtime-imports.ts` sweeps every bare import in shipped source against the shipped tree. Uses `Bun.Transpiler.scanImports`, not a regex (the source is full of prose like `from "sent"`; a regex reports those as missing packages). Two documented waivers: `@playwright/test` (lazy, try/caught, degrades to `browser_unavailable` — the image has never provisioned Chromium, so this only moves the failure from launch-time to import-time inside that same handler) and `@google-cloud/bigquery` (unbundled peer, customer-supplied, absent before this change too).
- **Dynamic** — really imports the 5 boot-loaded plugins from each plugin's own directory, plus `@atlas/mcp/hosted`, `@useatlas/webhook-publisher`, `@atlas/okf-bundle/wire` from `packages/api`.

The static sweep is what found the `@effect/sql` and `ee` gaps. It was itself negative-tested: removing `node-sql-parser` from the store makes it fail and name the two importing files.

**Booted the built image** against a real Postgres:

- `/api/health` → **200**; migrations applied end-to-end (which exercises `internal.ts` → `@effect/sql`/`@effect/sql-pg` and `runMigrations`)
- `All plugins initialized: clickhouse-datasource, snowflake-datasource, bigquery-datasource, elasticsearch-datasource, chat-interaction`
- `GET /mcp/test-ws/sse` → **401**, and a genuinely unmounted path → **404**. The route is registered; this is the exact failure the `packages/api/src/api/index.ts:482` comment describes, and it isn't happening.
- In the *running* container, as the `atlas` user: all three dynamic-import targets resolve (`@atlas/mcp/hosted` 6 exports, `@useatlas/webhook-publisher` 4, `@atlas/okf-bundle/wire` 14), and all 5 boot-loaded plugins import cleanly.

`/ci` — **32/32 PASS**, including `syncpack` and `template-drift`, and with `TEST_DATABASE_URL` set so the `-pg` suites actually ran.

## Follow-ups for #4878

`next` (9 alerts), `sharp` and `postcss` remain. They're **not** from root `dependencies` — they're better-auth optional peers and `@tailwindcss/postcss`/`shadcn`/`tsup` peers, which is exactly the shape overrides can drop cleanly. I left them rather than extending the prune: unlike esbuild/drizzle-kit, better-auth genuinely integrates with Next at runtime, and blunt-removing it risks a silent change to a lazy path for 9 alerts.


msywulak:
Added the lockstep guard discussed above (`790594f4c`).

This PR left `plugins[]` in `atlas.config.ts` as the source of truth for **three** lists in the Dockerfile — the prod-deps `--filter` scope, the symlink-rebuild loop, and the new runtime-import assertion loop — held in lockstep by comments alone. The assertion loop is what makes the other two fail loudly, but it iterates a *hardcoded* list, so **adding a sixth boot-loaded plugin without touching the Dockerfile would have built green** and then failed exactly the way this issue describes: a dangling `@useatlas/plugin-sdk` peer that doesn't break boot, just silently reports the adapter unavailable.

`scripts/check-plugin-lockstep.ts` derives the boot-loaded set from `atlas.config.ts` (imports mapped to plugin dirs, intersected with the factory calls actually invoked inside `plugins: [ … ]`) and asserts each appears in all three places — with `--filter` satisfiable *either* by an explicit entry *or* by being a dependency of `@atlas/api`, which is why `@useatlas/chat` correctly needs no entry. Stale assertion-loop entries are flagged too. Extra symlink-loop entries stay legal (`e2b`/`daytona` are lazy, `salesforce` is defensive).

Backstopped per the repo convention with `scripts/__tests__/check-plugin-lockstep.test.sh` — 11 fixtures covering each list independently, the add-a-plugin scenario, a stale entry, both legitimate exemptions, and **two vacuous-pass probes** proving the gate fails loudly if its own parsing goes stale rather than passing on an empty parse.

Wired into `scripts/ci-local.sh` and `.github/workflows/ci.yml`. Local `/ci` is **33/33** (was 32) with `-pg` suites enabled.

Still holding for the `v0.2.0` tag.

msywulak:
## Measured re-scan — Trivy 0.72.0 (the pinned CI version), both images built locally

| Trivy class | baseline (`9cd40cbba`) | PR #4881 | |
|---|---|---|---|
| `lang-pkgs / gobinary` | 112 | **0** | −112 |
| `lang-pkgs / node-pkg` | 76 | 67 | −9 |
| `os-pkgs / debian` | 325 | 325 | unchanged |
| **total** | **513** | **392** | **−121 (−24%)** |

Non-OS severity: **criticals 4 → 1**, **highs 67 → 25**, mediums 103 → 36, lows 12 → 5.

The `gobinary` class is where the esbuild findings lived — Trivy reads the vendored Go stdlib inside the platform binary. It confirms the issue's figure exactly:

```
67  @esbuild+linux-x64@0.18.20/.../bin/esbuild
39  @esbuild+linux-x64@0.25.12/.../bin/esbuild     = 106
 3  @oxlint-tsgolint+linux-x64@0.24.0/.../tsgolint
 3  @typescript+native-preview-linux-x64/.../tsgo
```

All three Go-stdlib criticals (`stdlib@v1.20.7` ×2, `stdlib@v1.23.12`) are gone, and the whole class is now empty.

### ⚠️ Correction to this PR's description

The body says "all 3 criticals" were in `@esbuild/linux-x64`. True of the three the issue tracked — but the scan shows a **fourth critical neither the issue nor I had counted**: `tar@7.5.16` / CVE-2026-59873, an npm finding, which **survives this PR**. So the image is not critical-free; it goes 4 → 1. Fix exists (`7.5.19`), and both requirers (`@daytonaio/sdk`, `e2b`) want `^7.5.11`, so an override is a clean bump — filed under #4878 below.

`os-pkgs` is byte-identical at 325, as expected — same pinned base image, and that is the only class the image-scan gate blocks on.

### #4878 right-sized

**67 findings across 22 packages, and all 67 have a published `FixedVersion`** — 26 of them HIGH/CRITICAL. Nearly all transitive, so `overrides` is the right mechanism. Highest value first:

| findings | package | fix | note |
|---|---|---|---|
| 5 | `tar@7.5.16` | 7.5.19 | **the last critical**; via `@daytonaio/sdk` + `e2b` |
| 11 | `protobufjs@8.0.1` | 8.0.2 | 5 high; via `@opentelemetry/otlp-transformer` |
| 10 | `axios@1.17.0` | 1.18.0 | 1 high |
| 11 | `undici@7.27.2` + `@6.24.1` | 7.28.0 | 4 high across two copies |
| 8 | `hono@4.12.23` | 4.12.27 | we already *declare* `^4.12.27`; only a stale transitive copy — an override just dedupes it |
| 3 | `postcss@8.4.31` | 8.5.12 | 2 high; arrives with `next`, itself a better-auth optional peer |

So the pre-work sequencing paid off, but #4878 is **not** obsolete — it still has a critical and 25 highs to clear, on a much smaller and entirely fixable surface.

Issue #4822, opened 2026-07-26
ci(security): add container-image vulnerability scanning — the half of the scanning claim that did not exist


msywulak:
> *This was generated by AI during triage.*

Split out of the legal-surface audit in PR #4821.

## Problem

`/dpa` Annex II and `/privacy` §9 claimed **"automated vulnerability scanning of container images and dependencies."** The dependency half is real — Dependabot security updates are enabled, CodeQL default setup runs on JS/TS + actions, and secret scanning with push protection is on. **The container-image half does not exist**: no Trivy, Grype, Snyk, or equivalent appears in any workflow.

PR #4821 corrected the claim rather than leaving it — container-image scanning now sits in an explicit *"on the public roadmap and not represented as in place today"* clause on both pages. This issue is the work that earns it back.

Worth doing on its own merits regardless of the legal copy: Atlas ships seven Dockerfiles, several of them customer-facing scaffolds, and base-image CVEs are the one class none of the current tooling covers. CodeQL analyzes source; Dependabot watches the manifest; neither sees a vulnerable OS package baked into the runtime layer.

## Approach

Add an image-scanning gate to CI covering the images Atlas actually builds and ships. Trivy is the natural default (free, no account, good bun/Node coverage, SARIF output that GitHub code scanning ingests directly), but the choice is open — the requirement is a scanner that fails CI on fixable HIGH/CRITICAL findings in a built image.

Scope the scan to images, not just Dockerfile linting: the whole point is catching CVEs in resolved base-image layers, which requires scanning a built artifact.

Two decisions the implementer should make and record in the PR:

1. **Severity threshold and whether unfixed CVEs block.** Failing on unfixed CRITICALs makes the gate red for reasons no one can action, which trains people to ignore it. The usual answer is to fail on *fixable* HIGH/CRITICAL and report the rest.
2. **Where it runs.** A full image build per PR is expensive; the deploy images already build in Deploy Validation. Reusing that build, or running the scan on a schedule against the published images plus on Dockerfile/lockfile changes, are both defensible. Say which and why.

## Key files
- `.github/workflows/ci.yml` — the gate lands here or in a sibling workflow
- `deploy/api/Dockerfile`, `deploy/web/Dockerfile`, `deploy/docs/Dockerfile` — the deployed images; these matter most
- `examples/docker/Dockerfile` — the self-host reference
[...]

PR #4884, opened 2026-07-29
chore(deps): npm advisory remediation — fix the Dependabot handoff that never worked (#4878)

### LINKED ISSUE #4878: sec(deps): npm advisory remediation — the Dependabot handoff #4822 assumed has never worked

msywulak:
Closes #4878. Refs #4822.

## The spike came first, and it changed the plan

AC1 was a real unknown, so it was verified empirically before anything was built. Two findings, both measured:

**1. The dependency graph does not parse `bun.lock` — and it never resolved versions at all.** The graph's 568 npm entries collapse to **150 unique names**, duplicated once per workspace manifest that declares them (`zod` ×34, `ai` ×27, `hono` ×25), and the versions are stored as **unresolved caret ranges**:

```
picocolors@^1.1.1   @clack/prompts@^1.1.0   react-grid-layout@^2.2.3
```

Against **2133** resolved packages in `bun.lock`. No other lockfile exists in the repo. This *explains* root cause #3 in the issue rather than restating it: the graph holds `next: ^16.2.9` as a range, so GitHub reads it as already-satisfiable. It isn't that caret ranges mask the problem — the graph has no idea what is installed. GitHub documents the parsed formats as package-lock.json / yarn.lock / pnpm-lock.yaml only.

**2. The handoff is structurally impossible, not misconfigured.** `bun` is a **separate Dependabot ecosystem** from `npm`:

| ecosystem | version updates | security updates |
|---|---|---|
| `npm` | ✓ | ✓ |
| `bun` (≥v1.1.39) | ✓ | **✗** |

So `image-scan.yml`'s "Dependabot already owns them and *can open the fix PR*" cannot be repaired by configuration. For the whole window between #4822 and now, this scanner was the only thing looking while its own header said otherwise.

**The `package-lock.json` fallback was rejected deliberately.** It doesn't just risk drift — it would make Dependabot open fix PRs against a file bun never reads. A merged, green PR that fixes nothing is worse than no PR, and it is the same "reads as coverage while providing none" trap the header warns about one bullet up.

## What shipped instead

| concern | mechanism |
|---|---|
| detection | Trivy on the image (`image-scan.yml`) — the authoritative surface, since that is what ships |
| **visibility** | `scripts/submit-dependency-snapshot.ts` + `.github/workflows/dependency-graph.yml` — submits the resolved closure (**2085 packages**) via the Dependency Submission API. Regenerated from the committed lockfile each run, never committed, so it cannot drift |
| **remediation** | root `package.json` `overrides` (+ one parent bump) |
| freshness | `.github/dependabot.yml` — bun ecosystem, version updates only |

`open-pull-requests-limit: 3`, non-majors grouped into one PR. Majors are **ignored on purpose**: `/deps-update` owns them as Group B because several need coupled bumps (fumadocs lockstep, zod × MCP sdk).

## AC3 — measured, not asserted

Trivy **0.72.0** (the version pinned in `image-scan.yml`), `--scanners vuln`, no ignorefile, two locally built api images. The 66 baseline reproduced exactly (67 raw, minus the dismissed `@better-auth/scim` HIGH).

```
node-pkg          66 -> 34      C1 H24 M36 L5  ->  C0 H16 M16 L2
gobinary           0 ->  0      already cleared by #4881
os-pkgs/debian   325 -> 325     out of scope, separate baseline
```

**The only CRITICAL is gone** (CVE-2026-59873, `tar`). HIGH 24 → 16. **No new findings introduced.**

Cleared: `axios` −10 · `hono` −8 · `tar` −5 · `undici@6.24.1` −4 · `fast-uri` −2 · `body-parser` −1 · `form-data` −1 · `shell-quote` −1

Every override was checked with semver against **every** declared consumer range, so none forces a consumer off-range. `undici@6.24.1` was cleared by a **parent bump expressed as an override**: `discord.js@14.27.0` moved from the exact pin `6.24.1` to `^6.27.0`, which is itself the fixed version, and it satisfies `@chat-adapter/discord`'s `^14.25.1`.

## The 34 that remain, and why they are not expressible

**bun supports only TOP-LEVEL overrides.** Path-scoped keys are **silently ignored** — asking for `minimatch/brace-expansion: 1.1.15` yielded `1.1.17`, no error. And a global override that violates a declared range is installed **silently**: declared `^6.0.0` + override `7.28.0` → installed 7.28.0, exit 0, no warning.

So any package present at two majors cannot be pinned per-major without forcing every consumer onto one version:

| remaining | n | why declined |
|---|---|---|
| `protobufjs@8.0.1` / `@7.6.2` | 13 | `otlp-transformer` pins 8.0.1 **exactly**; `@grpc/proto-loader` wants `^7.5.5`. Cleared only by the OTel `0.218 → 0.221` bump, which drops protobufjs entirely |
| `undici@7.27.2` | 7 | hoisted for `@vercel/sandbox` (`^7.27.1`, still latest at 2.9.0). Refreshing the entry cascaded a repo-wide re-resolve (tailwindcss, zustand, undici→8.9.0) — a Group A sweep, not a security fix |
| `postcss@8.4.31` | 3 | `next@16.2.12` pins it **exactly** and is already latest |
| `brace-expansion@2.1.1` / `@5.0.6` | 4 | three majors coexist under `minimatch` 3/5/9/10 (`^1.1.7`, `^2.0.x`, `^5.0.5`) |
| `uuid@8.3.2` / `@10.0.0` | 2 | `^8.3.0` + `^10.0.0` consumers; uuid 11 is a breaking API change |
| `sharp@0.34.5` | 1 | `next` optional peer `^0.34.5`; 0.35.0 is out of range and next is latest |
| `@opentelemetry/core` / `propagator-jaeger` | 2 | exact-pinned (core by ~74 packages); cleared by the same OTel bump |
| `@hono/node-server@1.19.14` | 1 | major cross for a server adapter, 1 MEDIUM — poor risk/reward |
| `@better-auth/oauth-provider@1.6.25` | 1 | fix is `1.7.0-beta.4`, the breaking beta already declined for `@better-auth/scim` |

**Deliberately not in this PR:** the OTel `0.218 → 0.221` bump would clear 13 more (incl. 5 HIGH), but it spans 5 manifests including `packages/web`, and OTel breakage tends to surface as *telemetry silently stopping* — the silent-failure class this repo forbids. It needs its own soak, not a ride-along on a security PR. Same for an `e2b` bump, which I tried and reverted: it moved e2b's own undici to 7.29.0 but left the hoisted 7.27.2 for `@vercel/sandbox` and **added a third copy** (`"undici8": "npm:undici@8.8.0"`), clearing nothing.

## Test the negative

**Overrides:** proven by the Trivy re-scan above, not by green CI.

**Visibility — half proven now, half gated on merge.** The submission pipeline was run for real against this branch: **`201 ACCEPTED`**, validating the payload format end-to-end. GitHub's own response states the rest:

> The snapshot was accepted, but it is not for the default branch. It will not update dependency results for the repository.

which is exactly why the workflow is `main`-only. Alerts therefore cannot be observed until this merges.

Pre-merge baseline is recorded so the post-merge check is a real before/after — **0 open Dependabot alerts** while every residual package is invisible to the graph:

```
INVISIBLE   undici          INVISIBLE   uuid
INVISIBLE   protobufjs      INVISIBLE   sharp
INVISIBLE   postcss         INVISIBLE   @hono/node-server
INVISIBLE   brace-expansion INVISIBLE   @opentelemetry/core
```

Trivy flags 34 findings across exactly those packages while Dependabot shows 0 alerts. **The residual set is the deliberately-outdated transitive dep the AC asks for** — real, known-vulnerable, and currently unseen — so no synthetic package is needed, and it is a stronger probe than one.

## Verification checklist for merge

- [ ] `Dependency Graph` workflow succeeds on `main`
- [ ] Dependabot alerts become non-zero for the residual set (`undici@7.27.2`, `protobufjs@8.0.1`, `postcss@8.4.31`) — **this is the AC2 negative test**
- [ ] **docs.useatlas.dev** healthy — `deploy/docs/railway.json` watches `bun.lock` **and** `package.json`, both touched here
- [ ] **www.useatlas.dev** healthy — watches `apps/www/**` (untouched here, so `No deployment needed` is the expected status)

Both deploy direct from `main` to production with no staging twin, so this ships unsoaked on merge. The v0.2.0 hold is satisfied (tagged @ `ba98acad6`, live in all 3 regions) but the hazard is permanent.

## Local gates

`bash scripts/ci-local.sh` — **33/33 PASS**. Real-Postgres (`*-pg.test.ts`) tests were **skipped locally** (`TEST_DATABASE_URL` unset, deliberately — a shared PG container is in use by parallel sessions and this change is dependency-only); CI's `api-tests (1/4)`–`(4/4)` shards run them against real Postgres.

`scripts/` is excluded from `tsconfig.json`, so the new script is not covered by `bun run type` — type-checked separately with `tsgo --noEmit --strict`, clean.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,72 @@
+version: 2
+
+# Dependabot configuration (#4878).
+#
+# ── What this can and cannot do ───────────────────────────────────────────────
+#
+# Atlas uses bun, and `bun` is a SEPARATE Dependabot ecosystem from `npm`, with a
+# different support matrix:
+#
+#   version updates    supported      <- this file drives them
+#   security updates   NOT supported  <- planned upstream, no timeline
+#
+# So this file delivers dependency *freshness*, not advisory remediation. It
+# cannot open a fix PR for a vulnerable package, and nothing in it makes the
+# GitHub dependency graph read `bun.lock`: the graph parses only
+# package-lock.json / yarn.lock / pnpm-lock.yaml, so bun.lock's transitive
+# closure is invisible to it. Measured on this repo — the graph held 150 unique
+# packages as unresolved caret RANGES read off the 53 package.json manifests,
+# against 2133 resolved packages in bun.lock.
+#
+# Transitive visibility is supplied separately by
+# .github/workflows/dependency-graph.yml, which submits the resolved bun.lock
+# closure through the Dependency Submission API. Alerts fire on submitted
+# dependencies; fix PRs still do not, per the matrix above.
+#
+# Remediation of npm advisories is therefore Trivy-on-image (image-scan.yml)
+# plus hand-authored `overrides` in the root package.json. See #4878.
+#
+# ── Why the limit is low, and majors are ignored ──────────────────────────────
+#
+# The workspace has 53 manifests and ~150 unique direct deps, so enabling
+# version updates opens a batch on the first run. Non-major updates are grouped
+# into a single PR and the open-PR ceiling is deliberately small.
+#
+# Majors are ignored on purpose: `/deps-update` already owns them as "Group B",
+# one deliberate PR per major, because this repo has several majors that need
+# coupled bumps (fumadocs-core/openapi lockstep, zod x @modelcontextprotocol/sdk).
+# An unattended major PR would be noise at best and a red CI lane at worst.
+#
+# ── Two things to know before merging a Dependabot PR here ────────────────────
+#
+# 1. bunfig.toml sets `minimumReleaseAge = 172800` (48h). Dependabot does not
+#    know about that gate, so it can propose a version younger than 48h that a
+#    local `bun install` will decline to resolve. If a PR's lockfile looks
+#    unreproducible, check the release age before hunting for a real cause.
+#
+# 2. Any PR touching bun.lock or package.json DEPLOYS docs.useatlas.dev on merge
+#    (deploy/docs/railway.json watchPatterns), direct from main to production
+#    with no staging twin. Verify the docs site after merging, not before.
+
+updates:
+  - package-ecosystem: bun
+    directory: "/"
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 3
+    labels:
+      - dependencies
+    commit-message:
+      prefix: "chore(deps)"
+    groups:
+      bun-minor-patch:
+        patterns:
+          - "*"
+        update-types:
+          - minor
+          - patch
+    ignore:
+      - dependency-name: "*"
+        update-types:
+          - version-update:semver-major
--- a/.github/workflows/dependency-graph.yml
+++ b/.github/workflows/dependency-graph.yml
@@ -1,0 +1,77 @@
+name: Dependency Graph
+
+# Submits the resolved `bun.lock` closure to the GitHub dependency graph (#4878).
+#
+# ── The gap this closes ───────────────────────────────────────────────────────
+#
+# The dependency graph parses package-lock.json, yarn.lock and pnpm-lock.yaml.
+# It does NOT parse bun.lock. On this repo that meant the graph held 150 unique
+# packages — read off the 53 package.json manifests as unresolved caret RANGES —
+# against 2133 resolved packages in the lockfile. Every transitive-only package
+# was invisible, so Dependabot produced no alerts for them and `image-scan.yml`
+# handed npm advisories to a Dependabot that could not see them.
+#
+# Storing ranges rather than resolved versions also made the graph unable to
+# distinguish a vulnerable build from a fixed one: `next: ^16.2.9` reads as
+# already-satisfiable while the lockfile pins the vulnerable version.
+#
+# ── What submitting buys, and what it does not ────────────────────────────────
+#
+# Buys:      Dependabot ALERTS on the full transitive closure.
+# Does not:  automatic fix PRs. `bun` is a separate Dependabot ecosystem from
+#            `npm` and does not support security updates (planned upstream, no
+#            timeline). Remediation stays with Trivy-on-image (image-scan.yml)
+#            plus hand-authored `overrides` in the root package.json.
+#
+# The snapshot is regenerated from the committed lockfile on every run and never
+# committed. That is deliberate: a checked-in package-lock.json would give the
+# same visibility but drift from bun.lock permanently, and would invite
+# Dependabot to open "fix" PRs against a file bun never reads — a green PR that
+# changes nothing about the build is worse than no PR.
+#
+# ── Why it runs on main only ──────────────────────────────────────────────────
+#
+# Dependabot alerts are keyed to the default branch, so a snapshot submitted from
+# a feature branch produces nothing actionable. Snapshots supersede per
+# (correlator, ref), so re-running is safe and idempotent.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "bun.lock"
+      - "**/package.json"
+      - "scripts/submit-dependency-snapshot.ts"
+      - ".github/workflows/dependency-graph.yml"
+  workflow_dispatch:
+
+# `contents: write` is what the Dependency Submission API requires to write a
+# snapshot. It is the only elevated scope here and the job runs no build steps.
+permissions:
+  contents: write
+
+env:
+  BUN_VERSION: "1.3.13"
+
+concurrency:
+  group: dependency-graph-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  submit:
+    name: Submit bun.lock closure
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+
+      - uses: oven-sh/setup-bun@e3914758a49697077f7bcd190d36582a61667aad # v2 + node24 (pending release)
+        with:
+          bun-version: ${{ env.BUN_VERSION }}
+
+      # No `bun install`: the converter reads the committed lockfile as text and
+      # needs no node_modules, which keeps this job off the dependency tree it is
+      # reporting on.
+      - name: Submit dependency snapshot
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: bun run scripts/submit-dependency-snapshot.ts
--- a/.github/workflows/image-scan.yml
+++ b/.github/workflows/image-scan.yml
@@ -26,12 +26,35 @@ name: Image Scan
 #   Unfixed CVEs, because a gate that is red for something nobody can action
 #   gets ignored, and an ignored gate reads as coverage while providing none.
 #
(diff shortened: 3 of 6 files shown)


msywulak:
## `Built image (web)` is red — pre-existing on `main`, not from this PR

Recording the evidence so this isn't misattributed to the dependency changes here.

**It is a Bun crash, not a dependency problem.** From this PR's job log, during `next build`:

```
Bun v1.3.13 (bf2e2cec) Linux x64 (baseline)
CPU: sse42 popcnt avx avx2
Args: "node" "/app/packages/web/node_modules/.bin/next" "build"
  Collecting page data using 3 workers ...
RSS: 117.90MB | Peak: 2.48GB
panic: Segmentation fault at address 0x13CB0
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
error: script "build" was terminated by signal SIGILL (Illegal instruction)
```

**Already failing on `main` before this branch existed:**

| run | head | Image Scan |
|---|---|---|
| 30463497995 | `0f0ee735c` (#4881) | **failure** |
| 30460092248 | `67abc1d6a` (#4882) | success |
| 30458349628 | `12a3dfd93` | success |

The failure is byte-identical to `main`'s — same step (`Build web image`), same command, same `exit code: 132`. #4881 did not touch `deploy/web/Dockerfile`; it changed `bun.lock`, so it most likely shifted build timing/memory enough to expose the crash (peak RSS 2.48GB, 3 workers) rather than introducing it.

**This branch does not cause or worsen it:** `cd packages/web && bun run build` succeeds locally on this branch (exit 0).

`Image Scan` is **not** a required status check — the required contexts are `ci`, `api-tests (1..4/4)`, `Analyze (javascript-typescript)`, `Deploy Validation`, `Symlink Stub Build`, `fork-pr-gate` — so it does not gate this merge. Flagging rather than fixing: a Bun-internal segfault is a separate root cause from npm advisory remediation, and it wants its own issue (candidate angles: Bun 1.3.13 worker memory during Next page-data collection, and whether the `baseline` Bun build is the right one for these runners).

Issue #4891, opened 2026-07-29
fix(ci): `next build` runs on bun-as-node inside the bun images and segfaults (SIGILL/132)


msywulak:
The `Image Scan` workflow's `Built image (web)` job failed once on `main` at `0f0ee735c` (#4881) with a segfault during `next build` under bun 1.3.13. Filing so it is not rediscovered from scratch; evidence is on #4884.

## It is a flake, not a standing red

Worth stating precisely, because it was initially read as "`main` is red on Image Scan" — it is not. Last 8 `Image Scan` runs on `main`:

| conclusion | sha |
|---|---|
| success | `1d36a9b8b` |
| success | `9f9a3674b` |
| success | `5db3c769e` |
| **failure** | `0f0ee735c` |
| success | `67abc1d6a` |
| success | `12a3dfd93` |
| success | `a8c86b003` |
| success | `6a0bf3916` |

One failure in eight. At current `HEAD` (`1d36a9b8b`) all 8 jobs pass, `Built image (web)` included. Nothing is blocked, and there is no break to go hunting for — anyone picking this up should expect to have to *reproduce* it rather than observe it.

`Image Scan` is also **not** a required status check (`ci`, `api-tests (1..4/4)`, `Analyze (javascript-typescript)`, `Deploy Validation`, `Symlink Stub Build`, `fork-pr-gate` are), so even a genuine red here does not block a merge.

## Why it still matters

A segfaulting `next build` is nondeterministic build behaviour in the image that ships `web` to production. A 1-in-8 crash rate on a build step is a real risk even though the gate it lands on is advisory — the same build runs on the deploy path, where a crash is a failed deploy rather than a red check.

## Key files
- `.github/workflows/image-scan.yml` — the `Built image (web)` job
- `deploy/web/Dockerfile` — the `next build` invocation and its bun pin
- `.github/workflows/ci.yml` — `BUN_VERSION`, cross-checked by the "Dockerfile bun version pins" step in `drift`

## Acceptance criteria
- [ ] Pull the failed job log from run `30475419881`'s predecessor at `0f0ee735c` and record the actual segfault signature in this issue before attempting a fix — right now we have "segfault during `next build`" and not much else
- [ ] Establish a real rate. One occurrence in eight cannot distinguish a 12% flake from a one-off; re-run the web image build on a fixed SHA enough times to bound it
[...]

Issue #4890, opened 2026-07-29
chore(ci): Dependabot's bun sweep structurally cannot pass the drift gate — create-atlas templates are outside the workspace


msywulak:
`.github/dependabot.yml` landed in #4884 and its first grouped sweep is #4886. That PR is red on `drift`, and the cause is structural rather than incidental to this particular batch: **every future weekly minor/patch sweep will fail the same way.**

## Why

The `drift` job runs `bun x syncpack lint` (`.github/workflows/ci.yml:75`). `.syncpackrc.json` includes `create-atlas/templates/*/package.json` in its `source` list and enforces one version per dependency across all listed manifests.

But those templates are **not** root workspace members. Root `workspaces` is `packages/*`, `apps/*`, `examples/nextjs-standalone`, `examples/embedded-mcp-onboarding`, `plugins/*`, `create-atlas-plugin`, `ee` — no `create-atlas`. `create-atlas` is its own install root with its own `create-atlas/bun.lock`, and the two template directories have no lockfile at all.

So Dependabot's `directory: "/"` bumps the root manifest and its workspace members, cannot reach the templates, and syncpack then reports the gap. On #4886:

```
3x zod          ✘ ^3.8.1  → ^3.10.1   in create-atlas/templates/{docker,nextjs-standalone}
3x stripe       ✘ ^22.2.3 → ^22.3.2   in create-atlas/templates/{docker,nextjs-standalone}
8x tailwindcss  ✘ ^4.3.1  → ^4.3.3    in create-atlas/templates/{docker,nextjs-standalone}
```

`drift` is a **required status check**, so this genuinely blocks the merge button.

## Why this converges with the /deps-update overlap

Flagged in #4888 but never given its own tracker: `/deps-update` already owns "Group A = all minors/patches in one PR," and this config now does the same thing automatically every Monday. Double-covered.

Those are the same decision, because the reason `/deps-update` does not hit this gate is that its script is `"deps:update": "bun update && bun x syncpack fix"` — the `syncpack fix` pass is what syncs the templates. **Dependabot has no way to run that.** So the overlap is not merely redundant; the automated path is the one that structurally cannot satisfy the gate, while the manual path already handles it by construction.

## Options

1. **Drop bun version updates; keep `/deps-update` as the sweep.** Resolves overlap and drift with one decision. Loses unattended freshness and the weekly forcing function.
[...]

Issue #4889, opened 2026-07-29
chore(deps): bump @daytonaio/sdk to clear protobufjs@8.0.1 — the OTel 0.221 bump does not


msywulak:
Follow-up from #4878. The residual npm advisory set left after #4884/#4885 included protobufjs, recorded there as unfixable because bun overrides are top-level only and the tree holds **two majors** (7.x + 8.x), so one override cannot pin both consumers.

That reasoning is correct, but the remediation lever was misidentified. The working assumption going into this was *"bump root OpenTelemetry 0.218 → 0.221, because `otlp-transformer@0.221.0` drops protobufjs outright."* **That is not the mechanism.** Measured against `bun.lock` on `1d36a9b8b`:

## Our OTel path never had protobufjs at 0.218

From the npm registry, `@opentelemetry/otlp-transformer` dependency sets:

| version | protobufjs |
|---|---|
| `0.217.0` | `8.0.1` (exact) |
| `0.218.0` | **none** |
| `0.221.0` | **none** |

`0.218.0` already dropped it. Our root/`packages/api`/`packages/web` pins are `^0.218.0`, so **the 0.218 → 0.221 bump clears zero protobufjs findings.** It is a freshness bump, not a security one.

## Where the two copies actually come from

```
protobufjs@8.0.1  ←  @opentelemetry/otlp-transformer@0.217.0  (nested, pinned EXACTLY "8.0.1")
                  ←  @daytonaio/sdk@0.185.0
protobufjs@7.6.2  ←  @grpc/proto-loader@0.8.1  ("^7.5.5")
                  ←  @grpc/grpc-js@1.14.4
```

`@daytonaio/sdk` carries its own nested OTel subtree at `0.217.0` — the last version that pins protobufjs. Ten nested `@daytonaio/sdk/**/otlp-transformer` entries in `bun.lock` all resolve to `0.217.0`. Bumping *our* OTel does not move a nested pin inside another package's subtree.

## The actual lever

`@daytonaio/sdk` latest is **0.202.0** (we pin `^0.185.0`, resolved `0.185.0`). At 0.202.0 its OTel pins move to `@opentelemetry/otlp-exporter-base: 0.220.0` / `sdk-node: ^0.220.0` — both past the 0.218 cutoff, i.e. **no protobufjs**. So bumping `@daytonaio/sdk` is what removes `protobufjs@8.0.1`.

[...]

Issue #4828, opened 2026-07-26
getExploreBackendType() reports "just-bash" for a fail-closed sandbox pin, where explore actually refuses to run


msywulak:
`getExploreBackendType()` collapses "no backend available" into `"just-bash"` unconditionally:

```ts
// packages/api/src/lib/tools/explore.ts
return firstAvailableBackend(plan, isBackendAvailable) ?? "just-bash";
```

`firstAvailableBackend()` returns `null` on exhaustion **regardless of `plan.onExhausted`** — it never consults the fail-closed flag that `planSandboxSelection()` computed.

## Why it matters

`deploy/api/atlas.config.ts` pins `priority: ["vercel-sandbox"]` with no `just-bash` in the list, so `onExhausted` is `"fail-closed"` on staging and all three prod regions. Drop or typo `VERCEL_TOKEN` on one regional service — a per-service Railway secret that shared vars do not inherit — and:

- `/api/health` reports `sandbox.backend: "just-bash"`, `explore.isolated: false`, `status: "degraded"`.
- The boot pre-flight logs `Explore tool: just-bash (no process isolation). Install nsjail or configure ATLAS_SANDBOX_URL…`.

Both are false, in opposite directions from #4824. The deployment is **not** unsandboxed — `runSandboxPlan` yields `kind: "fail-closed"` and explore **throws on every request**. And the remediation is unactionable: the priority pin excludes both `nsjail` and `sidecar`, so following the advice changes nothing. The real fix is the missing credential.

A totally broken explore tool presents as a healthy-but-unsandboxed deploy.

## Why it is not part of #4824

#4824 made the boot log agree with `/api/health`. It does — both now read the same resolver. This is a defect *in that shared resolver*, so fixing it only at the boot log would reintroduce the boot/health divergence #4824 exists to prevent. It needs to change the resolver, which means changing the reported value's type.

## Key files

- `packages/api/src/lib/tools/explore.ts` — `getExploreBackendType()`, the `?? "just-bash"` fallback
- `packages/api/src/lib/tools/backends/selection.ts` — `firstAvailableBackend()` ignores `plan.onExhausted`; `planSandboxSelection()` sets `"fail-closed"` when `just-bash` is absent from the pin
- `packages/api/src/api/routes/health.ts` — the `backend` zod enum + `isolated` boolean that would need a fail-closed member
- `packages/api/src/lib/tools/python.ts` — same planner, same blind spot
[...]

PR #4893, opened 2026-07-29
chore(deps): Group A sweep — server-side only, 35 packages

msywulak:
**Re-scoped.** This started as a 62-package sweep that went red on both image builds. It is now **server-side only (35 packages)**, with the web/docs-facing bumps split out so the majority can land while those are investigated. History of the earlier attempt is in the comments below.

First real Group A sweep since #3952 — everything between was targeted security or pin work. Dependabot's attempt at the same thing (#4886) is closed; its config could not pass the `drift` gate (#4890 → #4892).

## Swept

The `@ai-sdk` provider set, `@aws-sdk/client-bedrock`, `@clack/prompts`, `@clickhouse/client`, `@duckdb/node-api`, the `@effect` stack, `@modelcontextprotocol/sdk`, `ai`, `domhandler`, `effect`, `entities`, `hono`, `htmlparser2`, `jsforce`, `mysql2`, `nanoid`, `nodemailer`, the `oxlint` set, `snowflake-sdk`, `typescript`, `@types/node`.

Plus the two sandbox backends that are `optional`/`peer` and therefore filtered out of `bun outdated` — they need naming explicitly or they ride in a gap neither group covers: **`@vercel/sandbox` 2.2.1 → 2.9.0** and **`e2b` 2.28.2 → 2.36.1**. These are the items worth the closest review: 7 minors on the SaaS prod sandbox (`networkPolicy: "deny-all"`). Both exercised against the explore/sandbox and fail-closed suites.

**No new overrides.** Only the `hono` override moves, as a syncpack cascade of the `hono` bump itself.

## One hold inside the swept set: `jose` pinned at 6.2.3

`@vercel/sandbox` **exact-pins** `jose: "6.2.3"`, so moving top-level jose to 6.2.4 guarantees two copies. That peer-differentiates `@better-auth/core@1.6.25` into separate store paths and clashes better-call's `HookEndpointContext` *nominally* — structurally identical types reported incompatible because their paths differ.

The copies are a cross-product of two independent splits:

| store hash | jose | better-call |
|---|---|---|
| `1b9bf574` | 6.2.3 | 1.3.7 |
| `733cfafa` | 6.2.4 | 1.3.2 |
| `b5dad092` | 6.2.3 | 1.3.2 |
| `ce626f80` | 6.2.4 | 1.3.7 |

The better-call 1.3.2/1.3.7 split is **pre-existing on `main`** (`@better-auth/agent-auth` pins 1.3.2 against a top-level 1.3.7) and main's type gate passes with 2 copies. Adding jose's split makes 4, and it fails. Holding jose keeps both the sandbox bump and a green type gate.

## Deferred, each for a demonstrated reason

| held | why |
|---|---|
| all web/docs-facing (`react`, `react-dom`, `tailwindcss`+`@tailwindcss/*`, `lucide-react`, `recharts`, `radix-ui`+`@radix-ui/*`, `@tanstack/*`, `nuqs`, `react-grid-layout`, `react-hook-form`, `react-resizable-panels`, `shadcn`, `@ai-sdk/react`, `@hookform/resolvers`, `happy-dom`) | `Built image (web)` SIGILLs on the GitHub runner with these in — **but not because of them**; see below |
| `@shikijs/transformers` | 4.3.1 needs shiki-core 4.3.x; shiki is transitive via the held `fumadocs-core` (4.2.0), so `CodeToHastOptions` stops matching and `apps/docs` fails to build. A consequence of holding fumadocs |
| `@hono/zod-openapi` | 1.5.1 breaks 10 request-body validation tests incl. SSRF-guard 400s → #4894 |
| `just-bash` | 3.2.0 breaks 4 explore/sandbox fail-closed tests (#4828) → #4895 |
| `@opentelemetry/*` | excluded, which removes the `semantic-conventions` override the previous attempt needed. Freshness only, no advisory value (#4889) |
| `@playwright/test` **and** `@axe-core/playwright` | both sides of the `playwright-core` peer split — holding one does not prevent it |
| `fumadocs-*` | the 16.10 core/ui ↔ openapi 11 lockstep |
| `stripe` | minor 22.2.3 → 22.3.2 moves the pinned `apiVersion` **type** |
| `@typescript/native-preview` | pinned exactly by #4882 on purpose; excluded so a sweep cannot re-float the type gate |
| `@useatlas/types`, `chat`, `@chat-adapter/*` | published refs / contract-pinned |
| `@daytonaio/sdk` | no within-range update (caret on `0.x` pins the minor) → #4889 |

## The web image SIGILL is the runner, not the deps

Worth stating clearly because I got this wrong twice in the comments below. The **identical full-sweep tree builds clean in Docker locally** — exit 0, with `lucide-react@1.27.0` / `react@19.2.8` / `tailwindcss@4.3.3` confirmed in the build log. And `bun run build` passes both locally and in CI's non-Docker `build` job.

| where | result |
|---|---|
| `bun run build` locally | pass |
| `build` job, CI runner, no Docker | pass |
| `deploy/web/Dockerfile` in Docker **locally** | **pass** (exit 0, full sweep) |
| `Built image (web)` on the GitHub runner | fail ×2, SIGILL/132 |

So it is #4891, runner-specific. What misled me was that `main` passes while the branch failed twice — but a heavier build simply spends longer in the window where the bad codegen path is reachable (CI compiles in 51s vs 13.8s locally). Environmental *trigger*, sweep-sensitive *exposure*. Holding the web-facing bumps keeps that variable out of this PR rather than claiming to fix it.

## Verification

| gate | result |
|---|---|
| `bun run type` | pass |
| `bun run lint` | pass, **0 errors** |
| `bun x syncpack lint` | `✓ No issues found` |
| `bun install --frozen-lockfile` | no changes |
| `bun run test:api` | **990/990 files** |
| `bun run test:others` | **38/38 packages** |
| `bun run build:docs` | pass |
| `bun run build` (web) | pass |
| standalone Turbopack build | pass |

The docs and web builds are **not** in the deps-update skill's gate list, and `bun run type` does not cover `apps/docs` at all — which is exactly how the shiki break reached CI on the previous attempt. Both belong in that list.

Note `test-isolated.ts --affected` reports no affected tests: all changed files are manifests, so nothing imports them. A sweep is precisely where `--affected` proves nothing.


msywulak:
Follow-ups filed for the two holds: #4894 (@hono/zod-openapi 1.5.x validation change) and #4895 (just-bash 3.2.x backend selection). Both flagged security-relevant, since the failing assertions are SSRF guards and fail-closed sandbox selection respectively — the direction of each behaviour change needs establishing before either bump lands.

msywulak:
## The one red check is the known flake, not this sweep

`Built image (docs)` failed with:

```
error: script "build" was terminated by signal SIGILL (Illegal instruction)
process "cd apps/docs && bun run postinstall && bun run build && check-docs-canary.sh" exit code: 132
```

That is #4891 — bun 1.3.13 crashing mid-Next-build inside the image. Attribution evidence, since "a 62-package dep sweep broke the docs build" is the obvious reading and it is wrong:

- **#4886 failed this identical job** with a completely different dependency set (Dependabot's 66-package sweep)
- The original occurrence was `Built image (web)` on plain `main` at `0f0ee735c`, no sweep at all
- `fumadocs-*` is held entirely in this PR, so the docs toolchain is unchanged from `main`
- `apps/docs` builds fine locally on this branch, and `Standalone Example Build` (the other in-image Next build) passed here

Three occurrences, two different images, three different dependency sets → the common factor is bun, not the deps. I have posted the widened scope on #4891, including that it is SIGILL/132 rather than the segfault that issue currently claims, which points at codegen rather than memory.

**All required checks are green:** `ci`, `api-tests (1-4/4)`, `Analyze (javascript-typescript)`, `Deploy Validation`, `Symlink Stub Build`, `fork-pr-gate`. Plus `type`, `lint`, `lint-type-aware`, `drift`, `test-others`, `build`, `Boot Build`/`Boot Smoke`, both `Scaffold` jobs, `Standalone Example Build`, `CodeQL`.

I will re-run the failed job once the workflow finishes (individual jobs cannot be re-run mid-run) to confirm it is transient. **Worth being deliberate about the merge regardless:** the same `bun run build` runs on the docs deploy path, which goes direct from `main` with no staging twin — so if it crashes on the deploy rather than the scan, docs.useatlas.dev fails to update and nothing gates it.

msywulak:
## Correcting my previous comment — the docs failure was NOT the flake

I attributed `Built image (docs)` to #4891 above. That was wrong, and the reasoning that misled me is worth naming: CI reported `SIGILL`/132 (a crash, matching #4891's signature) and #4886 had failed the identical job with a different dep set. Both true, both insufficient. Building `apps/docs` locally reproduces a real incompatibility:

```
Type 'CodeToHastOptionsCommon<string> & CodeOptionsSingleTheme<string> & CodeOptionsMeta'
  is not assignable to type 'CodeToHastOptions<string, string>'
  79 |  ...(rehypeCodeDefaultOptions.transformers ?? []),
```

**Mechanism — a new coupling, and it is a consequence of holding fumadocs.** The sweep moved `@shikijs/transformers` 4.2.0 → 4.3.1, but `shiki` **core stayed at 4.2.0**, because shiki is transitive via `fumadocs-core` which this PR pins at 16.9.3 for the 16.10 lockstep. `CodeToHastOptions` comes from shiki core, so 4.3.1 transformers against 4.2.0 core does not typecheck. Holding fumadocs pins shiki, which transitively pins `@shikijs/transformers` — held at 4.2.0 to match, and it can move when the fumadocs 16.10 / openapi-11 migration moves shiki.

Presumably CI surfaced it as SIGILL rather than a clean type error because the failing type-check path crashes bun on the Azure runner. That is a nasty interaction: **a real type error masquerading as #4891's crash**, which is exactly how I got it wrong. Anyone triaging a SIGILL in an image build should reproduce the build locally before writing it off as the known flake.

**Why the gate battery missed it:** `bun run type` does not cover `apps/docs` — the script is `tsgo --noEmit` plus `@atlas/web` and `@atlas/www` type. The docs build type-checks itself, and neither `bun run build:docs` nor `bun run build` is in the deps-update gate list (it has only the standalone example build). Both belong there.

## `Built image (web)` is a separate matter

`packages/web` **builds clean locally** with the swept deps (exit 0), so its `SIGILL`/132 is the #4891 environmental crash rather than this sweep. Two red image jobs, two different causes. The push above will re-run both — if web SIGILLs again on an unchanged web toolchain, #4891 is more systematic than 1-in-9 and deserves priority.

msywulak:
## Re-scoped to server-side only — and my web diagnosis was wrong twice, now settled

Force-pushed `595cae094`. This PR is now **35 server-side packages**; the web/docs-facing bumps moved to #4896.

### The web SIGILL is the runner, not the deps

I called it environmental, then retracted that, and the retraction was wrong. Settled by building the image locally:

```
✓ Compiled successfully in 13.8s
EXIT=0
```

Verified as the right tree by the versions in the build log — `lucide-react@1.27.0`, `react@19.2.8`, `tailwindcss@4.3.3`. Same `deploy/web/Dockerfile`, same lockfile.

| where | result |
|---|---|
| `bun run build` locally | pass |
| `build` job, CI runner, no container | pass |
| `deploy/web/Dockerfile` in Docker **locally** | **pass** |
| `Built image (web)` on the GitHub runner | fail ×2 |

So: bun in a container **on the GitHub-hosted runner**. What misled me into retracting was that `main` passes while this branch failed twice — reading two-for-two as determinism. But the build compiles fine and dies ~55s later in post-compile, and a heavier frontend spends longer in that window (51s to compile on the runner vs 13.8s locally). Environmental trigger, dep-sensitive exposure. Details on #4891, priority raised there.

### One hold inside the swept set

`jose` pinned at 6.2.3. `@vercel/sandbox` **exact-pins** `jose: "6.2.3"`, so a top-level move to 6.2.4 guarantees two copies, which peer-differentiates `@better-auth/core` and clashes better-call's `HookEndpointContext` nominally. The copies are a cross-product of two splits — the better-call 1.3.2/1.3.7 one is pre-existing on `main` (2 copies, type gate green); adding jose's makes 4 and it fails. Holding jose keeps the sandbox bump *and* a green type gate.

Excluding `@opentelemetry/*` also removed the need for the `semantic-conventions` override the previous attempt required, so **this PR adds no new overrides** — only the `hono` override moves, as a syncpack cascade of the hono bump.

### Gate list gap this exposed

`bun run build:docs` and `bun run build` are **not** in the `/deps-update` gate list, and `bun run type` does not cover `apps/docs` at all. That is precisely how the shiki break reached CI on the previous attempt. Both ran clean here, along with 990/990 api test files and 38/38 `test:others`.

PR #4908, opened 2026-07-30
fix(chat): persist botUserId so Slack self-messages are filtered (#4907)

### LINKED ISSUE #4907: fix(chat): Slack self-message loop — botUserId never persisted, disabling the adapter's self-filter
Atlas answered its own Slack messages in an unbounded loop in **prod** on 2026-07-30, across two channels. ~15 agent turns in ~90 seconds, one a 144k-token cross-region query. Contained by disabling the Slack integration in `/admin` and turning off Event Subscriptions in the Slack app.

## Root cause

`saveInstallation` never persisted `botUserId`, which silently disables the chat-adapter's entire self-message filter in multi-workspace mode.

1. Slack delivers Atlas's own post as a `message` event in a subscribed thread.
2. `@chat-adapter/slack` takes the multi-workspace branch (`if (!this.defaultBotToken)`) → `resolveTokenForTeam` returns `{ token, botUserId: installation.botUserId }` — and that field is `undefined`.
3. `isMessageFromSelf(event)` has exactly three checks and **all three are dead**:
   - `ctx?.botUserId === event.user` → `ctx.botUserId` undefined
   - `this._botUserId` → null; multi-workspace deploys omit `botToken` by design (`adapter-registry.ts:153-155`)
   - `this._botId === event.bot_id` → null, same reason
4. `author.isMe` is false → the SDK core's `Skipping message from self` guard never fires.
5. `dispatchToHandlers` → thread is subscribed → `onSubscribedMessage` → `bridge.ts` logs `Follow-up message received` → agent runs → posts → **goto 1**.

Log signature: `agent finished` followed ~1s later by `Follow-up message received`, on repeat.

`lib/slack/store.ts`'s own header documented the gap — it said Atlas's writer never sets `botUserId` and deferred it to "a future `auth.test` round-trip" by the adapter. That round-trip does not exist in v4.23.0; the adapter only ever *reads* the field. Its sole writer is the SDK's `handleOAuthCallback`, which Atlas bypasses by running `SlackOAuthInstallHandler`. So nothing wrote it.

## Why it fired now

Two independent conditions, both required. The broken filter was latent since multi-workspace OAuth landed — Slack simply never delivered Atlas's own channel posts, because the app only subscribed to `app_mention`. The history-scoped reinstall for Company Brain added `message.channels` / `message.groups` subscriptions, which started delivering them into a filter that could never match.

Rolling back the subscriptions would re-hide the bug, not fix it.

## Key files

- `packages/api/src/lib/slack/store.ts` — `saveInstallation` / `SlackSaveInput` / the upsert
- `packages/api/src/lib/integrations/install/slack-oauth-handler.ts` — extracts `bot_user_id` at :251, didn't pass it on
- `plugins/chat/src/bridge.ts` — `onSubscribedMessage`, the reply circuit breaker
- `docs/architecture/chat-plugin-atlas-contract.md` — boundary table

## Acceptance criteria

- [ ] `botUserId` persisted to `chat_cache` on install, sourced from `oauth.v2.access`
- [ ] Absent `botUserId` omits the JSONB key (so the merge preserves a stored id) rather than writing null
- [ ] Missing `bot_user_id` from Slack logs a warning; install still succeeds
- [ ] Per-thread reply circuit breaker bounds any future loop regardless of cause
- [ ] Breaker fails open when the state backend is down
- [ ] Tests assert the negative — they fail without each fix
- [ ] Contract doc updated (boundary change)

## Follow-ups (not this PR)

- Existing prod installs still have no `botUserId` — they need a reconnect, or a backfill via `auth.test`.
- Decide whether `message.channels` / `message.groups` subscriptions are wanted at all: brain ingest is pull-based (`conversations.history`) and needs the *scopes*, not the push events.

msywulak:
Fixes #4907.

Atlas answered its own Slack posts in an unbounded loop in **prod** today — ~15 agent turns in 90 seconds across two channels, one of them a 144k-token cross-region query. Contained by disabling the integration in `/admin` + Event Subscriptions in Slack.

## Root cause

`saveInstallation` never persisted `botUserId`, which silently disables the chat-adapter's **entire** self-message filter in multi-workspace mode.

`isMessageFromSelf` has exactly three checks, and in SaaS all three are dead:

| Check | Why it fails |
|---|---|
| `ctx?.botUserId === event.user` | `ctx.botUserId` undefined — sourced from the chat_cache row we never wrote |
| `this._botUserId === event.user` | null — only populates from a single-workspace bot token, which SaaS omits by design |
| `this._botId === event.bot_id` | null — same reason |

So `author.isMe` is false for every self-post, the SDK core's `Skipping message from self` guard never fires, and each reply re-enters `onSubscribedMessage` → agent runs → posts → repeat. Log signature: `agent finished` followed ~1s later by `Follow-up message received`, on repeat.

`lib/slack/store.ts`'s own header documented the gap — it said Atlas's writer never sets `botUserId` and deferred it to "a future `auth.test` round-trip" by the adapter. That round-trip doesn't exist: the adapter only *reads* the field, and its sole writer is the SDK's `handleOAuthCallback`, which Atlas bypasses by running `SlackOAuthInstallHandler`.

## Why it fired now

Latent since multi-workspace OAuth landed. Slack simply never delivered the bot's own channel posts, because the app only subscribed to `app_mention`. The Company Brain history reinstall added `message.channels` / `message.groups` subscriptions, which started delivering them into a filter that could never match. **Rolling back those subscriptions would re-hide the bug, not fix it.**

## Changes

**Both install paths now persist `botUserId`** — I nearly shipped only the OAuth half:
- OAuth (`slack-oauth-handler.ts`) — from `oauth.v2.access`'s `bot_user_id`, which was already extracted and stored in `workspace_plugins.config` but never passed to the store the adapter actually reads
- BYOT (`admin-integrations.ts`) — from `auth.test`'s `user_id`, which the route was already fetching and discarding

Absent values omit the JSONB key rather than writing null, so the upsert's `value || EXCLUDED.value` merge preserves a stored id. A missing id warns at install rather than failing it — Slack sends the field on every bot install, so failing an otherwise-valid exchange over it would trade a real outage for a hypothetical one.

**Reply circuit breaker** (`bridge.ts`) — 12 replies/minute per thread, 15-minute cooldown, one notice then silence. Cause-agnostic backstop so the next trigger can't run unbounded. The existing dedup lock could never serve this role: it releases in `finally`, so a *sequential* loop reacquires it cleanly every time. Fails open when the state backend is down — it bounds a runaway, it isn't a security control.

**Contract doc** — the `botUserId` row said "adapter-only (never written by Atlas)", documenting the bug as intended behavior. Also added row 4 to the "How we got here" table, since this is a fourth instance of the exact class that doc exists to prevent, and named the generalisation: when Atlas takes over one of the adapter's write paths, it inherits every field that path wrote, not just the ones Atlas reads.

## Verification

- `bun run type` clean, `bun run lint` exit 0 with no new findings
- 41/41 affected API test files pass; `plugins/chat` 132 pass
- **Each of the four fixes was individually reverted and the corresponding test confirmed to fail.** Green alone would have proved nothing here — the first draft of the breaker tests passed vacuously because `MemoryStateAdapter` throws until `connect()` and the fail-open path swallowed it.

## Not in this PR

- **Existing prod installs still have no `botUserId`.** They need a reconnect (or a backfill via `auth.test`) before Slack is safe to re-enable. This is the deploy-order dependency, not an optional follow-up.
- Whether `message.channels` / `message.groups` subscriptions are wanted at all — brain ingest is pull-based and needs the *scopes*, not the push events.


msywulak:
## Review panel — changes applied (b2b53543e)

Four specialist reviewers ran the diff. Two findings were the same class this PR exists to close, so both are fixed here rather than deferred.

**The DM path had no breaker, and my comment claimed it couldn't loop.** The SDK's `dispatchToHandlers` checks DMs *before* subscription and returns early, and `onDirectMessage` shares `handleMentionOrDM` — so a self-post in a 1:1 DM never reaches `onSubscribedMessage`, needs no @-mention, and hit no guard at all. `bridge.ts`'s own comment already documented that dispatch order; I wrote a contradicting claim 180 lines below it. Both handlers now go through a shared `breakerShouldHalt`.

**Existing installs are now repaired.** The original fix only took effect on the *next* install, so every workspace installed before it — including the one from the incident — kept looping. Migration `0185` backfills `botUserId` from `workspace_plugins.config`, where the correct value has been sitting all along. This removes the deploy-order dependency I'd listed as a follow-up.

**The breaker now triggers on `author.isBot`, not rate alone.** Every event in the loop carries `bot_id`, so counting only bot-authored messages separates a loop from human traffic by *kind* — a busy thread can't be muted at any typing speed. This also retires a bad argument: my threshold comment cited "~15 turns in 90s" (= 10/min) to justify a 12/min trip, which **would not have fired**. Recounting the prod logs gives ~26 turns in 90s (~17/min). The rate is now a backstop, not the discriminator.

Also applied: validated breaker state on read (an unchecked cast made `count` undefined → `NaN` → never trips, never logs — the same silent-disable class as #4907, inside the backstop); narrowed the catch so a failed latch can't swallow the trip or its log line; BYOT warn on missing `user_id`; `opts?: Omit<SlackSaveInput, "botToken">` to kill the options-bag drift; pinned `chat_cache.value || EXCLUDED.value` in a test; window-decay and cooldown-expiry coverage via `setSystemTime`; corrected contract-doc rows 42/44 and the stale premise in `migrate-pg.test.ts`.

**Two test-quality notes worth recording.** Two of the original breaker tests passed against a *completely dead* breaker — `MemoryStateAdapter` throws until `connect()` and the fail-open path swallowed it. Every breaker test now asserts a `"muted"`/`"trip"` transition rather than only the absence of one, and the threshold is pinned to the exact trip index. Separately, my earlier claim that all four fixes were revert-verified was wrong: three were, the breaker *wiring* wasn't.

**Residual, stated plainly:** the breaker's call-site placement is still not pinned by a test. `breakerShouldHalt` is covered directly, but the bridge doesn't expose its `Chat` instance and mocking the `chat` module risks the partial-mock hazards this repo has hit before. Deleting the call from a handler would not fail the suite — there's a comment on the export saying so.

Verification: type clean, lint exit 0, 43/43 affected API test files, 138 in `plugins/chat`, and the `isBot` gate plus the malformed-state guard were each individually reverted and confirmed to fail.

Issue #4926, opened 2026-07-31
chore(brain): harden wedge-loop-pg fixture helpers — 4 diagnostic-destroying defects


msywulak:
Four defects in `wedge-loop-pg.test.ts`'s fixture helpers, found by the review panel on #4917 (PR #4925). All four were inherited verbatim into `temporal-loop-pg.test.ts` and **fixed there** — so that file is the reference implementation for every fix below, and the job is largely a copy-across.

**None of these can make a passing test fail or a failing test pass.** They are diagnostic-quality defects: they only bite once something else is already broken, and then they destroy or delay the signal that would tell you what. That is why this is a chore and not a bug — but it is also why they are worth clearing before anyone debugs this suite under pressure.

Do this **before #4918** (docs) so M2 closes on a clean test surface.

## The four

1. **`:724` — poisoned client returned to the pool.** `poolTransaction`'s `finally { client.release() }` hands a client with an in-doubt transaction back to the pool when the callback throws *and* the `ROLLBACK` also fails. The held locks then block `afterEach`'s `TRUNCATE`, so the run reports `timed out after 60000ms` instead of the reconcile error that actually happened. Requires a double failure to reach, but it is the one that costs real debugging time. `publish()` at `:577` already documents and guards this exact hazard (`client.release(err)` — destroy, don't recycle); `poolTransaction` does the opposite.

2. **`:720` — un-narrowed caught error.** `console.debug("fixture rollback failed", rbErr)` hands a raw `unknown` to the log and names no operation, so you cannot tell which transaction died. Violates CLAUDE.md § Error handling (type-narrow caught errors) — test-only, but the same file's `publish()` gets it right, so this is drift rather than convention.

3. **`:402` — `afterAll` leaks the pool if `DROP SCHEMA` throws.** `pool.end()` never runs, so open handles keep the bun process alive *and* the scratch `brain_*` schema leaks. Reachable via a lingering lock (see 1) or a statement timeout — `DROP SCHEMA … CASCADE` over a fully-migrated schema is slow. `beforeAll` already does this correctly for the bootstrap pool with `try/finally`.

[...]

Issue #4918, opened 2026-07-30
docs(brain): temporal model, correction verbs, conflict semantics


msywulak:
User-facing docs for what M2 ships — the trust-depth story is the wedge, so it must be legible: the bi-temporal model (as-of-now default, `asOf` point reads, supersession vs retraction, "invalidate never delete", tombstone-hides-history), the four correction verbs (retract as the GDPR-erasure verb), conflict semantics (surfaced-both-with-provenance, Atlas refuses to auto-arbitrate — this is a feature, say why), and the staleness stance (advisory, never auto-demoted).

Follow the docs portal's audience-segmentation invariants (self-hosted vs SaaS pages don't leak each other's assumptions). No ADR amendment is needed — T4 is fully specified in ADR-0036; these are docs-site pages.

## Key files
- `apps/docs/` — brain/knowledge section (extend the M1 searchBrain/review-gate pages)

## Acceptance criteria
- [ ] Temporal model page: supersession vs retraction distinction, `asOf` examples, frozen-grant behavior stated
- [ ] Correction verbs documented incl. GDPR-erasure positioning of retract
- [ ] Conflict page explains surfaced-both and why auto-arbitration is refused
- [ ] Claims cross-checked against shipped behavior (no aspirational docs); `/audit-docs` conventions
- [ ] Trust-tier labeling terminology consistent with ADR-0036 ("sole authority for what is queryable" — never the unqualified moat phrasing)

## Dependencies
- Depends on #4912–#4916 (document shipped behavior, not plans)

msywulak:
Shipped in #4931, merged to `milestone/brain-m2` as `153d8c634`.

Three pages in `apps/docs/content/shared/guides/` — `brain-temporal-model`, `brain-corrections`, `brain-conflicts` — mounted into both audience trees. All five acceptance criteria met; see the PR for the verification detail (claims traced to the M2 source, segmentation checked in the emitted HTML and markdown twins).

Closing by hand: `Closes` doesn't fire on a non-default base branch.

Issue #4917, opened 2026-07-30
chore(brain): e2e temporal-loop proof (conflict → correct → supersede → asOf)


msywulak:
The M2 counterpart of #4775's wedge-loop proof: one `-pg` e2e over the completed temporal stack. Two conflicting `single`-cardinality claims from different episodes → advisory `in-tension-with` at reconcile → both surfaced with provenance to reader and reviewer → reviewer publishes one, superseding at the gate (`valid_to` + `supersedes` edge) → default read returns only the survivor → `asOf` before the publish returns yesterday's truth under the frozen grant → `correct_fact` retract tombstones and hides history from every read including `asOf` → the retracted fact's `derives-from` dependents are flagged, not cascaded.

Inherited fixture discipline from #4775: build reader contexts with `resolvePrincipalContext` (a hand-built context makes every ACL assertion a statement about the fixture); seed the `workspace_plugins` install row or `upsertConnectorSyncState` silently no-ops; inject `now` (the 7-day backfill floor excludes real-clock fixtures); `GRANT_UNUSABLE` remains the only gate refusal reachable from the database.

## Key files
- `packages/api/src/lib/brain/__tests__/temporal-loop-pg.test.ts` (new) — modeled on `wedge-loop-pg.test.ts`
- `packages/api/src/lib/brain/__tests__/acl-visibility-pg.test.ts` — the asOf-with-changed-grant case may belong here per the M1 header-scoping notes

## Acceptance criteria
- [ ] The full loop above passes as one `-pg` test file under the isolated runner
- [ ] Negative arms included (test-the-negative): the loser is absent from default reads, present under `asOf`; the tombstoned fact absent from BOTH; the withheld-counterpart arm exercised with a reader who cannot see one side
- [ ] Fixture discipline rules above followed (documented in the file header)

## Dependencies
- Depends on #4912, #4913, #4914, #4915, #4916

msywulak:
Shipped in #4925 (`7a47a537`), merged into `milestone/brain-m2`.

[...]

Issue #4916, opened 2026-07-30
feat(brain): bi-temporal asOf reads in searchBrain


msywulak:
T4/T7's bi-temporal read: `searchBrain` gains an optional `asOf` parameter — a point read returning the facts valid at that instant (`valid_from <= asOf < COALESCE(valid_to, ∞)`), each gated by **that version's frozen ACL grant snapshot** (the row's own immutable `visible_to`) evaluated against **as-of-now audience membership** (T5's bi-temporal ACL). "What did we believe Monday" uses Monday's grant.

Two boundary rules from the ADR: **tombstoned facts stay hidden even under `asOf`** — retraction is the only way to hide history, and hiding is what it does; superseded facts DO appear inside their validity window (that is the point). **Default reads are completely unchanged** — as-of-now, superseded + tombstoned hidden — and regression-pinned. Historical results stay trust-labeled; malformed `asOf` is rejected with an actionable error, never silently ignored (fail closed, not fall through to as-of-now).

## Key files
- `packages/api/src/lib/brain/search.ts` — `searchBrainCore` temporal predicate branch
- `packages/api/src/lib/brain/acl.ts` — per-version grant evaluation (should already hold: the predicate reads the row's `visible_to`)
- `packages/api/src/lib/brain/reader-context.ts` — as-of-now membership resolution (unchanged, but the seam this rides)
- `packages/api/src/lib/tools/search-brain.ts` — tool param + description
- MCP `searchBrain` surface (`packages/api/src/lib/mcp/`) — param exposure, additive
- `packages/schemas/src/brain.ts` — `asOf` in request/response shapes

## Acceptance criteria
- [ ] `asOf` returns facts valid at that instant under each version's frozen `visible_to` + as-of-now membership — proven by a `-pg` test where a grant changed between versions
- [ ] Tombstoned (`invalidated_at`) facts hidden under any `asOf`; superseded facts visible inside their window
- [ ] Default (no `asOf`) behavior byte-identical to today — regression-pinned
- [ ] Agent + MCP schemas expose `asOf` additively; results remain trust-tier + provenance labeled
- [ ] Malformed/future-bounds `asOf` rejected with a context-specific error

## Dependencies
- Depends on #4912 (supersession is what creates history worth reading)

msywulak:
[...]

PR #4932, opened 2026-07-31
Brain M2: Temporal & Conflict Depth — milestone integration (take 2, with un-revert)

msywulak:
## Summary

Lands milestone #92 (**Brain M2 — Temporal & Conflict Depth**, ADR-0036 T4) on `main`. All 8 issues are complete: #4912–#4917 plus #4918's docs.

This is the second attempt. #4928 merged M2 prematurely and #4930 reverted it. **That revert is why this PR is shaped the way it is.**

## Why this branch is a revert + a merge

Reverting a merge commit makes git record those changes as intentionally undone, so a plain `git merge milestone/brain-m2` would have brought back **nothing** — the merge succeeds and the code simply isn't there. The failure is silent, which is what makes it dangerous.

So this branch is, in order:

1. `ca9ed6b81` — `git revert 0bb5d9130`, the un-revert of #4930
2. `86944792e` — merge of `milestone/brain-m2` (which contributed only the #4918 docs, since the un-revert already restored the code)

## Verification that it actually landed

The whole point being that the failure mode is invisible, the resulting tree was checked directly rather than inferred from a green merge:

| Check | Result |
|---|---|
| `git diff origin/milestone/brain-m2 HEAD` | only `main`'s own blog post (`apps/www`) — nothing from M2 missing |
| `correction.ts` · `tensions.ts` · `staleness.ts` · `correct-fact.ts` | all present |
| #4926's wedge-loop fixture hardening | present (`inspected` ×5, matching `main`) |
| The three #4918 docs pages | present |

## What M2 ships

- **#4912** — human-gated supersession at the publish gate. No autonomous supersession, ever.
- **#4913** — `in-tension-with` conflict clusters, surfaced-both-with-provenance, never ranked. Counterparts the reader may not see are reported as opaque handles rather than dropped.
- **#4914** — advisory staleness decay. Read-time derivation only: no stored score, no scheduler, no write path. Surfaces, never demotes.
- **#4915** — `correct_fact` and the four verbs (`retract` / `supersede` / `re-authority` / `pin`). `retract` is the only tombstone and the GDPR-erasure route.
- **#4916** — bi-temporal `asOf` point reads. Tombstones stay hidden under any `asOf`; the grant is frozen per version while membership resolves live.
- **#4917** — e2e temporal-loop proof: conflict → correct → supersede → `asOf`.
- **#4918** — three user-facing docs pages, mounted into both audience trees.

## Test Plan

- [x] Local `/ci` on this composed branch — **32/33 gates PASS**
- [x] Un-revert + merge verified against the resulting tree (table above)
- [ ] Manual testing — N/A, integration of already-reviewed slices
- [x] Unit tests — M2's own suites ride along; each slice passed CI on its own PR

⚠️ **One local test failure, believed to be the known WSL2 full-suite artifact:** `agent-compaction.test.ts` failed inside the 994-file full run but passes **46/46 in isolation**, and **none** of its four dependencies (`agent-compaction.ts`, `db/internal`, `gateway-catalog`, `settings.ts`) are touched by this branch. Its sibling `agent-compaction-resume` and `agent-compaction-integration` suites both passed in the same run. **CI here is the authoritative check** — if it goes red on that file, this note is wrong and it needs a real fix.

⚠️ `TEST_DATABASE_URL` was unset locally, so `*-pg.test.ts` were skipped — including M2's own `temporal-loop-pg` and `wedge-loop-pg` proofs. CI runs them, and they are the ones that matter most here.

## Notes

- **CodeQL runs on this PR.** It was structurally absent on every milestone-branch PR (default setup is `main`-only), which is the documented deferral — this is where it comes due.
- Milestone #92 closes on merge. Per the held-minor decision, M2 rides the `v0.2.x` patch train; the next `0.x.0` is the Brain-complete tag.


msywulak:
## Milestone-wide review — six-lens panel over the combined diff

Ran a multi-lens review over the whole arc (security/ACL · silent-failure · type design · test coverage · cross-surface consistency · repo-guideline conformance), briefed to find what per-slice review structurally could not see. Every finding below was re-verified in source before filing.

**One fix applied here** (`966162587`) — guidance only, no behaviour change:

The `searchBrain` system prompt claimed *"Retracted facts never appear, at any time."* #4913 made that false — a retracted fact still appears as a `tensions` counterpart, labelled by `invalidatedAt`. The sentence was written for the `asOf` axis, where it holds, and generalized one axis too far. Rescoped, and the counterpart case is now named. The `brain-temporal-model.mdx` callout carried the same absolute the other two brain pages already qualify.

**Filed as follow-ups** — none blocks this merge; all are additive:

| Issue | What | Severity |
|---|---|---|
| #4935 | Superseded tension counterparts carry no `validTo` — an arbitrated conflict reads as live on every surface | **P1** |
| #4936 | `correct_fact` reaches headless agent surfaces: `runAgent` defaults `tools` to `defaultRegistry`, bypassing the #4915 surface gate | **P1** |
| #4937 | KB `ingest-bundle` publish stamps `valid_to` with no supersession disclosure — third `runPublishPhases` caller | **P1** |
| #4938 | Temporal invariants that pass for the wrong reason — evadable tombstone scan, unfalsifiable retract-`status` assertion | **P1** |
| #4933 | Tool-schema + MCP descriptions omit the retracted-rival label (word-capped, hence separate) | P2 |
| #4934 | `correct_fact` agent-tool path writes no admin-actions audit row | P2 |
| #4939 | Disclosure/doc drift — `content-mode.md` carve-out unrecorded, `/retract` response divergence, inert correction markers | P2 |

#4935 is the one worth closing before M2 is called done: three independent lenses converged on it, `tensions.ts` gives the *identical* argument for carrying both temporal labels and implements only one, and it fires after **every** publish-gate supersession. Note a pre-M2 assertion (`review-honesty.test.tsx:539-540`, "M2 owns arbitration; this surface must not pre-empt it") now blocks the fix — that premise has shipped.

**Verified clean**, worth recording because it is most of the surface: visibility-predicate composition correct at every `brain_facts` SELECT (enumerated, including `<`/`<=` boundary agreement) · no ranking key anywhere · `asOf` a branded instant with one parse site, fail-closed · tombstones hidden in both temporal branches · trust tier required on every result variant · withheld arms genuinely `z.strictObject` on both shapes · `classifyFactForPromotion` enforced on the correction path · the promotion guard **strengthened** (`valid_to` added), not widened · no `lib/` → `api/routes/` imports, no explicit `any`, no new non-null assertions in production code · openapi regenerates with zero drift · no winner-implying styling on the web render · absence of migrations correct (M1 shipped the columns).

msywulak:
Superseded by #4950 — identical tree, moved onto `milestone/brain-m2`.

**Why:** this PR's head was `chore/brain-m2-to-main`, outside `ci.yml`'s `pull_request` filter of `[main, "milestone/**"]`. Every PR stacked onto it therefore ran `fork-pr-gate` and nothing else — five did (#4942/#4944/#4945/#4946/#4947), and three of the four authors independently fired a `workflow_dispatch` to get real coverage. Four manual workarounds for a namespace choice.

`milestone/brain-m2` already existed and is already gated, so the arc was moved there rather than special-casing a branch name in CI. #4948 had added that special case; #4949 removes it again, and #4949's own **21-check run is the proof the move worked** (`api-tests` 1/4–4/4, `type`, `lint`, `drift`, `build`, `test-others`, `test-e2e-integration`) versus `fork-pr-gate` alone here.

**Nothing was lost:** `milestone/brain-m2` was fast-forwarded `153d8c634 → 8e445bbf4`, ancestor-only, 12 commits gained and 0 lost, no force push.

Closing rather than leaving open so there is one integration PR, not two. **The review record here stays and is worth reading** — the six-lens panel comment on this PR is what produced P1s #4935/#4936/#4937/#4938, all now closed, plus P2s #4933/#4934/#4939 which remain open.

Issue #4943, opened 2026-07-31
chore(agent): make runAgent's `tools` a required parameter so the compiler enforces the surface gate


msywulak:
Raised by the #4936 review panel (type-design lens). #4936 closed the vulnerability; this closes the class.

#4936 made `runAgent`'s `tools` default fail closed (`nonDashboardRegistry`) and added `agent-runagent-call-sites.test.ts` to require every production call site to name its registry. `tools` is still **optional**, and the guard is a source-text scanner.

The reviewer proved with `tsc --noEmit` that a REQUIRED `tools` rejects all five bug shapes at compile time, including the two conditional spreads that were the actual pre-fix source:

| shape | required-param result |
|---|---|
| `...(reg ? { tools: reg } : {})` — the old `answer-adapter.ts` | TS2345 |
| `...(reg && { tools: reg })` — the old `chat.ts` | TS2345 |
| `...partialOptionsObject` | TS2345 / TS2379 |
| plain omission — old `resume-turn.ts` / `demo.ts` | TS2741 |
| `tools: maybeUndefinedVar` | TS2322 |

TS distributes the spread over the union, so the `{}` branch surfaces as `tools?: ToolRegistry` and fails the required target. TS 6 and TS 7 agree.

**Why it was deferred, not skipped.** `runAgent(` call sites across 22 test files omit `tools`. The migration is not purely mechanical: #4936 twice found that naming the registry at a call site dragged `lib/tools/registry`'s module graph into a test that previously only mocked `lib/agent`, breaking its partial `mock.module` (`cors.test.ts` needed `invalidateOrgExploreBackends` + `getSettingOverride`; `chat-plugin/__tests__/resume-turn.test.ts` needed `getRequestContext`). Expect that in a handful of the 22.

**What the scanner cannot do, and the compiler can:** aliased imports (`import { runAgent as run }`), untracked files, `.tsx`, any package not in `SCAN_ROOTS`, and — verified during the #4936 round-2 review — `tools: cond ? reg : undefined`, which typechecks today because `exactOptionalPropertyTypes` is off.

**What the scanner does that the compiler cannot:** catch `runAgent(opts)` where `opts: RunAgentInput` compiles clean but launders the choice through a bag. Keep the guard.

**Recommended shape** — required in the type AND fail-closed at runtime, which is strictly stronger than either alone:

```ts
// in the signature: tools: ToolRegistry   (required)
// in the body:      const active = tools ?? nonDashboardRegistry;
```

[...]

Issue #4936, opened 2026-07-31
security(brain): correct_fact reaches headless agent surfaces — runAgent defaults tools to defaultRegistry, bypassing the #4915 surface gate


msywulak:
Found by the milestone-wide review of PR #4932 (security lens), verified in source.

#4915 deliberately keeps `correct_fact` — a **brain-mutating write tool** — off the read-safe agent surface by registering it only when a `dashboardUrlResolver` is present (`lib/tools/registry.ts:308-314`), with a well-argued comment about the #4707 read-safe `POST /api/v1/query` admission. That gate is correct where it is applied. **It is bypassed by a default parameter.**

- `lib/tools/registry.ts:364-366` — `registerCoreTools(defaultRegistry, WORKSPACE_DASHBOARD_URL_RESOLVER)`, so **`defaultRegistry` contains `correct_fact`**.
- `lib/agent.ts:1084` — `runAgent({ tools: toolRegistry = defaultRegistry, … })`. The resolver signal is only consulted by callers that *pass* `tools`.

Two live call sites do not pass it:

1. **`ee/src/proactive/answer-adapter.ts:306`** — `...(toolRegistry ? { tools: toolRegistry } : {})`. `toolRegistry` (declared `:182`) is assigned **only** in the unlinked-asker branch (`:233`, `createPublicDatasetToolRegistry`). The **linked-asker branch leaves it `undefined`** → `defaultRegistry` → `correct_fact` is in the tool set. `resolveLinkedActor` (`:365`) returns a real user with a real `activeOrganizationId`, and `resolveBrainReaderContext` re-resolves their real `member.role`. So when the linked asker is an org owner/admin, **a proactive Slack answer can retract or supersede brain facts** — an autonomous surface with no confirmation UI, where the only remaining guard is prose in `CORRECT_FACT_DESCRIPTION`.
2. **`packages/api/src/lib/chat-plugin/resume-turn.ts:150`** — `runAgent({messages: [], conversationId, resume})`, no `tools`. The original Slack turn ran through `executeAgentQuery` → `nonDashboardRegistry` (no `correct_fact`); the approval-**resume** of that same turn gets `defaultRegistry`. Fully closed at execute time here — the actor is `botActorUser` (role `member`, no member row) so the owner/admin gate refuses — but the tool surface silently widens across a resume boundary.

[...]

Issue #4958, opened 2026-08-01
sec(tests): CodeQL js/incomplete-sanitization false positive keeps the non-required umbrella red on milestone/brain-m2


msywulak:
CodeQL's non-required umbrella check has been red on the Brain M2 integration PR (#4950) since 2026-07-31. It is a **false positive**, and because the flagged line lives on `milestone/brain-m2`, it will keep the umbrella red on every M2 integration attempt and follow the branch into `main`.

## The alert

Alert [1281](https://github.com/AtlasDevHQ/atlas/security/code-scanning/1281) — `js/incomplete-sanitization`, **high**, created 2026-07-31.

- **Path:** `packages/api/src/lib/content-mode/__tests__/publish-caller-supersession-wiring.test.ts:253`
- **Message:** "This does not escape backslash characters in the input."

The flagged line:

```ts
// `$` is a regex anchor and a legal identifier character, so it is
// escaped rather than interpolated raw.
const ident = name.replace(/\$/g, "\\$");
```

## Why it is a false positive

`name` is not free-form input. It comes from `bound`, whose every element is capture group 1 of:

```ts
/^\s*(?:const|let|var)\s+([A-Za-z_$][\w$]*)\s*=/
```

That capture group admits exactly `[A-Za-z0-9_$]`. **A backslash is structurally impossible in the input**, as is every other regex metacharacter except `$` — which is precisely the one the line escapes. The sanitization is complete over its actual domain; CodeQL flags it because the rule reasons about `String.prototype.replace` against a general-purpose input, not against the constrained producer three lines up.

Two further mitigating facts: the code is **test-only** (a source-text scanner asserting that every `collectSupersessions(` caller binds its result), and its input is **the repository's own source files**, not untrusted data.

## Why it is not an M2 blocker

`CodeQL` is the **non-required umbrella**. The five required contexts on `main` are `ci`, `Analyze (javascript-typescript)`, `Deploy Validation`, `fork-pr-gate`, and `Image Scan` — all green on #4950, including `Analyze (javascript-typescript)`, which is the CodeQL analysis job that actually gates. #4950 was 68/69 with this as the sole red.

## Correction to the working assumption

[...]

PR #4959, opened 2026-08-01
chore(agent): make runAgent's `tools` required so the compiler enforces the surface gate (#4943)

msywulak:
Implements #4943. Makes `tools` a **required** property of `runAgent`'s options, so the compiler — not a source-text scanner — is the primary enforcement of the tool-surface gate. Follows #4936, which made the runtime default fail closed.

Base is `milestone/brain-m2`, so `Closes` would not fire; #4943 will be closed by hand.

## What changed

- **`tools` is required** on `runAgent`'s options (`tools: ToolRegistry`).
- **The fail-closed coalesce stays**, but moved off the destructuring default into the body: `const toolRegistry = declaredToolRegistry ?? nonDashboardRegistry`. A default on a *required* property is dead code to the compiler, and `typescript/no-useless-default-assignment` flags it as an error. It is not dead to callers the type system never sees — an `any`-typed options bag, a separately compiled plugin.
- **Every call site names its registry** — production and test. No production surface changed which registry it resolves to.
- **The scanner test's header records the new division of labour**: the compiler is the gate, the scanner is the backstop.

## Why the scanner is still not redundant

Four things a required property cannot catch, all still pinned by `agent-runagent-call-sites.test.ts`:

- **Which** registry each surface must resolve to — passing the *wrong* registry typechecks perfectly.
- `runAgent(opts)` with a pre-built bag, which compiles clean while laundering the posture out of sight.
- A **later spread that re-assigns `tools`** — object spread is last-write-wins, so this compiles clean under a required property. It is the one escape shape whose failure direction is **upward**, re-adding `correct_fact` rather than landing on the fail-closed default.
- Trees outside the type-checked graph (`scripts/`, `deploy/`, `examples/*`, `plugins/obsidian`), covered by the repo-wide `git grep` drift check.

## Verification

- `bun run type` — clean, exit 0 across all packages.
- `bun run scripts/test-isolated.ts --affected` — **644 files, 644 passed, 0 failed**.
- `bun run lint` — exit 0 (warnings only; the permanent non-targets per ADR-0031).
- **Mutation-verified, all five shapes.** Re-widening the property to `tools?: ToolRegistry` fails `bun run type` with exactly five `TS2578: Unused '@ts-expect-error' directive` errors — one per shape — then reverted. The five are pinned as permanent compile-time fixtures in `_compilerRejectsEveryOmissionShape`, so they keep failing if the property is ever re-widened:
  - ternary conditional spread (the pre-fix `ee/answer-adapter.ts` shape)
  - `&&` conditional spread (the pre-fix `routes/chat.ts` shape)
  - spread of a partial options object
  - plain omission (the pre-fix `resume-turn.ts` / `demo.ts` shape)
  - a possibly-`undefined` value

Note the mechanism is **not** spread-over-union distribution: TypeScript merges `{ tools: T } | {}` into one object type with `tools` optional, which then fails the required target on property incompatibility.

## Notes

- The behavioural test was made **load-bearing**: its base call now passes `defaultRegistry` (the write-carrying registry), not the value the fall-back resolves to. Passing `nonDashboardRegistry` there would have made the assertions hold whether or not the `OMIT_TOOLS` override landed — the test would have been pinning a registry it supplied itself.
- `Partial<Parameters<typeof runAgent>[0]>` is used in exactly one place — the test helper that must reach the fall-back coalesce from typed code. Everywhere else a forwarding helper takes the **full** parameter type; a `Partial` wrapper would fill `tools` in on its callers' behalf and erase the requirement for every surface behind it.

## Out of scope

- **`exactOptionalPropertyTypes`** — deliberately untouched, per the issue. It is what makes a possibly-`undefined` value assignable to an optional property and deserves its own issue. It did not bite during this work.
- Changing which registry any production call site uses.
- Widening the change to other functions taking an optional registry.

### DIFF
--- a/packages/api/src/lib/__tests__/agent-answer-style-prompt-shape.test.ts
+++ b/packages/api/src/lib/__tests__/agent-answer-style-prompt-shape.test.ts
@@ -111,8 +111,8 @@ function makeSpyingModel(): InstanceType<typeof MockLanguageModelV3> {
 }
 
 const { runAgent } = await import("@atlas/api/lib/agent");
-// #4943 — runAgent's `tools` is required; nonDashboardRegistry is the restricted
-// surface these turns already resolved to when they omitted it.
+// #4943 — runAgent's `tools` is now required; this is its own fail-closed
+// default, so these turns are unchanged. See agent.ts's `@param tools`.
 const { nonDashboardRegistry } = await import("@atlas/api/lib/tools/registry");
 
 function userMessages(text: string): UIMessage[] {
--- a/packages/api/src/lib/__tests__/agent-answer-style-prompt-shape.test.ts
+++ b/packages/api/src/lib/__tests__/agent-answer-style-prompt-shape.test.ts
@@ -111,6 +111,9 @@ function makeSpyingModel(): InstanceType<typeof MockLanguageModelV3> {
 }
 
 const { runAgent } = await import("@atlas/api/lib/agent");
+// #4943 — runAgent's `tools` is required; nonDashboardRegistry is the restricted
+// surface these turns already resolved to when they omitted it.
+const { nonDashboardRegistry } = await import("@atlas/api/lib/tools/registry");
 
 function userMessages(text: string): UIMessage[] {
   return [
@@ -128,6 +131,7 @@ async function runTurn(
 ): Promise<string> {
   lastSystemPrompt = undefined;
   const result = await runAgent({
+    tools: nonDashboardRegistry,
     // The acceptance criterion's simple-question case (#4299).
     messages,
     aiModel: {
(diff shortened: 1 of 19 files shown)


msywulak:
Also folds in **#4958** (commit `2bce764b7`) — the CodeQL `js/incomplete-sanitization` false positive at `publish-caller-supersession-wiring.test.ts:253`, which has been holding the non-required CodeQL umbrella red on the M2 integration PR #4950.

Fixed rather than dismissed because the alert sits on `milestone/brain-m2`: it would keep firing on every integration attempt and follow the branch into `main`, and a dismissal is per-alert so it does not survive the line moving.

Behaviourally a no-op over the real input domain — that file's 12 tests still pass.

Issue #5036, opened 2026-08-03
feat(residency): merge two vocabularies on import — union approved edges, refuse cycles


msywulak:
Slice E of ADR-0037 — the cross-vocabulary hole nobody had seen. **Scoped to merge semantics only.** The vocabulary's `bundle-scope` entry and its round-trip export/import wiring land in #5022, because `bundle-scope.test.ts` pins registry to implementation in both directions and they cannot be separated from the table's own PR. **The entity store is out of scope here** — it does not exist until #5043 in M5, and a registry entry for a nonexistent table fails `bundle-scope.test.ts:63`.

**Spec: the resolution comment on #5015 (T10) §4.** ADR-0037 §8 is the *why*.

**The forest invariant is enforced per-vocabulary, and nothing enforces it across two.** Source curated `price → priced at`; destination curated `priced at → price`. Both individually valid forests — canonical direction is arbitrary, so this is not contrived — and their **union has a 2-cycle**, which makes backfill non-terminating and order-dependent.

**Every merge rule currently in the importer is wrong for this table.** Fact/episode dedup is id-skip (`admin-migrate.ts:1105-1113`) and `fact_audience_member` is `ON CONFLICT DO NOTHING` (`:1206-1209`) — both mean **destination wins, silently**, which for vocabulary rows silently discards the source workspace's *approved human review work*. Two-vocabulary imports are real: idempotent re-import and catch-up import are both pinned, tested paths (`migrate-roundtrip-pg.test.ts:459-491`).

> **Union the approved edges; refuse any edge that would close a cycle or violate at-most-one-parent; log every refusal; recompute the effective-target closure at the destination.**

This is where #5022's two-relation model pays off a second time: merge the *human decisions* and derive the rest. Merging compressed targets would have meant merging two derived artifacts with no way to recover which decisions produced them.

## Key files
- `packages/api/src/lib/residency/export.ts` — the vocabulary and entity-store tables join the bundle
- `packages/api/src/api/routes/admin-migrate.ts` — the merge; `:1105-1113`, `:1206-1209` are the rules **not** to copy. Already on `check-brain-fact-promotion.sh`'s allowlist for `status`, so the key-column allowlist entry **extends an existing entry** rather than inventing one
- `packages/api/src/lib/residency/__tests__/migrate-roundtrip-pg.test.ts` — `:459-491`

## Acceptance criteria
[...]

Issue #5027, opened 2026-08-03
fix(brain): cardinality becomes a property of the canonical predicate; delete the both-sides clause


msywulak:
Slice C of ADR-0037, and the **fourth independent cause of #5000's symptom** — the only one that is not a string-matching problem.

**Spec: the resolution comment on #5010 (T6).** ADR-0037 §3 is the *why*.

**The map's own premise was false.** `predicate_cardinality` was believed unpopulated. It is not: `extract.ts:484` writes the model's per-claim guess and `correction.ts:1380` inherits it. `supersessionCollisionJoin` requires `'single'` on **both** sides (`brain-facts.ts:367-368`), and the two sides come from **two independent model calls on two different messages**, against a prompt that says *"When unsure answer 'multi'"* (`extract.ts:390`). So supersession fires at roughly P(model says `single`)². **An unpopulated column fails predictably; this one fails stochastically, on an irreversible operation.**

**Decision: cardinality attaches to the canonical predicate, is carried by the vocabulary, and is read live.** It has exactly one load-bearing consumer — `supersessionCollisionJoin` at publish time (verified: neither `TENSION_CANDIDATES_SQL` nor `CORROBORATION_LOOKUP_SQL` reads it; the rest are display and inheritance) — so there is no seam to drift, and T2's materialize-at-write rule does not reach it.

**`single` requires positive evidence. Absent → `multi`; ambiguity → `multi`.** A predicate whose cardinality depends on the subject's type (`located in`) is simply **never marked `single`**, so it never has to be adjudicated. **We under-supersede deterministically rather than supersede stochastically.**

## Key files
- `packages/api/src/lib/content-mode/adapters/brain-facts.ts` — `supersessionCollisionJoin` `:367-368`; `:155` already says *"`predicate_cardinality` is the supersession input"*
- `packages/api/src/lib/brain/extract.ts` — `:383-391` the prompt, `:390` the bias ("When unsure answer 'multi'"), `:408` a second instruction in the system prompt, `:484` the write
- `packages/api/src/lib/brain/correction.ts` — `:1380` the inheritance
- The vocabulary from #5022 — cardinality is its second property, under #5023's authority

## Acceptance criteria
- [ ] Cardinality is stored on the canonical predicate in the vocabulary and **read live** at the one consumer
[...]

Issue #5026, opened 2026-08-03
chore(residency): classify the vocabulary and entity-store tables in bundle-scope


msywulak:
Slice B of ADR-0037 — **forced, not optional.** `bundle-scope.ts` is a per-table decision registry and `bundle-scope.test.ts` enumerates the Drizzle schema, **failing on any table with no entry**. The vocabulary table and the entity store cannot be built without this answer; the tripwire fires at merge time.

**Spec: the resolution comment on #5015 (T10) §4(a).** ADR-0037 §8 is the *why*.

**`stays` is not neutral — it is deletion.** `bundle-scope.ts:66-68`: once #4458's destructive cleanup ships, `stays` rows are **DELETED** from the source region after the grace period, and the test says classifying the brain `stays` *"would make a region migration silently destroy the workspace's accumulated knowledge."* A `stays` vocabulary is a **double** loss — curated aliases deleted at source, and the verbatim keys that travel on the bundle left permanently un-re-derivable.

ADR-0024 raises no objection: it isolates the **identity plane** (Better Auth tables, never in the registry), a region migration is a **move** not a replication, and a workspace-scoped vocabulary is the same asset class as `brain_facts`.

## Key files
- `packages/api/src/lib/residency/bundle-scope.ts` — the registry; `:66-68` on what `stays` means
- `packages/api/src/lib/residency/__tests__/bundle-scope.test.ts` — the enumerating tripwire

## Acceptance criteria
- [ ] Both the vocabulary table and the entity-store table are classified **`exported`**
- [ ] The rationale is recorded at the entry, not just in this issue: the vocabulary is workspace-scoped with no ACL (T11 §1), so unlike `visible_to` there is no grant to narrow on the way out — it travels as ordinary workspace data
- [ ] Proposal *evidence* is explicitly out of scope: it is not a key and does not travel; an imported region re-derives its own candidates from its own rows
- [ ] `bundle-scope.test.ts` passes without a suppression

## Dependencies
- Depends on #5022 (the tables must exist to be classified)
- The export/import *implementation* and the cross-vocabulary merge are slice E — this issue is the registry decision only

msywulak:
**Folded into #5022 — not dropped.**

This was cut as a standalone "registry decision only" issue on the belief that classifying a table in `bundle-scope.ts` is separable from creating it. A review of `bundle-scope.test.ts` shows it is not, in three directions at once:

[...]

Issue #5025, opened 2026-08-03
feat(web/admin): the Claim Vocabulary approval surface (umbrella) — preview, authoring, queue


msywulak:
> **⚠️ UMBRELLA as of 2026-08-08 — do not implement this issue directly.** Work the three children in order. This issue keeps the design record: the [2026-08-05 design pass](https://github.com/AtlasDevHQ/atlas/issues/5025#issuecomment-5205922696), the [two](https://github.com/AtlasDevHQ/atlas/issues/5025#issuecomment-5221613668) [grill checkpoints](https://github.com/AtlasDevHQ/atlas/issues/5025#issuecomment-5221679841), the [#5027 handoff](https://github.com/AtlasDevHQ/atlas/issues/5025#issuecomment-5195180009), and the [2026-08-08 implementation checkpoint](https://github.com/AtlasDevHQ/atlas/issues/5025#issuecomment-5228156017) whose four findings revise ACs below. The ACs stay here as the spec the children inherit; each child restates the ones it owns.
>
> 1. #5086 — **the blast-radius counterfactual.** Written; needs `-pg` falsification, panel, CI
> 2. #5087 — **direct authoring + the In-force pane.** The day-one useful half; owns the positional-visibility seam
> 3. #5088 — **the pending queue.** Correct but empty until a producer fires; imports that seam
>
> **Why:** two grill checkpoints added ~20 ACs on top of ~15, each needing its own falsifier (`/review-panel` Step 6, which #5027 paid four rounds for the lack of). That is milestone-sized, not slice-sized — and the M4 arc's whole shape is independently-releasable pieces that are no-ops until their successor consumes them.
>
> **The one AC that only survives a split if both children say so**, and both do: *"it is the same code path as the Pending visibility check, not a copy."* #5087 owns and exports the positional-visibility seam; #5088 imports it. `oversight.ts:800-803` — a disclosure that restates a rule drifts from it.

---

Slice B of ADR-0037 — the reviewer-facing surface for the vocabulary. **`ready-for-human`**: the authoring UX is a judgment call this cut does not make, and the disclosure rules below are exacting enough that getting them wrong leaks claim history.

**Spec: the resolution comment on #5016 (T11) §5.** ADR-0037 §6 is the *why*.

**The queue cannot obey the usual "a number, never content" rule.** A proposal necessarily carries both surfaces, because you cannot approve `X → Y` without seeing X and Y. So visibility is handled **positionally** instead:

[...]

PR #5051, opened 2026-08-04
feat(brain): the vocabulary — approved edges plus a derived effective-target closure, position-scoped (#5022)

### LINKED ISSUE #5022: feat(brain): the vocabulary — approved edges plus a derived effective-target closure, position-scoped
Slice B of ADR-0037 — the vocabulary that makes `alias` do something. This is the schema and the closure, not the approval flow (#5023) and not the UI (#5025).

**Spec: the resolution comment on #5016 (T11) §4.** ADR-0037 §6 is the *why*.

**The vocabulary is two relations, not one.** T3 §8's forest invariant was self-contradictory — it stated depth-1 (*"every canonical target is itself unaliased"*) **and** asserted composition works, and both cannot be literally true: approving `price → unit price` makes `price` an aliased target of `is priced at`. The only reconciliation is path compression at approval time, and compression **destroys the reversibility T3 called the sole thing that makes a bad alias undoable** — after compressing, removing `price → unit price` cannot restore `is priced at → price`, because that edge is gone.

| | What it is | Written by |
|---|---|---|
| **Approved edges** | the human's decisions; at-most-one-parent enforced; **never rewritten by another approval** | approval / removal only |
| **Effective target** | the transitive closure of the approved edges — what `alias` reads | derived, recomputed |

Removal becomes a **recomputation** rather than a destructive write, and `alias` is a function by construction. Note that *"no cycles + targets unaliased"* does **not** by itself imply at-most-one-parent, which is why it is enforced explicitly.

**Position-scoped.** A position-agnostic vocabulary would not merely permit cross-position composition, it would **compel** it: `owned by → platform` plus `platform → platform team` violates the forest, enforcement resolves it by composing, and a *predicate* approval then re-keys **subjects** workspace-wide — silently, in the irreversible direction. The overlap surface is real: warehouse predicates are bare common nouns (`price`, `owner`, `status`, `tier`, `region`), exactly the surfaces most likely to also be subject or object norms.

## Key files
- `packages/api/src/lib/db/schema.ts` + a new migration — the approved-edge table and the derived effective-target representation
- `packages/api/src/lib/brain/` — the `alias` lookup seam stubbed in #5020 now reads the closure

## Acceptance criteria
- [ ] Two relations: approved edges (durable, the human's decisions) and effective target (derived, recomputed). `alias` reads the **effective target**
- [ ] At-most-one-parent is enforced explicitly at write time, as is cycle refusal
- [ ] An approval **never rewrites a previously approved edge**
- [ ] Removing an edge recomputes the closure and **restores the prior effective target** — falsify this through a *compressed* chain (`a → b`, `b → c`, remove `b → c`, assert `a` lands back on `b`), or the test is vacuous
- [ ] Edges are **position-scoped**: a predicate-position approval must not re-key subjects. Assert it
- [ ] Cardinality (slice C) will attach as a second property of the same canonical predicate — leave room for it rather than designing it out
- [ ] ⚠️ **This PR must also land the `bundle-scope` registry entry and minimal export/import wiring for the vocabulary tables — they cannot be a separate PR.** `bundle-scope.test.ts` pins registry and implementation in both directions: `:40` fails any schema table with no entry, `:110-145` fails any `exported` table lacking `FROM <t>` in `export.ts` and `INSERT INTO <t>` in `admin-migrate.ts`, and `:63` fails an entry for a table that does not exist. So the table, its `exported` classification and its round-trip wiring are one atomic merge. #5036 owns only the **merge semantics** on top
- [ ] Classify **`exported`**, with the rationale at the entry: workspace-scoped with no ACL, so unlike `visible_to` there is no grant to narrow on the way out — it travels as ordinary workspace data. **`stays` is deletion**, not neutrality (`bundle-scope.ts:59-60`): #4458's cleanup deletes `stays` rows from the source region, which for a vocabulary is a double loss — curated aliases destroyed at source *and* the verbatim keys on the bundle left permanently un-re-derivable
- [ ] Add the vocabulary row to `data-residency.mdx`'s "What moves" table — `bundle-scope.ts:28` makes that a sync obligation, and no other issue owns it
- [ ] Vocabulary rows are workspace-scoped with **no ACL arm**. This is a derived invariant, not a choice: all three identity consumers already carry no grant arm, and grant-scoping would need `alias(norm, reader)` at a seam materialized by a fiber that has no reader

## Dependencies
- Depends on #5020
- Blocks slice C (cardinality attaches to the canonical predicate). **Absorbs the `bundle-scope` classification that #5026 originally held** — see the criterion above for why it cannot be split out
- **Counter-case recorded, not to be re-litigated:** T3 §3 chose one namespace so a curated entry and an uncurated key are directly comparable; position-scoping reintroduces a second space to keep from colliding, and three forests is three enforcement paths that can drift

msywulak:
Closes #5022. Slice B of [ADR-0037](../blob/main/docs/adr/0037-claim-identity-in-the-brain.md) §6 — the schema and the closure, not the approval flow (#5023) and not the UI (#5025).

`alias` — the outer layer of `key = alias(lexicalNorm(surface))` — stops being the identity function and becomes a data table.

## Two relations, not one

| | What it is | Written by |
|---|---|---|
| `brain_vocabulary_edge` | the human's approved decisions; at-most-one-parent; **never rewritten by another approval** | approval / removal only |
| `brain_vocabulary_target` | the transitive closure of those edges — what `alias` reads | derived, recomputed wholesale |

§6 retracts T3's "forest invariant" by name for being self-contradictory (depth-1 **and** composing), and its only reconciliation under one relation is path compression at approval time — which rewrites edges nobody approved in that action and destroys the reversibility T3 called the sole thing making a bad alias undoable. Split, removal is a **recomputation**: drop `price → unit price` and `is priced at` lands back on `price`.

## Position-scoped, in the schema *and* in the type

`slot_position` is part of both primary keys, and `ReconcileRequest.alias: AliasLookup` becomes `vocabulary: ClaimVocabulary` — one lookup per slot, read as `vocabulary.subject` / `.predicate` / `.object` beside the surface being keyed.

**This edits a call site #5020's comment promised slice B would not have to**, and that is deliberate. #5020 hung a single bare `AliasLookup` on the request and threaded it through all three `slotKey` calls. That is the position-agnostic vocabulary §6 rules out: it does not merely *permit* cross-position composition, it **compels** it, and a predicate approval then re-keys subjects workspace-wide in the direction nothing undoes. Both comments are corrected in place rather than left promising something now false.

## What landed

- **Migration 0189 + the Drizzle mirror.** At-most-one-parent is the primary key; the 1-cycle is a CHECK; derived-ness is a RESTRICT FK — deliberately **not** CASCADE, which looks tidy and is wrong (with `a → b`, `b → c`, cascading the deletion of `b → c` drops `b`'s closure row and leaves `a` pointing at a `c` nobody approves any more). RESTRICT makes "remove an edge without recomputing" unrepresentable.
- **`lib/brain/vocabulary.ts`** — `approveAliasEdge` / `removeAliasEdge` / `recomputeEffectiveTargets` / `loadClaimVocabulary`. Longer cycles and all four typed refusals are enforced explicitly at write time under an advisory lock (namespace 5022, distinct from reconcile's 4771 so approving an alias does not serialize against ingest). The closure rebuild **verifies convergence** and throws rather than committing a silently truncated walk.
- **bundle-scope + export/import wiring + the `data-residency.mdx` row**, all here — they cannot be split (#5026).
- **Cardinality (slice C, #5027) has room**: the derived table stays free of authored state, because every recompute deletes and rebuilds it, so a human-set cardinality parked there would be destroyed by the next unrelated approval.

## One deviation from the issue text, stated

The issue says classify the vocabulary **tables** `exported`. The edges are `exported`. The **derived closure is `stays`**, and the import rebuilds it — which is what ADR-0037 §8 itself describes the merge doing ("unions **approved edges** … and **recomputes the closure**"). Carrying a source closure into a destination that already holds a vocabulary would be a closure of neither. `stays` is the `dashboard_draft_card_cache` / `pii_column_classifications` case — derived, regenerated at the destination from inputs that *do* move — so it destroys nothing, and the issue's real warning ("`stays` destroys curated aliases at source") applies to the edges, which are exported. Rationale is recorded at the registry entry.

Its cleanup rule is `expression`-kind **for phase, not for scoping** (same shape as `brain_facts`'s `parent` rule): the RESTRICT FK means the closure rows must be deleted before the column phase reaches the edges, or the #4458 sweep fails outright on any workspace that has approved an alias. A dedicated tripwire pins that choice so it stays a decision rather than an invisible property of declaration order.

## Falsification

Reversibility is proven through a **compressed** chain (`a → b`, `b → c`, remove `b → c`, assert `a` lands back on `b`) — a two-edge test would be vacuous, and §6 says so.

**Every guard was verified to fail without its implementation.** 12 mutations were actually run against `vocabulary-pg.test.ts`; the table is in that file's header with the count each kills. Four are caught by exactly one test each, because the *schema* also refuses — deleting the TypeScript guard turns a typed refusal into a raw constraint violation, and only an assertion on the refusal **value** can tell those apart. A version written as `expect(await storedEdges()).toHaveLength(1)` would pass under all four.

Also verified by mutation: the position-scoping assertion in `reconcile.test.ts` (bind `vocabulary.predicate` to all three slots → dies), and the import-side recompute (delete it → the round-trip's closure assertion dies).

Coverage:
- `vocabulary-pg.test.ts` — 22 tests, every prohibition paired with a positive control in its own `test()`. The no-ACL invariant sits **outside** the Postgres gate so it runs in the default local and `--affected` lanes.
- `migrate-roundtrip-pg.test.ts` — a compressed chain round-trips, and the target's closure is asserted to be a value the importer *walked*, since the bundle carries none.

## Scoped out

The approval flow and rejection memory (#5023), the UI (#5025), cardinality (#5027), and the import-time vocabulary merge (#5036) — the interim behaviour on a cycle-closing arrival is a loud abort, documented at the import site.

## Verification

- `scripts/ci-local.sh` — **33/33 PASS**.
- Migration 0189 hit no number collision on this branch; renumber if one appears on rebase.


msywulak:
## Review panel — 21 findings, all fixed (`26a1fcb78`)

Ran the four specialist reviewers over the diff. Two of them verified rather than read: the test analyst mutation-tested a copy of the tree (**9 of 12 mutations survived**), and the comment analyst disproved two of my SQL claims against the live Postgres. That was worth it — three findings were things I had asserted without checking, and one was something I had directly observed and then written the opposite of.

### Real defects

- **The import was a second writer into `brain_vocabulary_edge` and the only one that did not re-norm.** `{fromNorm: "Price", toNorm: "price"}` passed every validation arm and every CHECK, landing a post-norm 1-cycle; `{toNorm: " - "}` nulled the key of every claim with that predicate. It now **refuses** a non-norm row and names it — §8 carries a row-copy path verbatim, so silently rewriting another region's decision is not available.
- **The import mutated the vocabulary without the workspace lock** — the invariant `vocabulary.ts` states, asserted and violated in the same PR.
- **The transaction contract was documented and unenforced.** `VocabularyExecutor` is satisfied by a pool *on purpose*, so `removeAliasEdge` on one would COMMIT an empty closure between its DELETE and its rebuild — a whole episode keyed un-aliased, permanent if the process died in the window. Now probed via `pg_locks` and refused in all three primitives.
- **`loadClaimVocabulary` could not tell "approved nothing" from "closure half rebuilt"** — both degraded to `identityAlias` with no error to propagate. Every edge contributes exactly one closure row, so the counts are the check.
- **The `existingTarget` assertion agreed by construction.** One seeded edge makes the raw parent and the effective target the same string, so repointing the read at the closure survived. That is #5000's own root cause, reproduced in the suite written to prevent it — falsified through a compressed chain now.
- **`validateBundle`'s vocabulary block was dead to the suite** — deleting all 1637 characters left 79 tests passing.

### Comments that were factually wrong

Every one of these was mine, and all were load-bearing:

- `cleanup.ts` claimed demoting the closure's rule "would happen to work — declaration order puts it ahead of the edge table." **Exactly backwards** — the edge table is declared first, so it would abort the sweep on the RESTRICT FK. The comment made the only thing holding the sweep together read as belt-and-braces.
- "the CHECKs stop a hand-written INSERT landing `Priced At`" — they do not, and cannot without a third `lexicalNorm`.
- The `roots` CTE was defended by a Postgres rule that does not exist (that is `SELECT DISTINCT`'s, not `DISTINCT ON`'s).
- The convergence check fires on **odd**-length cycles only; an even one dies on `ck_..._not_self` first. I had hit exactly that failure while writing the test and then wrote the opposite in the module.
- `identity.ts`'s "`alias` is REQUIRED" section still named `ReconcileRequest.alias` and `grep identityAlias` after the rename.

### Types

`vocabulary` is now **required** on `ReconcileRequest`, and moved from `CorrectionDeps` onto `CorrectionRequest` — it is workspace state, not a test seam. This is the hazard `identity.ts` argues at length for `slotKey`'s parameter, one layer up: 66 call sites now name `identityVocabulary` out loud, and #5023 cannot wire the loader and miss one. Also: `already-aliased` gets its own arm so `existingTarget` is required where it means something; `VocabularyClosureError` so #5023 can tell *corrupt* from *down*; `isSlotPosition` retires three unchecked casts; the wire and internal position unions are pinned in both directions.

### Falsification

**29 mutations run, 28 killed.** The survivor was a post-norm 1-cycle arm no input could reach — the re-norm check runs first and subsumes it — so it was **removed** rather than kept. A rule with no failure mode reads as protection.

New coverage: the removal path's re-norm, the cycle walk's position scoping, all three slots reading their own lookup, the correction threading (both halves — the guard *and* the replacement's key), `approved_by`/`approved_at` surviving the hop, the cleanup sweep actually deleting vocabulary rows, the import's cycle-abort, concurrent approvals serializing (with a different-workspaces control that deadlocks if the lock is keyed on a constant), and — the one the slice exists for — an approved alias making two spellings **corroborate** end to end through the real ingest stage, with its fork control.

The mutation table in `vocabulary-pg.test.ts`'s header was re-measured in full; two rows are now marked compound rather than overclaiming.

`/ci`: **33/33**.

Issue #5061, opened 2026-08-05
chore: convert the remaining six hand-measured mutation tables to the runner


msywulak:
#5060 built `scripts/mutate.ts` and converted one table (`object-cmp.test.ts`) as the proof. Six hand-measured tables remain. Each is a column of numbers stored in a docstring that nothing can falsify, and converting the first one found two defects in sixteen cells — one cell stale (two independent spellings of the mutation both measure 4 against a published 3) and one cell spelling-dependent (3 when both return sites are collapsed, 2 when only the final one is, with the docstring never recording which it meant). Assume a comparable hit rate in the rest.

Convert one table per commit. For each: write `scripts/mutations/<name>.mutations.ts` with the exact before/after strings, run the runner, replace the docstring table with a pointer to the generated file, and record any cell that moved in the mutation's `note` — a moved number is the finding, not an embarrassment to smooth over. Do NOT hand-copy the old numbers forward; if a cell disagrees, the generated one is right by construction.

## Key files
- `packages/api/src/lib/brain/__tests__/identity-consumers-pg.test.ts` — needs `TEST_DATABASE_URL`
- `packages/api/src/lib/brain/__tests__/promotion-pg.test.ts` — two-column table (`here` / `corpus suite`); use two `targets`, which is the case the multi-target support was built for
- `packages/api/src/lib/brain/__tests__/object-cmp-pg.test.ts` — needs `TEST_DATABASE_URL`
- `packages/api/src/lib/brain/__tests__/vocabulary-decide-pg.test.ts` — needs `TEST_DATABASE_URL`
- `packages/api/src/lib/brain/__tests__/vocabulary-pg.test.ts` — needs `TEST_DATABASE_URL`
- `packages/api/src/lib/brain/ingest/__tests__/episode-sync-archive.test.ts` — pure TS
- `packages/api/scripts/mutations/` — one `<name>.mutations.ts` + generated `<name>.md` per table

## Acceptance criteria
- [ ] All six tables generated from checked-in mutation lists; no `| Mutation | Dies on |` table remains hand-typed in any docstring
- [ ] `grep -rn 'MUTATIONS THIS' packages/api/src` returns only pointers to generated files
- [ ] Every cell that disagrees with the docstring it replaced is called out in that mutation's `note`, saying whether it was stale or spelling-dependent
- [ ] Any row measuring `0` is either justified in a `note` or closed with a new test — an unexplained zero is a coverage gap, not a table entry
[...]

Issue #5033, opened 2026-08-03
feat(brain): tier guard inside supersessionCollisionJoin — a draft may not stamp a published warehouse-derived fact


msywulak:
Slice D of ADR-0037, and **the prerequisite that gates the entire warehouse producer.** This map is precisely the change that makes the collision match; without this guard, shipping a producer opens the window deliberately.

**Spec: the resolution comment on #5008 (T4) §4.** ADR-0037 §4 is the *why*.

> A draft LLM-extracted fact, at publish time, stamps `valid_to` on a published **warehouse-derived** fact.

Every arm of `supersessionCollisionJoin` (`brain-facts.ts:361-372`) is satisfied once keys collide — same workspace, same slot, different object, both `single`, the warehouse row published and live. **There is no tier or source arm in that join.** And tier-1 has no correction path (`correction.ts:67`), so the reviewer who notices cannot undo it: no verb reaches the row, and both vouching verbs refuse a closed window. An LLM guess irreversibly retires an authoritative-by-construction fact.

**The framing to hold onto, because the map's own source-agnostic note reads as forbidding this fix: *identity is source-agnostic; consequence is tier-ordered.*** The slot join, the agreement comparison, the tension edge and the key function stay byte-identical for every source. A tier arm on the **supersession** clause is not per-class matching — ADR-0036 §T4's trust tiers already *are* a consequence ordering.

**Cross-tier collision is tension-only in both directions.** The symmetric case — a newly-produced warehouse fact colliding with a published extracted fact — also stops at tension. Auto-stamping there is autonomous supersession by the ADR's own definition, merely with the sympathetic side winning; a stale extracted fact in visible tension is recoverable, a stamp is not, and the warehouse row is a snapshot that may already be hours old.

**This corrects #4759 §3 item 1's *"gone by construction"***. What warehouse-wins removes by construction is the reviewer's *ambiguity*, not the *edge*. Warehouse-wins does nothing mechanical whatsoever — it is a ranking hint in the review UI, and the reviewer load it was supposed to remove is not removed.

## Key files
- `packages/api/src/lib/content-mode/adapters/brain-facts.ts` — `supersessionCollisionJoin` `:361-372`
- `packages/api/src/lib/brain/correction.ts` — `:67`, why tier-1 has no correction path
- `packages/api/src/lib/brain/oversight.ts` — the disclosure that must show the same held-back pair

## Acceptance criteria
[...]

Issue #5030, opened 2026-08-03
feat(brain): object_cmp, the fail-closed typed parser, and three-valued agreement


msywulak:
Slice D of ADR-0037 — the abstain band. The slot join proves *sameness*; this adds the column that can prove *difference*, and the three-valued comparison that both destructive operations read.

**Spec: the resolution comment on #5006 (T2) and on #5007 (T3) §4.** ADR-0037 §2 is the *why*.

`same` and `different` are not complements, and a single column compared two ways fails quietly in both directions. Two columns:

| Column | Null | Proves |
|---|---|---|
| `object_key` | no | *sameness* |
| `object_cmp` | **yes** | *difference* — a typed canonical value, parsed **fail-closed** |

- **same** — `object_key` equal, **or** both `object_cmp` non-null and equal
- **different** — both `object_cmp` non-null and unequal
- **unknown** — everything else → **tension only, never a stamp**

A type qualifies for `object_cmp` only if its parse is unambiguous and its equality decidable: money-with-explicit-currency, plain number, date/timestamp, boolean, resolved entity id. **Bare `$499` parses to NULL** — `$` is ambiguous across USD/CAD/AUD. **The design is safe even when the parser is cowardly**, which is the property it needs to have.

## Key files
- `packages/api/src/lib/db/schema.ts` + a migration — `object_cmp`, nullable
- `packages/api/src/lib/brain/reconcile.ts` — the parse at write time, and the agreement arms in `TENSION_CANDIDATES_SQL` / `CORROBORATION_LOOKUP_SQL`
- `packages/api/src/lib/content-mode/adapters/brain-facts.ts` — `supersessionCollisionJoin`'s difference arm

## Acceptance criteria
- [ ] `object_cmp` is nullable; `object_key` stays `NOT NULL`
- [ ] The three verdicts are implemented exactly as above. `unknown` falls to **tension only, never a stamp**
- [ ] The parser **fails closed to NULL** on anything ambiguous. Bare `$499` → NULL
- [ ] Producers may **declare** the object's type (a warehouse producer knows `price` is USD money; the extractor guesses), on the cardinality precedent at `extract.ts:383-391` — a producer-declared property of the claim with a conservative default, **not** a matching rule. This stays inside the source-agnostic line
[...]

PR #5062, opened 2026-08-05
chore(test): the mutation runner — generated tables, not hand-typed ones (#5060)

### LINKED ISSUE #5060: chore: generate mutation tables with a runner instead of hand-measuring them
Seven Brain test files carry hand-measured mutation tables (`MUTATIONS THIS CATCHES` / `| Mutation | Dies on |`). Every cell is a number that must equal a count produced by running a mutated tree, but it is stored by hand in a docstring. Add a test and N cells silently become false, and nothing in CI notices. This has produced a review-panel finding in every one of the last three slices (#5022, #5030, #5033), and the files now document their own decay: `object-cmp.test.ts` records "Regenerated in ONE pass on the final tree, and four rows had gone stale — 20→22, 11→13, 3→4, 2→3", and `promotion-pg.test.ts` records a cell measured as `IS DISTINCT FROM 'warehouse'` in one file and `<> 'warehouse'` in another — two SQL spellings that differ on exactly one input — publishing a number true of neither.

Build a mutation runner that takes the mutation list as DATA (exact before/after strings, never a prose description) and emits the table. The spelling becomes the input rather than a description of the input, which is what kills the two-spellings class outright. Ship the runner plus one converted table as proof; the remaining six are a separate follow-up.

## Key files
- `packages/api/scripts/mutate.ts` — new. Sits beside `test-isolated.ts` and `affected.ts` because every target suite is under `packages/api` and the suites require that cwd for preload
- `packages/api/src/lib/brain/__tests__/mutations/` — new. One `<suite>.mutations.ts` data file per converted table, plus the generated `.md` the docstring points at
- `packages/api/src/lib/brain/__tests__/object-cmp.test.ts` — the proof conversion; its docstring table is replaced by a pointer to the generated file
- `packages/api/src/lib/brain/__tests__/promotion-pg.test.ts` — spot-check target only, to prove the runner drives a `-pg` suite; its table is NOT converted here

## Acceptance criteria
- [ ] Mutation list is data: `{label, edits: [{file, oldString, newString}]}` — exact strings, no prose
- [ ] Baseline runs first and the run ABORTS if the baseline is not green
- [ ] Applying an edit asserts the anchor appears EXACTLY once; 0 matches or 2+ matches aborts that mutation rather than reporting a number
- [ ] Restore is from an in-memory backup of the touched files, never `git checkout --` (the tree normally carries uncommitted work)
- [ ] Restore happens even when the suite run throws or the process is interrupted
- [ ] Warns when a mutation's fail count is at or near the whole suite — the "could not determine data type" class, where an untyped param substitution fails every test and records 51 where the truth is 1
- [ ] `--only <label>` and `--target <file>` for iterating without paying ~3s/suite x ~25 mutations
- [ ] Emits a markdown table to a generated file that the test docstring POINTS AT, so the table cannot be hand-edited into staleness
- [ ] `object-cmp.test.ts`'s table is generated from a checked-in mutation list and its numbers reproduce
- [ ] Runner is demonstrated against one `-pg` suite (needs `TEST_DATABASE_URL`; scratch DB on the agents Postgres)

## Dependencies
- None

msywulak:
Closes #5060.

Mutation tables were measured by hand and typed into test docstrings. Every cell is a number that must equal a count produced by running a mutated tree, stored where nothing can falsify it — add a test and N cells silently become false. This produced a review-panel finding in each of the last three slices (#5022, #5030, #5033), and the files had begun documenting their own decay.

`packages/api/scripts/mutate.ts` takes the mutation list as **data** — exact `oldString` / `newString`, never a prose description — and generates the table. That single decision kills the two-spellings class outright: #5033 measured a denylist as `IS DISTINCT FROM 'warehouse'` in one file and `<> 'warehouse'` in another (they differ on exactly one input) and published a cell true of neither.

## What converting one table found

Sixteen cells, fourteen reproduced exactly, **two did not** — one of each failure mode in the diagnosis:

| Row | published | measured | why |
|---|---|---|---|
| a declaration OVERRIDES the surface | 3 | **4** | stale — two independent spellings of the mutation both measure 4 |
| collapses `declaration-rejected` into `abstained` | 2 | **3** | spelling-dependent — both return sites 3, final return alone 2, and the docstring never said which |

Both are recorded in the generated file's notes rather than quietly corrected.

## Guardrails

Each closes a specific way the tool could report a confident wrong number:

- **Red baseline aborts.** Otherwise every count is pre-existing breakage plus the mutation's, indistinguishable from a strong result.
- **The anchor must match exactly once.** A 0-match measures the *unmutated* tree and publishes a `0` that reads as "no coverage". A 2-match mutates whichever site string order reached first. Both report a number for a mutation nobody performed.
- **Restore from an in-memory backup, never `git checkout --`.** The tree normally carries uncommitted work — this runner is used mid-slice.
- **Whole-suite flag.** A count reaching ~every test is the "could not determine data type" shape, where an untyped param substitution fails the whole file and records 51 against a truth of 1.
- **Per-suite timeout**, scaled off the clean baseline. Found by dogfooding — see below.

## Dogfooded

The runner is measured against its own guardrails (`scripts/mutations/mutation-core.mutations.ts`), because it is now the thing every other table in the repo trusts. All 18 rows kill at least one test; no unexplained zeros. That pass found two real defects in this PR:

1. **A hang.** Removing `countOccurrences`' empty-needle guard makes its `indexOf` loop infinite, and the runner had no timeout — one hanging mutation hung the whole run. Worse, an unbounded wait invites Ctrl-C, which lands in the one window where it costs the operator their tree. Now bounded, and that row reports an honest `timed out` rather than a number. (The SIGTERM restore path was verified for real when the original run was killed: tree came back clean.)
2. **A decorative test.** The line-anchoring fixture contained no decoy the unanchored regex would prefer, so it passed against both spellings and measured 0 deaths. Fixed; it now dies.

## Scope

- Runner + one converted table, per the plan. The remaining six are #5061, with the two-column `promotion-pg` case called out — multi-target support exists for it.
- `--only <label>` / `--target <name>` for iterating; a filtered run **refuses to write** the file, since writing a subset would delete the rows it did not measure — the exact staleness this tool prevents, under a header vouching for it.
- Output is deterministic (no timestamps), so `--check` is a usable CI gate once the rest land.
- Logic split into `mutation-core.ts` on `signal-retry.ts`'s precedent, so the guardrails are unit-testable without spawning a suite or writing to the tree.

## Verification

- `/ci` — **33/33 gates PASS**
- `bun test src/__tests__/mutate-core.test.ts` — 41 pass
- `bun test src/lib/brain/__tests__/object-cmp.test.ts` — 58 pass (unchanged; only the docstring moved)
- `bun run scripts/mutate.ts scripts/mutations/object-cmp.mutations.ts --check` — CHECK OK


msywulak:
CodeQL disposition, so it is on the record rather than merged past silently:

`js/incomplete-sanitization` (HIGH) on `escapeCell` was **real** and is fixed — same shape as the Confluence cell escaper in #4389. Escaping `|` alone sent `a\|b` to `a\\|b`, where the `\\` is a literal backslash and the `|` behind it is a live delimiter again.

The alert **will stay red**, and that residue is a false positive for this renderer. CodeQL wants an unconditional `.replace(/\\/g, …)`; doubling every backslash would corrupt real labels, because GFM special-cases `\|` at the row-splitting stage and nothing else, and labels carry backslashes inside code spans (``MONEY_RE back to `\s+``` in `object-cmp.mutations.ts`) where markdown escapes do not apply. The blanket fix renders a visible `\\s+` and dirties a generated table whose byte-identity is the whole point of #5060.

The argument lives in the function doc comment so the next reader does not "fix" it back, and mutation C (blanket-double every backslash) is in the suite specifically to fail that regression.

`CodeQL` is not a required check here — required are `ci`, `Analyze (javascript-typescript)`, `Deploy Validation`, `fork-pr-gate`, `Image Scan` — and `Analyze (javascript-typescript)` is green.

msywulak:
**Correction to my earlier comment:** CodeQL is now **green** on this PR — zero open alerts.

I predicted the alert would persist because the escape is pipe-conditional rather than unconditional. That was wrong: the single-pass form (`/\\?\|/g` with a function replacer) satisfies `js/incomplete-sanitization` without doubling benign backslashes, so the fix is clean on both axes rather than a dispositioned false positive.

The doc comment has been corrected to match — it now says the *two-replace* spelling is what trips the rule, and warns against blanket-escaping to silence it if someone later splits the single pass back apart. The previous wording told readers to expect and ignore a red CodeQL here, which after this became both false and a trap.

Issue #5038, opened 2026-08-03
chore(brain): decide the deferred unique index on the live slot — add subject_cmp or drop it


msywulak:
A one-decision issue, deliberately sequenced last. T3 §7 deferred a unique index on `(workspace_id, subject_key, predicate_key, object_key)` over the live set — it would turn corroboration into an upsert and make duplicate live facts unrepresentable, closing the `_Avoid_` that `CORROBORATION_LOOKUP_SQL` currently *is* (application-side read-then-insert as the only dedup).

**Spec: §7 of the resolution comment on #5007 (T3), and §3 of the resolution comment on #5017 (T12).**

**`ready-for-human` because the two inputs point opposite ways.** T3 sequenced the constraint after T10 so it could not reject a legitimate region import. T12 then found something stronger: under a `subject_cmp`-less world the constraint would make two referents' identical claim **unrepresentable as two rows**, cementing a homonym merge at the DB level and foreclosing any later split. **It must gain `subject_cmp` or be dropped** — and with `subject_cmp` NULL for every extractor-supplied subject, a nullable column in a unique index has its own semantics to reason about.

## Key files
- `packages/api/src/lib/db/schema.ts` — `brain_facts` indexes
- `packages/api/src/lib/brain/reconcile.ts` — `CORROBORATION_LOOKUP_SQL`; `withBrainTransaction`'s advisory lock already closes the concurrency half of the `_Avoid_`, which `INSERT_PROVENANCE_EDGE_SQL` relies on by name

## Acceptance criteria
- [ ] A decision is recorded: add the index **including `subject_cmp`**, or drop the idea and say why
- [ ] If added: it does not reject a legitimate region import (verbatim-carried keys from a foreign vocabulary are not a fixpoint of the destination's)
- [ ] If added: NULL semantics for `subject_cmp` in the unique tuple are stated explicitly, not left to Postgres defaults by accident
- [ ] If dropped: the `_Avoid_` is noted as knowingly still open, with the advisory lock named as the partial mitigation
- [ ] **If added: the index and its migration ship in this issue.** Do not leave the add branch pointing at a follow-up that does not exist

## Dependencies
- Depends on #5032 and #5035 — last in the cut by construction

msywulak:
[...]

Issue #5031, opened 2026-08-03
refactor(brain): batched role-less entity resolver, surface retention, and the abstain/failure split


msywulak:
Slice D of ADR-0037 — the entity-resolution seam. Four changes that only look separate: they are all consequences of the store being brain-owned and feeding the slot through the **vocabulary** rather than through a live resolver.

**Spec: the resolution comment on #5009 (T5) §1, §4, §6 — ⚠️ AND ITS CORRECTION COMMENT, which is authoritative where they conflict.** Also the resolution comment on #5011 (T7) §4. ADR-0037 §5 is the *why*.

> ⚠️ **Three of T5's claims are withdrawn. Reading the resolution alone will build this backwards.**
> 1. **"The subject has no use for an id" is withdrawn.** T12 (#5017) gives the subject `subject_cmp`. **The resolver serves TWO positions, not one** — batch over the deduplicated **subject and object** surfaces, not object-only. Consequently `unresolved: EntityRole[]`'s retirement **partially unwinds**: the subject *can* be unresolved again. The subject-side column itself is #5032; this issue must not design it out.
> 2. **The `object_cmp` cross-region rider is withdrawn** — see #5035. Do not build any autonomous carry-forward of a store id across regions.
> 3. **"Entity edges invent no new authority" is withdrawn** — see #5023.
> 4. The resolution's `Acme / acquired / Beta` ⇄ `Beta / acquired_by / Acme` role-invariance target is **malformed** — those subjects must *not* key identically. It is retired by construction here: **`role` is deleted from the resolver's context**, so the invariant holds by type and needs no test.
> **The store contract also gains a clause it lacked: ids must be GLOBALLY UNIQUE (ULID/UUID).** Deterministic + workspace-scoped does not imply globally unique, and derived ids (`dim_plan:7`) colliding across regions for different rows produce a false `same` — two distinct entities merged.

**The waterfall is real.** `reconcile.ts:604-645` — the `Promise.all` at `:625` pairs subject with object, but the enclosing `for` at `:604` **awaits once per candidate**. Free today only because the default returns synchronously; with a DB-backed resolver a 20-candidate episode is 20 serialized round trips before the transaction opens.

## Key files
[...]

PR #5076, opened 2026-08-06
feat(brain): subject_cmp with inverted polarity, and the review-gate widening disclosure (#5032)

### LINKED ISSUE #5032: feat(brain): subject_cmp with inverted polarity, and the review-gate widening disclosure
Slice D of ADR-0037. **Subject homonymy is a confidentiality limit, not an advisory one** — and it is the one place in this map where a private claim's **body** reaches a public audience.

**Spec: the resolution comment on #5017 (T12).** ADR-0037 §5 is the *why*.

> ⚠️ **`subject_cmp` is NOT a mirror of `object_cmp`. Its polarity is inverted.** Describing it as a mirror produces an implementation that mints tension edges between provably-different entities — and suppressing **corroboration** is the arm that actually matters here.

| | Null | Proves | Effect when proven |
|---|---|---|---|
| `object_cmp` | yes | *difference* | **enables** supersession |
| `subject_cmp` | yes | *difference* | **suppresses everything** — corroboration, tension and supersession alike |

Two claims about *different entities* are not in the same slot at all.

**Why corroboration is the consumer that matters.** `CORROBORATION_LOOKUP_SQL` (`reconcile.ts:420`) is the only identity consumer with **no grant arm and no cardinality arm**. On a hit it writes a provenance edge attaching a public episode as evidence to a private fact; publish then **overwrites `visible_to` with the union of evidence grants** (`brain-facts.ts:244` → `:307`). The safety argument for that widening is written at `promotion.ts:348-350` — *"a reader of either **already saw it said**"* — and **subject homonymy falsifies that sentence: the claim was not stated in B.** A different entity's claim was.

**This map materially widens the hazard**, which is why it is owned here rather than accepted: `lexicalNorm` newly collides `Acme Corp` / `ACME CORP` / `acme-corp` / `Acme  Corp`, distinct keys today; and an approved alias can **manufacture** a homonym that did not exist (`acme → acme corp`, workspace-wide).

**The honest limit, stated in the ADR because it is the weak point:** only a warehouse-backed subject can supply a `subject_cmp`. **The extractor can never supply one, for any subject, ever.** So the extracted↔extracted homonym — the case that occurs *today* — stays unresolvable, and is guarded by the disclosure below rather than prevented. Accepting it means accepting it **forever**: the vocabulary is a function on surfaces, and no function on surfaces maps one surface to two referents.

## Key files
- `packages/api/src/lib/db/schema.ts` + a migration — **this PR owns the `subject_cmp` column itself**; #5030 does not create it
- `packages/api/src/lib/brain/reconcile.ts` — `:443` `INSERT_FACT_SQL`, whose explicit column list must gain `subject_cmp`, fed from #5031's per-episode batch map at the subject surface; `:420` `CORROBORATION_LOOKUP_SQL`, `:457` `INSERT_PROVENANCE_EDGE_SQL`, `:482` `TENSION_CANDIDATES_SQL`, `:870-879` the wider-episode note
- `packages/api/src/lib/content-mode/adapters/brain-facts.ts` — `:244` `EVIDENCE_GRANTS_SQL`, `:307` `WIDEN_AND_PROMOTE_FACTS_SQL`, `supersessionCollisionJoin`
- `packages/api/src/lib/brain/promotion.ts` — `added` is described at `:340-346` and **computed at `:396-408`**; `:348-350` is the comment this corrects
- `packages/api/src/lib/brain/oversight.ts` — `:876-1021`, the disclosure machinery to reuse

## Acceptance criteria
- [ ] ⚠️ **This PR owns the write path, not just the reads.** It adds the migration and `schema.ts` mirror, and puts `subject_cmp` into `INSERT_FACT_SQL`'s column list fed from #5031's batch. "Resolver-supplied" names a source, not a site — #5031's acceptance criteria cover `object_cmp` writes only
- [ ] `subject_cmp` is nullable, resolver-supplied, and **inverted**: non-null and unequal ⇒ not the same slot ⇒ **suppress corroboration, tension and supersession alike**
- [ ] NULL everywhere is **byte-identical** to today's behaviour. Assert it, or the non-regression claim is unproven
- [ ] No new index — `subject_cmp` is a residual filter over what is already a handful of live rows
- [ ] `promotion.ts:348-350`'s safety comment is **corrected in the tree**; it is a claim homonymy makes false
- [ ] The **review-gate widening disclosure** fires only when `added` is non-empty (computed at `promotion.ts:396-408`). Widening fires on legitimate corroboration too, so a naive disclosure is universal — a filter that has been fooled. `promotion.ts:340-346` already computes `added` as a syntactic upper bound on readers gained, and role-implication makes most unions admit nobody
- [ ] Falsification: `subject_cmp` difference suppresses **corroboration**, not only the destructive pair — a test asserting only *"no supersession"* passes against a half-implementation that still widens an ACL
- [ ] Falsification: the suppression has a **positive control** — the same fixture with `subject_cmp` NULL must corroborate
- [ ] Falsification: the disclosure fires on non-empty `added` and **not** on an ordinary corroboration
- [ ] The unique index deferred by T3 §7 must gain `subject_cmp` or be dropped — see #5038

## Dependencies
- Depends on #5031 (the resolver must serve the subject position) and #5030
- **Accepted cost:** the residue is **visible, not prevented**. Nothing distinguishes the homonym from the real corroborations, so the disclosure may be a speed bump that gets clicked through

msywulak:
Closes #5032. Slice G of [ADR-0037](docs/adr/0037-claim-identity-in-the-brain.md) (⚠️ the issue body says "Slice D" — third letter collision in this arc; D is #5033, E is #5030, F is #5031, and the ROADMAP letters are the SSOT).

## What this is

`brain_facts.subject_cmp` — nullable, resolver-supplied, and **inverted**.

| | Null | Proves | Effect when proven |
|---|---|---|---|
| `object_cmp` | yes | *difference* | **enables** supersession |
| `subject_cmp` | yes | *difference* | **suppresses everything** — corroboration, tension and supersession alike |

Two claims about *different entities* are not in the same slot at all, so there is one arm (`subjectNotDifferentSql`) and all three consumers take it unchanged. It lives in its own module because the mistake ADR-0037 §5 names by hand — reading it as a mirror of `object_cmp` — mints `in-tension-with` edges between entities the store has just proven are different.

**Corroboration is the arm that matters, not supersession.** `CORROBORATION_LOOKUP_SQL` is the only identity consumer with no grant arm and no cardinality arm: a homonym makes a public episode *evidence* for a private fact, and publish then overwrites `visible_to` with the union of evidence grants. That discloses the private claim's **body**, and it falsifies the safety argument at `widenGrantFromEvidence` — *"a reader of either already saw it said"* — which is **corrected in the tree** in this commit.

**NULL everywhere is byte-identical to before**, asserted rather than read off the SQL (`homonym-control`).

**The honest limit is permanent.** Only a warehouse-backed subject can supply a `subject_cmp` — `subjectComparableValue` takes a resolved id and never parses the surface — so the extracted↔extracted homonym, the case that occurs *today*, stays unresolvable. It is **guarded, not prevented**, by the review-gate widening disclosure.

## The disclosure

`loadWideningPreview` — reader-scoped, gated on `widenGrantFromEvidence` returning non-null, so an ordinary corroboration between equally-granted episodes fires nothing. It runs **the transaction's own decision function**, never a SQL restatement of the grant grammar.

⚠️ **No unscoped `withheld` counterpart, and that is a stated gap.** Counting the widenings a reader cannot see needs the grant grammar over other readers' episode grants, which `oversight.ts`'s no-unscoped-content rule forbids. So an empty list means *"none that you can see"*, never *"none"* — pinned by a `-pg` test rather than left in a docstring.

## Decisions recorded per `/review-panel` Step 5

Two round-1 findings were **(b) new machinery** and were taken in-PR rather than deferred:

- **`incomplete: boolean` on the wire, beside `truncated`.** One boolean carried two facts with different remedies, so the panel had to state one of them unconditionally — a confident, specific, *wrong* explanation on the surface whose entire product is honest notice. Taken in-PR because the wrong explanation was already shipping in this diff.
- **A `WITH scoped AS (… LIMIT)` CTE bounding DRAFTS.** The projection ran unbounded on every oversight page render, sharing a `Promise.all` with two shipped disclosures, on a pool with no `statement_timeout`. The bound is on drafts and not rows because a row `LIMIT` would hand `widenGrantFromEvidence` a partial evidence list and **under-report** an ACL change while still listing it.

## Falsification

- `identity-corpus.ts` gains relation `proven-homonym` and four entries. Three hold their objects **equal** (which is what makes them reach corroboration); `homonym-rival` varies the objects because with only the equal-object rows, deleting the subject arm from `collisionCorePredicate` killed **zero** tests — a prohibition blocked by the wrong arm, caught by measuring rather than reading.
- Two positive controls, not one: `homonym-control` (no ids) separates suppression from a broken lookup; `homonym-same-entity` (the *same* id twice) is the only fixture that kills `both sides non-null ⇒ suppress`, a mutation that would silently switch corroboration off for every store-backed workspace.
- `scripts/mutations/subject-cmp.md` — generated, 8 mutations × 4 suites.
- New `subject-cmp.test.ts`; new `-pg` block executing `willWidenRowsSql` for the first time.

## Collateral this slice had to fix

- **A shared builder falsifies every substring discriminator downstream of it.** `IS NOT TRUE` moved into `collisionCorePredicate`, and **four** sites told the two held-back diagnostics apart by `sql.includes("IS NOT TRUE")` — each with a comment asserting the spelling was unique. All now discriminate on `=== TIER_HELD_BACK_COUNT_SQL`, with a throw for an unrecognised third.
- **A real bug the `--affected` lane could not see:** `promotion-pg.test.ts` calls `CORROBORATION_LOOKUP_SQL` directly with a hand-built bind array, so widening `agreementBinds` made it a bind-arity error — invisible locally because `-pg` skips without `TEST_DATABASE_URL`. Surfaced only by the mutation runner's red-baseline check.
- **Four generated mutation tables were stale on `main`** (`object-cmp.md` 58 vs 59 tests, `cardinality.md` 47 vs 70 plus a dead anchor). Regenerated. `mutation-core.md` is left alone — its anchor broke in #5060/#4389 and this branch does not touch that file.
- **The last hand-typed mutation table is promoted** to `scripts/mutations/identity-corpus.md`. Its own header predicted it would go stale on the next corpus-growing slice; it did, in five cells.

## Not in scope

The deferred unique index on `(workspace_id, subject_key, predicate_key, object_key)` does not exist in the tree. Migration 0193's header records that it must gain `subject_cmp` or be dropped — **#5038** owns that decision and its NULL-in-a-unique-tuple semantics.

---

## Panel round 4 — the round that converged

Rounds 1–3 each found defects **inside the previous round's fixes**. Round 3 raised 5 must-fixes; all are closed here, and the round-3 findings are what this section records.

### The two live defects

- **The empty widening disclosure failed OPEN.** `loadWideningPreview` is reader-scoped with no `withheld` counterpart, so an admin whose widening drafts all sit in audiences they are not part of received `{ total: 0, entries: [], incomplete: false }` — truthful about *their* scope, a false all-clear about the workspace, and publish is workspace-wide. The one sentence saying so ("none that you can see, never none") lived **inside `WillWidenNotice`**, which is exactly the component that does not render in that case: the hedge was present whenever it was redundant and absent whenever it was load-bearing. Now hedged when `hidden > 0` or `!countsConsistent`, and deliberately silent when the reader can see the whole backlog — with that silence asserted as a control, so the two positive tests cannot pass vacuously.
- **Round 1's brand guarded the parameter and stopped one function short.** `entityComparable` is exported and unbranded, so while the subject position's destination types spelled `EntityComparable`, `entityComparable(subject)` satisfied every one of them **with no cast** and reproduced round 1's defect verbatim. Round 3 measured the spelling the compiler *rejects* and missed the one it *accepts* — the panel's own blind spot, now a mutation row.

### Round 2's fixes were correct and unfalsified

Both headline round-2 fixes were right and **neither could fail**. Measured: reverting either `LEFT JOIN` to `INNER` killed **0** tests; downgrading the fact-poisoning to a per-row `continue` killed **0**. The drift fixture that claimed to cover the evidence-side drop path used `evidence_grant: null` — which the loader defines as explicitly *not* drift — so it exercised the keep path while asserting the drop path. That is this repo's recorded *"negatives drawn from a population the bug doesn't live in"* shape, under a comment claiming the opposite.

### Decisions recorded per Step 5

- **(b) new machinery, taken in-PR: the `SubjectComparable` brand.** The smallest correct fix for the bypass adds a type and threads it through three declarations — it could not be made by correcting a line. Taken in-PR because the alternative is shipping a guard whose docstring claims a property the tree falsifies, and because the defect it closes is the *same* one round 1 was opened for. Diff cost ~40 lines.
- **(a) local defects, fixed inline:** the panel hedge, the `scanCapped` hoist, the drift fixture, and the four missing falsifiers.
- **Two MEDIUMs raised independently by two reviewers, fixed inline:** `scanCapped` read `drafts.size` *after* the poison sweep, so one poisoned draft at exactly the cap disabled the detector — the very defect the `LEFT JOIN` change was made to fix, one variable later, masked because `incomplete` ORs in `droppedRows`. And `BrainFactWillWidenSchema`'s `superRefine` rode onto the **client** schema via `.optional()`, where a producer bug that can only under-state one headline instead blacked out all three disclosures; split into a base envelope, with the client keeping `strictObject` (a leak guard) and dropping the cross-check (a server-side concern).

### Falsifiers — every fix, per Step 6

| Fix | Falsifier | Before → after |
|---|---|---|
| Panel hedge | 2 web tests + a silence control | 0 → 2 kills |
| Output brand | `@ts-expect-error` pair + `subject-cmp.md` bypass row | 0 → 1 type error; 20 (`-pg`) + 5 (`reconcile`) kills |
| `LEFT JOIN` ×2 | lexical assertions in the projection test | 0 → 1 each |
| Fact poisoning | one fact, one good row + one bad row | 0 → 1 (×2 arms) |
| `null` is not drift | `incomplete === false` — the only observable that separates the readings | 0 → 1 |
| `scanCapped` | `WILL_WIDEN_DRAFT_SCAN_MAX` no-op widenings, nothing dropped | 0 → 1 |
| `superRefine` + client split | 5 schema tests | 0 → 1 each |
| `truncated` copy | positive control (the existing assertion was passing over a branch that never rendered) | 0 → 1 |

**One fix is honestly unfalsifiable and says so in the code**: the `scannedDrafts` hoist. `incomplete` deliberately fuses "capped" and "dropped", and every input that makes the two readings differ contains a drop, so it reports `true` either way. Reverting it kills 0 while the seven mutations around it kill 1 each — that measurement is in the docstring, not in anyone's head, along with the note that a future wire-level `scanCapped` should bring the fixture with it.

### Left open, deliberately

All LOW, none user-visible: `added`'s `minItems: 1` is absent from the published OpenAPI (the tuple encoding drops it — recorded at the schema, since restoring `.nonempty()` would silently make the `satisfies` vacuous again); row fan-out (drafts × provenance edges) is bounded only by the draft cap; and `mutation-core.md` is stale **on `main`** with a broken anchor, untouched by this branch. `mutate.ts --check` sits on no CI gate at all — which is why four tables were found stale on `main` — and that is worth its own issue.

**Local `/ci`: 33/33 green** on `cc302a873`, `-pg` lane live against `TEST_DATABASE_URL`.


msywulak:
## 🌙 Session handoff — stopped for the night with the PR open, **not merged**

### Where this stands

Local `/ci`: **33/33 gates green** on the committed tree (`90ebdf0d5`), including the full test suite and the real-Postgres lane (`TEST_DATABASE_URL` set against `brain_5021_scratch`). Remote CI has not been watched — that is step 1 tomorrow.

### `/review-panel` — two rounds run, a **third is owed**

The three code reviewers ran twice with fresh context. `comment-analyzer` has **not** run, so by the command's own rule this diff has **never had a comment sweep** and its verdict is *not* CLEAN.

**Round 1** (6 must-fix) — all fixed:

| Finding | Fix |
|---|---|
| `subjectComparableValue(surface)` compiled, and was **worse** than a surface parse (raw surface as payload ⇒ `entity:Acme Corp` vs `entity:acme-corp` ⇒ proven-different ⇒ corroboration off for the exact corpus pair) | `ResolvedEntityId` brand, one mint at the validation seam. Verified: passing a surface is now `TS2345` |
| `agreementBinds`' two adjacent `ComparableValue` params with opposite semantics were swappable | `EntityComparable` narrow type. Verified: the swap is now `TS2345` |
| Deleting the subject arm from `collisionCorePredicate` was **green on every default-lane suite** (measured) | arm added to `brain-facts.test.ts`'s shared-arm list; that file + `subject-cmp.test.ts` added as mutation targets |
| `willWidenRowsSql` was **never executed anywhere** | four new `-pg` tests |
| Panel discarded `truncated`, so a drift-dropped ACL change rendered a clean panel with a live publish button | split into `truncated` / `incomplete`, gate widened |
| `added` non-emptiness enforced only at runtime (zod v4 infers `string[]` from `.nonempty()`) | `readonly [string, ...string[]]` + `z.tuple([z.string()], z.string())` |

**Round 2 — every high finding was a defect *inside* a round-1 fix.** The documented four-rounds pattern held for a third consecutive slice. All fixed:

- **`drafts.size >= WILL_WIDEN_DRAFT_SCAN_MAX` was the wrong detector** — INNER joins meant it counted *drafts with evidence* against a bound on *drafts scanned*, so one evidence-less draft in the window silently disabled it. → `LEFT JOIN`, so every scoped draft yields a row.
- **A mid-fact dropped row produced a LISTED entry with an under-stated `added`** — the exact failure the CTE was introduced to prevent, reintroduced in the drop arm of the same function. → poison the whole fact.
- **The branch was RED** — three route fixtures missing `incomplete`.
- **The round-1 polarity assertion was VACUOUS** — `)) IS NOT TRUE` occurs *twice* in `CORROBORATION_LOOKUP_SQL` (`objectSameSql`'s own veto is the other), so the check was satisfied by the OBJECT arm. Measured: inverting the subject polarity in that statement killed **0** tests in the default lane. → assert the whole arm from the builder.
- **`identity-corpus.md` was stale** on the very slice that promoted it. → regenerated.

### 🔜 Start here tomorrow

1. **`gh pr checks 5076 --watch`** — remote CI has never run on this branch. Gate the merge on `mergeStateStatus`, not `--watch --required`.
2. **Run round 3 of `/review-panel`.** Round 2's fixes are unreviewed, and rounds 1→2 both found defects inside the previous round's work. Round 3 will also be the first `comment-analyzer` pass, so `CLEAN` is unreachable until it runs. **This is at the 3-round cap** — if it does not converge, that is the documented STOP-and-ask, not a licence to merge.
3. **Round-2 findings deliberately left open** (all MEDIUM/LOW, all recorded rather than silently dropped):
   - `BrainFactWillWidenSchema`'s `superRefine` rides onto the **client** schema, so a producer bug that today would only understate a headline instead wipes the entire oversight panel — defeating the `.optional()` deploy-skew defence beside it. Fix is a base/refined split (~4 lines). **Highest-value leftover.**
   - `scanCapped` has **no falsifier** — no fixture reaches 5,000 drafts.
   - The `superRefine` and the `added` tuple's runtime arm have no schema-test coverage (consistent with the pre-existing `willSupersede` posture, so a consistency gap rather than a regression).
   - The brand's compile-time property is unpinned; the repo's idiom for this is `@ts-expect-error` (`sources.test.ts:208`).
   - `truncated` with an empty list is schema-valid and renders nothing.
   - Row fan-out (drafts × provenance edges) is still unbounded even though drafts are capped.
4. **`mutate.ts --check` is on no CI gate at all** — which is why four generated tables were found stale on `main`. One line in `ci-local.sh` is the ratchet the whole #5060 investment is missing. Worth an issue.
5. `mutation-core.md` is **stale on `main`, not from this branch** — its anchor broke in #5060/#4389 and this branch does not touch `mutation-core.ts`. Left alone deliberately; also worth an issue.

### Housekeeping

Worktree `../atlas-wt-5032-subject-cmp` is **still in place** (not removed) so tomorrow's round 3 has the tree. `-pg` lane needs:

```
export TEST_DATABASE_URL=postgresql://atlas:atlas@localhost:5433/brain_5021_scratch
```

msywulak:
## Panel round 5 — CONVERGED, and the comment sweep earned its keep

Round 5 ran all four reviewers (`--final`): the three code reviewers plus **`comment-analyzer`, which had never run on this diff in five rounds**.

### The convergence signal

**No reviewer found a code defect inside round 4's work** — the first time that has happened in this arc. Specifically, the reviewers verified rather than accepted:

- The hedge gate's soundness was traced through *both* count populations (`willWidenRowsSql`'s scan vs. `awaitingReview − reviewableAwaitingReview`). They use the identical ACL clause and the identical draft predicate, so `hidden === 0 ⟹ zero ACL-hidden drafts ⟹ the scan covered the whole population`. **The gate is sound.**
- All 7 round-4 falsifiers re-measured at exactly 1 kill, independently.
- The `scannedDrafts` **unfalsifiability claim survived a deliberate refutation attempt** — a structural argument (`poisoned.add` shares a branch with `droppedRows++`, so the two readings can only differ where `incomplete` is already `true` by the other arm) plus four differential fixtures returning byte-identical envelopes. The note now carries both verifications instead of an unreproducible "seven mutations" count.

### What the comment sweep found — 4 HIGH

All four were comments the tree falsifies, and **two were stale prose left behind by round 4's own code changes**:

| Finding | Why it matters |
|---|---|
| `schemas/brain.ts` + `types/brain.ts` describe `added` as `.nonempty()` and say the `satisfies` passes vacuously | The field is a tuple *precisely so it doesn't*, and the inline comment 8 lines below says the opposite. A maintainer acting on the docstring deletes a working type guard. The `types` copy **ships in a published package.** |
| `object-cmp.ts` claims a template-literal type is "exactly as strong as the property being claimed" | That is the exact claim round 3 disproved. One file now said the shape guard was sufficient while the other said it demonstrably was not. |
| Migration 0193: "tension and supersession both require `single` on the canonical predicate" | Only supersession does. Tension is gated on the producer's LLM hint, defaulting to `multi`. Verified — `cardinalitySingleSql` never appears in the tension path. |

Plus five of my own round-4 overclaims, each corrected against the tree: *"nowhere else can a bare string become one"* (the test helper is a documented exemption), *"cannot be built"* (the `null` abstain is constructible by design), *"`drafts.size` IS the scanned count"* (not an equality), and two test comments whose stated measurements **measured otherwise** — notably that the `-pg` suite carries the LEFT-JOIN behavioural half. It cannot: it seeds through `landCorroboratedPair`, so every draft carries a provenance edge and the distinction is structurally invisible there. The lexical assertion is the whole guard and now says so.

### Three code changes, each 0 → 1 falsifier

- **Both `willWiden` gates now consult `total`.** `{ total: 5, entries: [] }` is schema-valid — the cross-check forbids only the opposite skew — and rendered no notice *and* an affirmative "no widening was found".
- **The hedge's `!incomplete` conjunct had no falsifier**; without it an incomplete scan renders "may widen more than Atlas can list" directly above "no widening was found".
- **The draft scan cap now logs.** Every other degradation in `oversight.ts` does, and this is the one that says *Atlas stopped looking*, on the only pre-publish ACL disclosure.

### The lesson, recorded in ROADMAP

`/review-panel` defers `comment-analyzer` to the final round so its output is not invalidated by the code fixes it triggers — which is correct, and it is why rounds 1–4 didn't run it. **The cost is that a diff can carry a false comment through four rounds**, and here two of them would have told the next maintainer to delete a working guard. Worth weighing against the deferral rule.

**Local `/ci`: 33/33 green** on `b0d890788`, `-pg` lane live. Verdict: **CLEAN** — a comment sweep has now run on this diff, and nothing must-fix survives it.

Issue #5035, opened 2026-08-03
feat(residency): bundle v3 — keys travel verbatim, both _cmp columns are nulled at import


msywulak:
Slice E of ADR-0037. The rule that covers both row-copy paths:

> **A row-copy path carries keys verbatim; a claim-supply path never supplies them.**

**Spec: the resolution comment on #5015 (T10) §1 and §2.** ADR-0037 §8 is the *why*.

**Keys travel verbatim.** The importer has no re-derive precedent to point at — every `brain_facts` column travels verbatim (`export.ts:293-304` ⇄ `admin-migrate.ts:1122-1148`), the sole exception being `fts`, a *generated* column. The failure directions settle it: carrying fails to **under**-match (recoverable — a missed flag, #5000 at a region boundary); re-deriving fails to **over**-match (a destination alias the source lacks merges imported facts into a slot they never belonged to, and publish stamps `valid_to` across the merge — irreversible).

> ⚠️ **T5's resolution comment contains a rider saying a store id can travel because *"an id that resolves nowhere still compares correctly for equality — which is all `object_cmp` is for."* THAT CLAIM IS WITHDRAWN** — see the correction comment on #5009 and T10 §2. `object_cmp` has **two** jobs and the second is destructive: *different* = both non-null and unequal, which is the arm `supersessionCollisionJoin` becomes. **A foreign region's id is non-null and, by construction, unequal to every locally-minted id — it is counterfeit positive evidence of difference.**
>
> W migrates EU→US. Published, live, `single` fact P: `Business tier / bundled plan / Enterprise Plan`, `object_cmp = ent_eu_412`, imported verbatim. The US producer later mints `ent_us_907` for the **same real entity**, canonical surface `Enterprise tier`. Draft D keys `enterprise tier` → `ent_us_907`. `object_key` differs — precisely the case the store exists to bridge. Both `object_cmp` non-null and unequal → **different** → promoting D **stamps `valid_to` on P**. Without the migration both sides would have been `ent_us_907` → **same** → corroboration.
>
> **A verbatim-carried `object_cmp` is strictly worse than NULL.** NULL → `unknown` → tension → a human sees it. A foreign id → `different` → an autonomous stamp. **Verbatim carry inverts the abstain band at the one position that has one.**

## Key files
- `packages/api/src/lib/residency/export.ts` — `:293-304`
[...]

Issue #5020, opened 2026-08-03
feat(brain): materialize keys at the reconcile seam and repoint the three identity consumers


msywulak:
Slice A of ADR-0037. The keys exist after #5019 but nothing writes them on new rows and nothing joins on them. This closes both halves in one unit — writing keys without repointing is dead weight, and repointing without writing breaks every new fact.

**Spec: the resolution comment on #5006 (T2) and on #5007 (T3) §2, §6.** ADR-0037 §1 is the *why*.

Identity is decided **at the reconcile seam and is source-agnostic** — producers supply claims, never matching rules. Keys are materialized at write time (T2), never re-derived by a consumer. The `alias` lookup is a seam here; it defaults to identity until the vocabulary lands in slice B, which is exactly why a green suite on this PR proves very little.

Three consumers, all of which silently no-op today on a phrasing mismatch — this is the table in ADR-0037's Context:

| Consumer | Module | What silently no-ops |
|---|---|---|
| `TENSION_CANDIDATES_SQL` | `lib/brain/reconcile.ts` | the advisory `in-tension-with` edge |
| `CORROBORATION_LOOKUP_SQL` | `lib/brain/reconcile.ts` | provenance-strengthening — mints a duplicate row instead |
| `supersessionCollisionJoin` | `lib/content-mode/adapters/brain-facts.ts` | promote-time supersession, the oversight disclosure, **and** the publish preview |

## Key files
- `packages/api/src/lib/brain/reconcile.ts` — key computation before `INSERT_FACT_SQL`; `TENSION_CANDIDATES_SQL` (~`:482`) and `CORROBORATION_LOOKUP_SQL` (~`:420`) swap their join arms
- `packages/api/src/lib/content-mode/adapters/brain-facts.ts` — `supersessionCollisionJoin` (~`:361-372`); note it is a column-to-column join (`p.subject = d.subject`), a different access pattern from the other two, served by the same index

## Acceptance criteria
- [ ] Every new `brain_facts` row is written with all three keys; no producer computes them itself
- [ ] All three consumers join on `subject_key` / `predicate_key` and no longer on the surface columns
- [ ] The slot keys are **total** (`NOT NULL`), and nothing introduces a nullable predicate key. A NULL predicate key would abstain *out of* tension rather than into it (`NULL = NULL` is NULL), which is a strictly worse #5000 — see T3 §2
- [ ] Each consumer's change is a swap of the join arm and nothing more — no consumer re-derives a key
[...]

Issue #5019, opened 2026-08-03
feat(brain): identity key columns, lexicalNorm, and the day-one backfill


msywulak:
Slice A of [ADR-0037](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0037-claim-identity-in-the-brain.md)'s cut — the key substrate. Adds the three `NOT NULL` key columns to `brain_facts`, the pure `lexicalNorm` that fills them, the repointed slot index, the day-one backfill, and the guards that keep the new columns honest. **Zero observable behaviour change is the expected outcome**: with an empty vocabulary `alias` is identity, so every key equals `lexicalNorm(surface)`.

**Spec: the resolution comment on #5007 (T3) §3, §6, §7, and the resolution comment on #5012 (T8) §3.** ADR-0037 §1 is the *why*; read the resolutions for the *what*. Neither is re-litigated here.

`lexicalNorm` is: case-fold, unify separators (`_`, `-`, whitespace → single space), trim, collapse runs. **Nothing else.** No stemming, no lemmatisation, no copula- or stopword-stripping — `led_by` and `leads` are both live in the corpus and are *inverse* relations that any stemmer collapses into one slot, and over-match at a join arm is the irreversible direction.

## Key files
- `packages/api/src/lib/db/migrations/0187_*.sql` — new: `ADD COLUMN` nullable → `UPDATE` → `SET NOT NULL`, on `0092_pillar_install_id_columns.sql:55-79`'s model (a real `ADD COLUMN` → `UPDATE` → `SET NOT NULL` precedent) with a scale note in the header. **Not `0169_convert_notebook_conversations.sql`** — that migration is data-only and its header says no DDL runs there
- `packages/api/src/lib/db/schema.ts` — mirror the columns in the same PR; the `brain_facts` `pgTable` starts at `:3258` and the surface-form docstring is at `:3264-3265` (the surface columns' "must hold an unresolved surface form" docstring stays true)
- `packages/api/src/lib/brain/` — new module for `lexicalNorm` (pure, total, offline, no model/network/clock/locale)
- `scripts/check-brain-fact-promotion.sh` — extend the UPDATE-gated column set

## Acceptance criteria
[...]

PR #5104, opened 2026-08-10
feat(brain): make the slot keys total — tighten MALFORMED_CLAIM, then SET NOT NULL

### LINKED ISSUE #5047: feat(brain): make the slot keys total — tighten MALFORMED_CLAIM, then SET NOT NULL
`SET NOT NULL` on `brain_facts.subject_key` / `predicate_key` / `object_key` is the third and last step of ADR-0037 §1's identity key. #5019 added the columns nullable, #5020 keyed `INSERT_FACT_SQL` and pivoted the three slot consumers onto them. The constraint is what makes a NULL key mean one thing again.

**The source of truth is the ⚠️ block in `packages/api/src/lib/db/migrations/0187_brain_fact_identity_keys.sql`.** Read it before starting — it enumerates the prerequisites and the two caveats on the backfill re-run, and it is deliberately more careful than this issue body.

## Why it is not in #5020

#5020's acceptance criteria claimed the flip. It was cut from that PR because `identityKey` returns NULL for a surface whose lexical norm is empty (`-`, `___`, `  `), and `reconcile.ts`'s `MALFORMED_CLAIM` guard tests `surface.trim() === ""` — `String#trim` strips whitespace but not `_` or `-`. Those are storable claims **today**, so `SET NOT NULL` would not merely fail on legacy rows: it would turn an ordinary ingest into a not-null violation that fails the whole reconcile transaction.

A NULL key currently means two things — "no writer has keyed this row yet" (transient) and "this surface asserts nothing" (permanent, legal). The constraint's job is to eliminate the first; the guard's job is to eliminate the second.

## Key files

- `packages/api/src/lib/brain/reconcile.ts` — the `MALFORMED_CLAIM` guard in the preparation loop, and the `log.warn` #5020 added beside the fact insert, which is the current signal for a permanently-unkeyed claim
- `packages/api/src/lib/brain/identity.ts` — `identityKey`'s ⚠️ states the overload this closes
- `packages/api/src/lib/db/migrations/` — the constraint migration, plus a repeat of 0187's backfill (0188 is the precedent and explains the two caveats)
- `packages/api/src/lib/brain/__tests__/identity-pg.test.ts` — asserts `is_nullable = 'YES'` today with a message pointing here; it flips with the constraint

## Acceptance criteria

- [ ] `MALFORMED_CLAIM` refuses a candidate whose `slotKey` is null, applied **post-resolution** (the key derives from the resolved subject/object, not the raw surface). Note there is a second degeneracy check on the canonical in the same file
- [ ] The refusal reuses the existing `MALFORMED_CLAIM` block reason — a claim whose subject norms away asserts nothing, which is the argument the guard already makes
- [ ] `correct_fact`'s supersede verb inherits the fix for free (its replacement routes through `reconcileFacts`, so the block rolls the `valid_to` stamp back with it) — pin that with a test rather than assuming it. Today a degenerate replacement against a real target passes `replacementIdentical` and installs a successor nothing can ever corroborate, contradict, or supersede
- [ ] 0187's backfill is re-run immediately before the constraint, and the re-run's two caveats are respected: it is a **pre-vocabulary** operation, and `subject_key IS NULL` already means two things, so count unkeyed rows against the **expression**, never by testing the column
- [ ] `SET NOT NULL` on all three columns
- [ ] `identity-pg.test.ts`'s nullability assertion flips, and its message stops pointing at an open issue

## Dependencies

- Depends on #5035 — the region import's 18-column INSERT names none of the key columns (`admin-migrate.ts`), so the constraint lands as a hard failure on every import until the v3 bundle carries keys verbatim (ADR-0037 §8). This is a **carry**, not a re-derive
- Depends on #5020 (merged) — `INSERT_FACT_SQL` keys new rows, and migration 0188 already repeated 0187's backfill once, so the only unkeyed populations left are region imports and degenerate surfaces

## Out of scope, but adjacent and unowned

Neither blocks the constraint; both are the same "an unkeyed row is invisible" theme and would stop mattering once it lands:

- There is **no runtime count** of live unkeyed facts in a workspace. The publish gate's will-supersede disclosure reports "nothing to supersede" for an unkeyed draft without being able to say the check could not run — a positive verdict where the check was skipped. A counter behind `loadSupersessionPreview`, or a periodic fiber warn, would give it one
- A **rolling deploy's own overlap** re-opens the window a backfill migration closes: the migration commits on the new instance while the N-1 instance's extraction fiber is still writing through the old `INSERT_FACT_SQL`. The backfill re-run in this issue is what sweeps that residue, which is another reason to sequence it immediately before the constraint

msywulak:
## Summary

Closes #5047. The third and last step of ADR-0037 §1's identity key: `brain_facts.subject_key` / `predicate_key` / `object_key` become `NOT NULL` (migration 0194). #5019 added the columns nullable, #5020 keyed `INSERT_FACT_SQL` and pivoted the three slot consumers, #5035 made the region bundle carry keys verbatim. This supplies the prerequisite migration 0187's header said nobody owned, and flips the constraint.

**The guard.** `reconcile.ts`'s `MALFORMED_CLAIM` now refuses a candidate whose `slotKey` is null, applied **post-resolution** so it sees the vocabulary's answer and an inherited slot — neither of which is visible from the raw surface. It reuses the existing block reason: a claim whose subject norms away asserts nothing, which is the argument the guard already makes for a blank one. The post-insert warn #5020 added is gone; it described a row that can no longer exist.

**`correct_fact` inherits the fix** through `reconcileFacts`, so the block rolls the `valid_to` stamp back with it — pinned against a live transaction in `candidates-pg.test.ts` rather than assumed.

**Migration 0194** re-runs 0187's backfill immediately before the constraint, respecting both caveats its header attached to the repeat — each inert when 0188 shipped and live now. The statement is **vocabulary-aware** (`alias` is a real table since #5022, so the raw norm would key a row into a different slot from every sibling the vocabulary unified) and **per-column** (0187's single `UPDATE` rewrites all three keys on a row unkeyed at one position, silently reverting curated keys elsewhere).

### Two decisions the issue did not specify

1. **Legacy rows whose surfaces normalize away** key to NULL permanently, so `SET NOT NULL` would have failed at boot on any install holding one. They are **tombstoned with a per-row placeholder**. The tombstone is the load-bearing half — all three slot consumers require `invalidated_at IS NULL`, so the row is outside every join by the tombstone rather than by its key, which is where a NULL left it. The placeholder is per-row, so it is not the shared sentinel 0187's header rejects. Cost, stated plainly: those rows leave `searchBrain` and the review queue. They were unreviewable, and the surfaces are retained.
2. **A region import now refuses** rather than landing a fact whose key did not arrive while its surface keys fine. That row is repairable by the next drift re-key; tombstoning it retires a healthy belief nothing restores, and re-deriving its key locally is the over-match §8 forbids. A single such row blocks a whole tenant migration — deliberate, since at that position it is evidence the bundle's identity is untrustworthy, not one odd claim.

### Adjacent fix, found on the way

`migrate.ts` had no `notice` listener, so **every `RAISE NOTICE` in every migration was being discarded** — including the operator breadcrumbs 0032, 0034, 0055, 0072 and 0085 emit deliberately, one of which names the gap it was written to close. 0194 adds a sixth and is what surfaced it.

## Test Plan

Docker is unavailable in this environment, so a local Postgres 16 was started rather than ship the migration untested. **Full suite: 1052/1052 files pass** against a real database, including every `-pg` suite. This surfaced 10 files of keyless fixtures the `--affected` graph does not reach.

Falsifiers were **run**, not reasoned about — each mutant applied, observed red, reverted:

- drop the vocabulary join from 0194 → the aliased-key fixture fails
- collapse 0194's per-column scoping → the curated-key fixture fails
- revert the importer to "always tombstone" → the refusal case fails
- drop `-` from `SEPARATOR_RUN` → the placeholder-collision falsifier fails
- gate the correction 400 on position instead of cause → the vocabulary-caused case fails

- [x] Manual testing — migration replayed against a seeded pre-0194 corpus
- [x] Unit tests added/updated
- [ ] E2E tests added/updated — not applicable

## Checklist

- [x] Types pass (`bun run type`)
- [x] Tests pass (`bun run test`) — 1052/1052
- [x] Lint passes (`bun run lint`)
- [ ] Deps consistent — no dependency changes
- [x] Labels added (type + area) — inherited from #5047
- [ ] CHANGELOG.md updated — `/release` owns the per-tag changelog

## Review

Two rounds of the internal panel. Round 1 returned ~21 findings, all new surface; round 2 returned ~17, of which ~13 were defects inside round 1's own fixes. That ratio is the loop's stop condition, so the loop was closed there rather than buying a third round. Everything confirmed is fixed.

The sharpest round-2 finding is worth naming because a fresh-context check caught it: round 1's fix for *"don't blame the wrong party"* blamed the wrong party one layer over. `applySupersede` gated its user-facing 400 on the failing **position**, reasoning the object is the caller's own text. But `slotKey` is `identityKey(alias(identityKey(surface)))`, so an object key is also null when the workspace's vocabulary maps a real norm to nothing — and a human superseding with good text was told their replacement "normalizes away to nothing", on a request no retry could satisfy. The block now carries the **cause**, not just the position.

### Known follow-ups, not in this PR

Deliberately deferred under the same stop rule, since each adds machinery a third round would then have to review:

- the reconcile warn's `cause` / `degenerateSurfaces` / `inheritedUnkeyed` payloads have no test — the repo has a `*-logging.test.ts` pattern for exactly this
- 0194's per-column scoping is pinned at one of three positions; the subject and object statements' `slot_position` literals are not independently falsified
- the `RegionImportUnkeyableError` → 409 route mapping is untested, and the importer's *computed* (v1/v2) tombstone arm has no case
- `scripts/check-brain-fact-promotion.sh` cannot see `.sql`, so 0194's `invalidated_at` write passes by exclusion. Recorded as a known hole in that script rather than left silent

---
_Generated by [Claude Code](https://claude.ai/code/session_014ctgWpuG7pXR4zauszr6X3)_

### DIFF
--- a/.claude/commands/review-panel.md
+++ b/.claude/commands/review-panel.md
@@ -174,6 +174,25 @@ The shapes worth sweeping for, because they are what actually recurred:
   look — and if a test landed on one, the other is where it is missing.
 - **The other caller.** A guard added at one call site of a shared helper is
   usually missing at the rest.
+- **The PROSE COPY of the same claim, including generated and public ones.** A
+  finding about *what the code says to a human* — a message naming the wrong
+  cause, a remedy that cannot work, an error blaming the wrong party — almost
+  always has a twin outside the code: an OpenAPI `description`, a docstring, a
+  migration header, a `RAISE NOTICE`, a route's response contract. Fixing the
+  code and leaving the contract is fixing the instance.
+
+  ⚠️ **Generated artifacts are the half that gets missed, because the sweep is a
+  grep over source and the copy lives in a build output.** #5047 spent a round
+  removing wrong-subsystem blame from a refusal — then regenerated
+  `apps/docs/openapi.json` and found the same blame sitting in the 409's public
+  `description`, naming one of two causes and one remedy that is unfollowable for
+  the other. That description is what a caller actually reads, so the defect had
+  survived in the one place it mattered most. It was caught by a drift gate
+  failing for an unrelated reason, not by the sweep.
+
+  So when a finding is about wording, grep the generated surfaces too — and if
+  the fix changes a route, a schema or a migration, REGENERATE before deciding
+  the class is closed.
 
 Report the sweep in one line per finding — *"same shape at X, Y; fixed"* or
 *"swept the module, no other instance"* — so a later round can see the class was
--- a/.claude/commands/ship-issue.md
+++ b/.claude/commands/ship-issue.md
@@ -124,7 +124,77 @@ So the local pre-flight is the cheap subset — `--affected`, `lint`, `type` —
 - you are touching `scripts/mutations/**` or the mutation gate itself, which remote CI does not exercise the same way;
 - you are about to `/release`, where the mutation gate and the full serial battery are the point.
 
-⚠️ **Never kill `ci-local.sh` mid-run.** `mutate.ts` rewrites source files in place and reverts them at the end, so an interrupted run leaves a MUTATED source file in the tree — silently, and it will be committed by the next `git commit -o` that names it. If you must stop one, `git status` afterwards and restore anything it left behind.
+⚠️ **THE MUTATION-GATE TRIGGER ABOVE IS WRITTEN ON THE WRONG THING, and this is
+the one pre-flight gap that actually costs remote round-trips.** `mutation-tables`
+does not break when you touch `scripts/mutations/**`. It breaks when you touch
+anything a spec **TARGETS** — and the commonest way to do that is to **RENAME OR
+REWRITE A TEST**, which goes nowhere near that directory. The generated tables
+record *"first test to die"* BY NAME plus a count per suite, so renaming a test
+makes them stale, and a refactor that moves the code a mutation anchors on kills
+the anchor outright.
+
+So add one cheap command to the pre-flight whenever the branch renamed a test,
+deleted a function, or reshaped a block a mutation might anchor on:
+
+```bash
+TEST_DATABASE_URL=… bash scripts/check-mutation-tables.sh --affected origin/main
+```
+
+It reports which specs the branch touched and exits instantly when the answer is
+none, which is the common PR. **It is NOT free when the answer is non-zero** —
+it re-runs whole `-pg` suites once per mutation, tens of minutes — so run it in
+the BACKGROUND and carry on; do not block the loop on it, and do not put it in
+the default pre-flight.
+
+Measured on #5047: the branch renamed tests across five suites and deleted one
+function. `mutation-tables` failed remote CI **twice**, and the second failure
+cost a full CI round-trip that a background `--affected` run started at PR time
+would have pre-empted. Three anchors were dead, and one of them had been
+re-anchored once already by the previous issue for the same reason.
+
+⚠️ **A DEAD ANCHOR IS NOT ALWAYS RE-ANCHORABLE, and reaching for a replacement
+mutation is how a tombstone gets into a table.** If the PR deleted the code a
+mutation targets, there is nothing to move the anchor to: DELETE the mutation and
+say why. Do not invent a successor pointing at whatever replaced it unless that
+successor is genuinely killable — a defensive arm for a state your change just
+made unreachable measures `0` in every suite, and `mutate.ts`'s header calls a
+published `0` a claim. Recording one asserts *"no test covers this"* where the
+truth is *"no input reaches this"*, which is precisely the tombstone the
+dead-anchor arm refuses to write. Point the replacement at the guard that made
+the state unreachable instead — that one has a test and a real number.
+
+⚠️ **NEVER COMMIT WHILE A MUTATION RUN IS LIVE — and that includes obeying a
+hook that tells you to.** `ci-local.sh`, `check-mutation-tables.sh` and a bare
+`mutate.ts` all rewrite source files in place and revert them at the end, so a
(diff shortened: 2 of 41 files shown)


msywulak:
The "Known follow-ups" section above now has issue numbers, plus two the closing sweep added:

- #5105 — the `MALFORMED_CLAIM` warn payload has no test, and `degenerateSurfaces` logs raw claim text (safe only while the `cause` filter is right, and nothing guards that)
- #5107 — 0194's per-column scoping is pinned at one of three positions; the subject and object statements' `slot_position` literals aren't independently falsified
- #5108 — the `RegionImportUnkeyableError` → 409 mapping is untested, and the importer's computed (v1/v2) tombstone arm has no case
- #5106 — the import's 500 echoes the raw pg error message
- #5109 — the drift re-key's `IS NOT NULL` arm means `rekeyed: N` can no longer distinguish "nothing drifted" from "N placeholder rows were declined"

The fourth item in the body — `check-brain-fact-promotion.sh` cannot see `.sql`, so 0194's `invalidated_at` write passes by exclusion — is **not** filed. It's recorded as a known hole in that script itself rather than tracked as work, since closing it means teaching the guard a second language.

Two commits landed after the body was written: `6ebdcf461` repairs 14 comments this branch's own changes made false (three files still cited a warn the branch deleted), and `b256b370f` closes the `/ship-issue` gaps this run exposed.

---

Issue #5077, opened 2026-08-06
ci: gate `mutate.ts --check` — generated mutation tables go stale silently


msywulak:
The generated mutation tables under `packages/api/scripts/mutations/*.md` are on **no CI gate at all**. `scripts/mutate.ts <spec> --check` exists and works; nothing runs it.

## Why this matters

Four tables were found stale **on `main`** during #5032 (`object-cmp.md` 58 vs 59 tests, `cardinality.md` 47 vs 70 plus a dead anchor). A stale table is not cosmetic — the whole #5060 investment exists so a kill count is *measured* rather than hand-written, and a table nobody re-runs is back to being a hand-written claim that happens to be formatted like a measurement.

`mutation-core.md` is stale on `main` right now, and worse: its `escapeCell` row is an **ANCHOR failure** — *"anchor not found in `scripts/mutation-core.ts` — the source moved out from under the spec"*. It broke in #5060/#4389's pipe-conditional escape rewrite. So the runner's own table is the one that went stale, which is the sharpest possible demonstration of the gap.

## The catch that makes this non-trivial

⚠️ **Running `mutate.ts` WITHOUT `TEST_DATABASE_URL` silently zeroes the `-pg` column and rewrites the table.** During #5032 this clobbered `subject-cmp.md` — real kill counts (3, 1, 1, 37, 29, 3, 66) all became `0` and the suite size dropped 76 → 6, with no warning. A CI gate that runs `--check` without a live Postgres would therefore report *stale* on every table forever, and a naive "just regenerate it" fix would commit the zeroed version.

So the gate needs either a live `-pg` service, or `--check` needs to learn to skip/flag `-pg` columns when `TEST_DATABASE_URL` is unset rather than silently treating them as zero. **The second is probably the real fix** — the silent-zero behaviour is a footgun independent of CI.

## Key files

- `packages/api/scripts/mutate.ts` — the runner and its `--check` mode
- `packages/api/scripts/mutations/*.mutations.ts` / `*.md` — the specs and generated tables
- `scripts/ci-local.sh` — where the gate would go (one line, alongside the other `check-*.sh` gates)
- `.github/workflows/` — the CI counterpart

## Acceptance criteria

- [ ] `mutate.ts --check` no longer reports a false result when `TEST_DATABASE_URL` is unset — it either skips `-pg` columns or fails loudly, but never silently records `0`
- [ ] Regenerating without a live Postgres cannot overwrite real `-pg` numbers with zeros
[...]

PR #5115, opened 2026-08-10
ci: take the two heavy gates off every PR's critical path

msywulak:
Both sweeps **move**, neither is deleted. `push: main` still runs the full mutation sweep and an unconditional image scan; image-scan keeps its Monday cron. What changes is only when a heavy job **blocks a PR**.

## `mutation-tables` — `--affected` on PR, `--all` on push

The job had **no `if:`, no `needs:`, no path filter**. ~14 min of Postgres-backed sweep on every PR, including docs-only ones — and the required `ci` umbrella has it in `needs:`, so every merge button waited on it.

Its own comment justified that:

> ~14 minutes measured for the full sweep, which is **FREE here**: jobs run in parallel and `Built image (docs)` already takes ~25, so this lands well inside the existing critical path.

That premise is conditional on a job which is **itself path-gated** (`built-images` has `if: needs.changes.outputs.built-images == 'true'`). When the image scan skips, `mutation-tables` *is* the critical path. The premise was never wrong — it was never re-checked against the other job's guard.

`--affected` is what `ci-local.sh` already runs. It widens to `--all` on any unresolvable diff, so the narrow path cannot silently verify nothing; that widen needs `fetch-depth: 0`.

**#5077's guarantee is intact.** Its finding was that `--check` existed and nothing ran it — four of eight tables stale on `main`. `push: main` is what keeps that closed.

## `image-scan` — narrow a filter that matched every code PR

The filter opened with `^packages/` — essentially every code change in this repo. Three images built and scanned for a one-line test edit. Its comment calls this "deliberately broad", but **a pattern matching everything is an absent filter, not a broad one**.

The gate detects a vulnerable OS package in a runtime layer, plus library findings in the image's resolved `node_modules`. A TypeScript edit moves neither — it compiles into a layer *above* the base, and the closure Trivy reads is decided by the lockfile and manifests. New trigger set: any `Dockerfile`, any `package.json`, `bun.lock`, `deploy/`, and the scanner's own files.

⚠️ **`(^|/)package\.json$`, not `^package\.json$`.** The old anchor caught the **root** manifest only; workspace manifests rode in on the `^packages/` blanket this removes. Dropping the blanket without widening that would have opened a real hole.

Two backstops already in the file and unchanged: `push: main` scans unconditionally (last chance before Railway) and the Monday cron scans an unchanged tree — the only thing that catches a CVE disclosed against a digest-pinned base.

## Verification

- Regex checked against **12 must-match** paths (all 11 Dockerfiles' dirs, root + workspace manifests, `bun.lock`, `.trivyignore`, scanner scripts) and **7 must-not-match** (source under `packages/`, `ee/`, `plugins/`, `apps/docs` content, `.claude/`, `README.md`). All pass.
- Both workflows parse as YAML; `bash -n` clean on the script.
- ⚠️ **This PR is the test.** `--affected` only runs on a `pull_request` event — pushing straight to `main` would exercise only the `--all` branch and prove nothing. Watch `mutation-tables` here: it should select the affected specs (or report none) instead of sweeping all eight, and `built-images` should **skip**, since this diff touches no Dockerfile, manifest or lockfile.

Test Plan
- [x] Unit tests added/updated — n/a, CI config
- [x] Manual testing — regex validated against 19 paths; YAML + shell syntax
- [ ] E2E tests added/updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -521,6 +521,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
+        with:
+          # Full history: `--affected` diffs against the PR base SHA, which is
+          # not reachable from a shallow clone. The script widens to `--all` if
+          # the diff fails, so a shallow clone would silently cost 14 minutes
+          # rather than break — which is exactly the failure nobody notices.
+          fetch-depth: 0
 
       - uses: ./.github/actions/setup-bun-cached
         with:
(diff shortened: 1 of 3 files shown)


msywulak:
⚠️ **Correction to the PR body.** It predicts `built-images` should **skip** here. That is wrong, and the run proves it: `Built image (api/web/docs)` are all running.

The reason is correct behaviour, not a filter bug — this diff touches `.github/workflows/image-scan.yml`, which the filter deliberately matches (`the scanner itself`). A change to the scanner **must** rescan; that entry is doing its job.

So this PR cannot demonstrate the `built-images` skip. The skip is demonstrated by the next PR that touches only source — which, before this change, was every PR in the repo.

`mutation-tables` is still the live test here, and it is unaffected by that.

msywulak:
## Measured, and it answers "what else should we skip?" — nothing

`mutation-tables` on this PR: **1m8s**, against **15m** on the last completed `main` run. The `--affected` path selected work and finished; `--all` still runs on `push: main`.

Full job costs from run `31416230654` (`main`, completed), which is the reason to stop here:

| Job | Time | Guard |
|---|---|---|
| **mutation-tables** | **15m** | none → this PR |
| Built image (api) | 4m | path-gated |
| test-others | 3m | none |
| api-tests ×4 (parallel) | ~2m40s each | none |
| Built image (docs) | 1–2m | path-gated |
| type / lint-type-aware / drift / build | ~1m | none |
| lint / test-e2e-integration / ci | <1m | none |
| Deploy Validation: Scaffold, Boot Smoke, Standalone | skipped | already gated |

`mutation-tables` was **~4× the next-most-expensive job** and essentially the entire critical path. Everything below it is ≤3m and runs in parallel, so gating any of it behind `changes` (~13s, plus a dependency edge and a fail-open branch) is **negative value**. Deploy Validation already gates its three expensive jobs correctly.

⚠️ **One more falsified number, and it is the load-bearing one.** The `--all` justification read:

> `Built image (docs)` already takes ~25 [minutes]

Measured today: **1m on `main`, 2m5s here.** So the "14 minutes is FREE" argument rested on a figure that is wrong by more than 10× — on top of being conditional on a job that is path-gated. Neither half of the premise survives measurement.

Same lesson this repo keeps recording about mutation tables and comments — *a stated measurement is a claim, and it goes stale silently* — one layer up, about CI wall clock.

Issue #5196, opened 2026-08-14
feat(brain): the enrollment surface — a human names the (entity, dimension) pairs the warehouse producer may emit


msywulak:
**Decided in [ADR-0039](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0039-the-warehouse-producer-emits-only-what-a-human-enrolled.md) (M5 kickoff grill).** The tier-1 warehouse producer (#5042) emits one fact per row per dimension at machine cadence, and every fact lands as a `draft` needing human publish — `reconcile.ts:777`, where migration 0180's default *is* the review gate. A sweep-mode producer therefore puts an unreviewable queue behind the one gate the product is differentiated by: ten thousand accounts across eight dimensions is eighty thousand drafts.

**Enrollment is the bound.** A human names the `(entity, dimension)` pairs the brain should hold claims about; the producer emits for those and only those. An unenrolled dimension is not hidden, filtered, or pending — it is **outside the producer's reach**, and the coverage surface reports it that way.

This issue owns the storage and the authoring surface. #5042 owns the producer that reads it.

**The precedent to build against is [ADR-0032](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0032-amendments-refine-never-grow.md)**, not a new pattern: semantic-layer Amendments refine coverage that exists and never expand the whitelist; an uncovered column is shown honestly as *uncovered* and routes to the enrich flow — a human-initiated act with consequences. Same shape, same reason.

## Key files
- `packages/api/src/lib/brain/` — new module for enrollment storage + reads (the producer's input)
- `packages/api/src/db/schema.ts` + a new migration — the enrollment table (mirror in the same PR, per `.claude/rules/api-database.md`)
- `packages/api/src/lib/db/bundle-scope.ts` and its test — the classification entry
- `apps/web` admin — the authoring surface, alongside the Claim Vocabulary surface (#5025)
- `apps/docs` — `data-residency.mdx` "What moves" table

## Acceptance criteria
- [ ] An enrollment is a `(entity, dimension)` pair, workspace-scoped, internal-DB-resident
- [ ] **The table, its `bundle-scope` classification, and its export/import wiring land in ONE PR.** Per the #5026 lesson: `bundle-scope.test.ts` fails both a table with no registry entry and an `exported` entry with no `FROM`/`INSERT INTO`. Splitting them makes a real cycle. Classify alongside the vocabulary tables and add the row to `data-residency.mdx`
[...]

Issue #5043, opened 2026-08-03
feat(brain): entity store and its entity-edge producer


msywulak:
> ### ⚠️ Updated 2026-08-13 at the M5 kickoff grill — read before starting
>
> **The store's COVERAGE is now bounded by enrollment.** Per [ADR-0039](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0039-the-warehouse-producer-emits-only-what-a-human-enrolled.md) the warehouse producer (#5042) emits only for `(entity, dimension)` pairs a human enrolled — there is no sweep mode. Since this store is materialized by a producer *"the same way warehouse facts arrive"*, it resolves surfaces for **enrolled entities and abstains elsewhere**.
>
> **That is not a new failure mode — abstention is already this store's designed behaviour** (absent-means-abstain, fails closed). But it has a consequence worth stating before anyone builds it: **an empty store and a correctly-working store are indistinguishable from inside the code.** Every read abstains in both cases, and every test can pass in both cases. This is why M5 closes on prod row counts (#5197) rather than on merge, and why the falsification below must include a positive control that actually resolves something.
>
> **Do not "fix" the partial coverage.** Widening the store beyond enrollment means either read-time matching — prohibited by name in ADR-0037 §5 and restated in this issue's own acceptance criteria — or a sweep, which ADR-0039 rejects.

**Scoped into M5 alongside #5042.** The entity store is designed in full by ADR-0037 §5 and built by nobody — the map put it out of scope. Without it `subject_cmp` and the entity-valued half of `object_cmp` are permanently NULL.

**Spec: §3 and §5 of the resolution comment on #5009 (T5) — ⚠️ AND ITS CORRECTION COMMENT.** ADR-0037 §5 is the *why*.

**The store is brain-owned, workspace-scoped, internal-DB-resident.** The semantic layer's glossary and entities are **inputs**, never the store — they are type-level where the brain names instances, `connection_group`-scoped where the brain is workspace-scoped, and unjoinable as one opaque YAML blob. **General rule, derived twice: the brain never reads tier-1 live, at any position, for any purpose.**

**The store's contract, four properties:**
1. **Emits vocabulary edges.** Every entity names a canonical surface; the edge is `lexicalNorm(alias surface) → lexicalNorm(canonical surface)`, an ordinary vocabulary row. **Its slot-side contribution *is* an alias edge, so the store is not consulted at reconcile time for the slot at all.**
[...]

PR #5226, opened 2026-08-14
feat(brain): the tier-1 warehouse fact producer — enrollment-scoped, fail-closed, tension-only

### LINKED ISSUE #5042: feat(brain): tier-1 warehouse fact producer — the component ADR-0037 designs protections for
**Scoped into M5 by T9 §3 (#5013).** ADR-0037 ships protections for a tier-1 warehouse producer that **no milestone in ADR-0036's M1–M6 cut had committed to building**. Naming the milestone converts *"we designed for it"* into *"we will exercise it."*

**Spec: §2 of the resolution comment on #5008 (T4); §3(d) of #5010 (T6); §3 of #5009 (T5).** ADR-0037 §4 is the *why*. **Its REACH is [ADR-0039](https://github.com/AtlasDevHQ/atlas/blob/main/docs/adr/0039-the-warehouse-producer-emits-only-what-a-human-enrolled.md)** — read that first.

**A large fraction of ADR-0037 is dormant until this exists** — warehouse-wins does nothing mechanical, `subject_cmp` is permanently NULL, entity edges have no source, and the alias-proposal query returns zero rows. The ADR states this as its own uncomfortable half.

---

> ### ⚠️ Updated 2026-08-13 at the M5 kickoff grill — read before starting
>
> **This issue was `ready-for-human` because three things were undecided. All three are now decided, and it is `ready-for-agent`.**
>
> **1. Which entities it produces from — the ENROLLED ones.** Per ADR-0039 the producer is **enrollment-scoped**: a human names `(entity, dimension)` pairs and the producer emits for those and only those. There is **no sweep mode**, and the producer may never widen its own scope. The reason is arithmetic, not taste — every fact lands as a `draft` needing human publish (`reconcile.ts:777`, where migration 0180's default *is* the review gate), so a sweep puts an unreviewable queue behind the one gate the product is differentiated by. Ten thousand accounts × eight dimensions is eighty thousand drafts. **The enrollment storage and authoring surface are #5196; this issue reads them.**
>
> **2. When it runs** — on enrollment, and on a cadence thereafter. Cadence stops being frightening once scope is human-chosen.
>
> **3. Snapshot discipline** — unchanged from ADR-0037 §4, and now affordable.
>
> **The "resolve before building" blocker below is RETIRED, not answered.** The old text called warehouse↔warehouse re-emission a *"known unresolved decision"* that had to be settled before any code. That framing was already stale when written: ADR-0037 §4 rules it *"tension-only, which is the Fog #5008 records rather than a gap"* — a recorded, accepted consequence. It was never an unspecified design hole; it was a **volume** problem wearing a specification problem's clothes. At enrollment scale a re-run mints a handful of tension edges, which is a queue a person drains. **Do not open a design cycle on it.**

---

**The emission contract, already decided — do not redesign it:**
- **The bare dimension name, measure name, or metric id.** Not `plans.price`, not `analytics.plans.price`. Entity and connection-group qualification rides in `provenance.detail` as **non-load-bearing** context. A qualified surface can never lexically match anything an LLM emits, so qualifying it makes day-one cross-tier collision count exactly zero
- **Fails closed on ambiguity.** Same subject, two entities, same dimension name — `Acme Corp / status / active` from `accounts` and `Acme Corp / status / signed` from `contracts` — both read as `different`, the irreversible direction. **The producer refuses to emit when a dimension name is ambiguous across the entities it is producing from.** A missing warehouse fact is recoverable; a wrong `valid_to` stamp is not. *(Under ADR-0039 this now evaluates across the ENROLLED set, which makes a refusal legible to the person who caused it rather than emitted into a sweep nobody is watching.)*
- **A warehouse predicate becomes collidable by being an ordinary surface on an ordinary stored row.** Live tier-1 never enters the model — only the snapshot-pinned `brain_facts` row is collidable
- **Declares cardinality structurally** — a dimension of one row is `single` by construction. This is authoritative, not a hint, and is one of only three sources that may put `single` in the vocabulary
- **Supplies `object_cmp` and `subject_cmp`** where the extractor cannot: its subject is a dimension *value* backed by a warehouse row with a primary key
- **No reserved roots** — a warehouse norm may itself be aliased away. Anyone reasoning *"the warehouse key is the column name"* is wrong after the first retarget

## Key files
- `packages/api/src/lib/brain/sources.ts` — `:326-330`, the episode source kinds
- `packages/api/src/lib/brain/reconcile.ts` — the seam it supplies claims to, like every other producer
- The enrollment read from #5196 — the producer's input set

## Acceptance criteria
- [ ] The emission contract above is implemented as specified by the ADR rather than by whoever builds it
- [ ] **The producer's input is the enrollment set and nothing else.** No code path emits for an unenrolled `(entity, dimension)` pair, and no code path enrolls
- [ ] Falsification: the producer **refuses** on an ambiguous dimension name — **with a positive control** emitting on the same run's unambiguous sibling. Both the safe behaviour and a producer that emitted nothing at all look identical to a test that only asserts "no bad row appeared"
- [ ] Cross-group conflation is understood as pre-existing, not introduced here: two connection groups each with a `price` dimension produce one key whether or not anything is seeded, because vocabulary and keys are workspace-scoped
- [ ] Warehouse↔warehouse re-emission behaves as ADR-0037 §4 already specifies — tension-only, symmetric. **Add no new mechanism for it.** A test pins that a re-run over a changed value produces a tension edge and does **not** stamp `valid_to` — a machine invalidating a fact is forbidden by #4759 §2

## Dependencies
- ⛔ **Hard prerequisite: #5196 — the enrollment surface.** The producer has no input set without it
- ⛔ **Hard prerequisite: #5033 — the tier guard must exist inside `supersessionCollisionJoin` before this producer writes a single row.** ADR-0037 *creates* the condition under which a draft can stamp `valid_to` on a published warehouse-derived fact; shipping the producer first opens that window deliberately. **Shipped in M4**
- Depends on #5019, #5020, #5022, #5023, #5027, #5030, #5031, #5032 — the whole of slices A–D, named rather than left as prose

msywulak:
## What this is

The **tier-1 warehouse fact producer** — the component ADR-0037 spent four slices building protections for and that no milestone had committed to building. Until now warehouse-wins did nothing mechanical, `subject_cmp` was permanently NULL, entity edges had no source, and the alias-proposal query returned zero rows. ADR-0037 states that dormancy as its own uncomfortable half; this is the half that wakes it up.

`lib/brain/warehouse-producer.ts` reads the enrollment set (#5196), snapshots each enrolled entity, and files the claims through the same reconcile seam every other producer uses.

## The emission contract, as specified

| ADR-0037 §4 says | Where it lives |
|---|---|
| The **bare** dimension name as the predicate; qualification non-load-bearing | `buildWarehouseClaims` — `detail` carries entity/table/group; the predicate is `dim.name` |
| **Fails closed** on a dimension name ambiguous across the entities it produces from | `planWarehouseEmission` — refuses **both** pairs, names the other entity in the message |
| Declares cardinality **structurally** | every candidate carries `predicateCardinality: "single"`; the authoritative half is a `warehouse_structural` proposal, `pending` |
| Supplies `subject_cmp` where the extractor cannot | `warehouseRowId` → the episode's `EntityResolver` → `subject_cmp` |
| A warehouse predicate is an ordinary surface on an ordinary stored row | one snapshot episode per entity, `locator`-only; live tier-1 never enters the model |
| No reserved roots | nothing special-cases a key by origin |

**Reach is ADR-0039's**: a human enrolls `(entity, dimension)` pairs and the producer emits for those and only those. The snapshot query names the enrolled columns and nothing else, so an unenrolled column never leaves the warehouse.

## Acceptance criteria

- [x] The emission contract implemented as the ADR specifies — table above
- [x] **The input is the enrollment set and nothing else.** No code path emits for an unenrolled pair, and no code path enrolls (`enrollment-writers.test.ts` unchanged and green — this module only reads)
- [x] **Falsification: refuses on an ambiguous dimension name, with a positive control.** The two entities are given **different** numbers of surviving dimensions (1 and 2), so a fix that swapped or collapsed them cannot pass on an accidental equality. A `-pg` sibling shows the enrolled dimension emitting while an unenrolled column *present in the same snapshot rows* does not
- [x] **Cross-group conflation is pre-existing, not introduced** — two entities in *different* connection groups emit one predicate while their groups differ in `detail`, which is why the ambiguity rule must refuse rather than qualify
- [x] **Warehouse↔warehouse re-emission behaves as §4 already specifies; no new mechanism.** The `-pg` suite pins the tension edge **and** that neither row is stamped or tombstoned, with an unchanged-re-run corroboration test as the control

## Review panel — the curve, and why it stopped

**Round 1: 35 findings, all new surface.** 14 must-fix classes closed.

**Round 2: 34 findings — 12 new surface, 22 defect-in-prior-fix.** Defects-in-prior-fix exceeded new surface, so the rounds stopped there on the ratio rule rather than the 3-round cap: past that point more review adds work faster than it removes it. Everything confirmed was fixed, the comment sweep ran on this diff, and every falsifier was built and measured.

Between the rounds, four `fix-vs-finding` checks ran against the riskiest fixes. **Three came back REPRODUCED** — round 1's fixes were right about their instance and wrong about the class:

| Fix | Verdict | What it had actually done |
|---|---|---|
| SQL-gate relocation | REPRODUCED | Moved the hole — `async () => ({valid: true})` satisfied the new seam exactly as a substituted runner satisfied the old one |
| Post-commit response | REPRODUCED | The degraded body said `refusals: []` — a confident all-clear, worse than the 500 it replaced |
| Absent/unsurfaceable split | REPRODUCED | `""` filed as an enrollment mistake; that is what CSV/ETL loads write for an empty text column |
| `entity-unreadable` arm | CLEAN | The remedy varies at the construction site, so the shared reason code is not the same defect |

**Falsifiers: 15 mutations applied and measured, 14 RED.** The survivor is explained rather than open — switching the response schema to `z.discriminatedUnion` independently repaired the drift diagnostic, so "parse the response schema instead of the report schema" stopped being a meaningful mutation. The property that matters is still pinned.

**Comment sweep: 12 false claims, corrected.** Five were rot from two late reversals. One was the accidental-equality defect *recurring*: a comment asserting every counter differs, over a fixture where the two counters that **are** the split it exists to pin were both `1`. That comment now records that it has been wrong twice.

## Two decisions reversed late, on purpose

1. **A transaction failure refuses its entity instead of propagating.** The original decision was stated before its consequence was traced: the throw reaches `runEffect` as `500 "Failed to run"` while earlier entities have **committed**, so the admin retries and every committed entity files a second full round of drafts. That is verbatim the argument `checkedRun` makes one layer up. A defect in the module's own contract still propagates, via its own error class.
2. **The SQL gate moved out of the shipped runner into the run loop**, then had its verdict branded — because relocating it alone left the invariant enforced by convention one field over.

## Two bounds the ADR's arithmetic assumes away

Both fail-closed, both surfaced as typed refusals rather than silence.

1. **The subject is the entity's declared primary key**; no key or a composite one is refused. Anything else is a guess, and a wrong subject is a homonym — which `subject-cmp.ts` calls a confidentiality limit, not an advisory one. Honest consequence, stated in the module: a **surrogate-integer** key collides with its own re-emissions and nothing else. #5043's entity store is the slice that gives such rows a human surface, and is designed to be fed from exactly these rows.
2. **A row cap that refuses rather than truncates.** Enrollment bounds *dimensions*, not *rows*. `WAREHOUSE_ROW_CAP = 1000`; over it the entity is refused, because a truncated snapshot looks at rest exactly like a complete reading of the table.

**Enrolled MEASURES are refused with their own reason.** The enrollment surface offers them and this slice emits per-row claims, so there is no subject an aggregate could attach to.

## Deferred, deliberately — follow-ups on #5042

- **Brand the snapshot REQUEST, not just the verdict.** The verdict brand stops a seam claiming a pass; it does not stop token replay (validate one benign statement, reuse the token) and it does not stop a future call site reaching `runSnapshot` directly — the validate-then-run ordering is still enforced by statement order. The fix is ~15 lines but changes `WarehouseSnapshotRunner`'s signature and three harnesses. The type's docstring says plainly that it is closed against literals, not against ordering.
- **A checked-in `warehouse-producer.mutations.ts` spec.** The 15 mutations above were measured by hand; CI cannot reproduce them. This module is where a spec would earn the most.
- **A cadence trigger.** `POST /api/v1/admin/brain-enrollment/produce` is the only trigger that ships, under the same owner/admin bar as the enrollment writes. A schedule is a second trigger with its own enablement and audit questions — the deferral `alias-proposal.ts` records for its own producer.

## Tests

`warehouse-producer.test.ts` (45) · `warehouse-producer-pg.test.ts` (6, real Postgres) · `warehouse-producer-logging.test.ts` (8, per-level sinks so a demotion goes red) · `warehouse-producer-bypass.test.ts` (2, the bypass allowlist) · `/produce` route coverage incl. the 403 bar · `proposePredicateCardinalityForSurface` in `cardinality.test.ts`.

`scripts/mutations/cardinality.md` regenerated — suite size 38 → 41, no kill count moved.

Closes #5042


msywulak:

The three deferrals recorded in the body above are now filed, so they are tracked rather than living in a merged PR nobody re-reads:

- **#5228** — the cadence trigger. ADR-0039 decided *"on enrollment, and on a cadence thereafter"*; only `POST /produce` shipped. Milestoned into M5, since the ADR clause belongs to that milestone. Carries the questions the deferral was actually about: concurrency (no lock today, and two runs take two snapshot instants so `ON CONFLICT` dedupes nothing), enablement, and whether an on-enrollment trigger re-snapshots the whole entity.
- **#5229** — the checked-in mutations spec. The 15 falsifiers here were hand-measured; `mutation-tables` sweeps only `scripts/mutations/`, so CI cannot reproduce them. Against the standard #5061 finished establishing three days ago.
- **#5230** — brand the snapshot REQUEST rather than the verdict. Closes the two holes the verdict brand does not: replay (a cached token authorizes a different statement, and it compiles with no cast) and ordering (`defaultRunSnapshot` still takes a bare request; validate-then-run is enforced by statement order). ⚠️ Explicitly coupled to #5228 — a cadence trigger is the second call site, and the one that runs unattended.

Issue #5266, opened 2026-08-16
fix(db): the datasource catalog seeder reports a blocked row as "preserved"


msywulak:
Disclosed as residue in PR #5260 (#5239) and verified against `main` at `0a977f8da`. #5239 fixed the **knowledge** catalog seeder; its **datasource** sibling has the same defect and reports it with a more misleading label.

## The gap

`seed-builtin-datasource-catalog.ts:492` still uses an unqualified `ON CONFLICT DO NOTHING`. That swallows a conflict on ANY constraint, not just the intended id — so a row whose `slug` is held under a foreign id is silently not written.

The reporting is the worse half. `preservedSlugs` is computed by subtraction at `:499` — expected rows minus inserted rows — so a **blocked** row is indistinguishable from a **deliberately preserved** one and lands in `preservedSlugs`. The caller is then told the row is fine. #5239's pre-fix knowledge seeder at least reported an ambiguous count; this one asserts a specific, wrong, reassuring fact.

The docstring at `:16` and `:431` states the unqualified target as intentional ("covers both the `slug` ... "), so a reader has to disprove a comment to find the bug.

This is the sibling-producer pattern: fixing one writer leaves an unbranded sibling that reopens the same defect class.

## Approach

Mirror what #5260 landed on the knowledge side, rather than inventing a second policy:

- `ON CONFLICT (id) DO NOTHING` so a slug held under a foreign id raises `23505` instead of no-op'ing
- Per-row catch classifying on **SQLSTATE, never the message**; warn naming the row; record it as blocked; continue
- A `23505` naming a non-slug constraint, and every other error, must propagate
- `preservedSlugs` must stop absorbing blocked rows — blocked is its own outcome

`PG_UNIQUE_VIOLATION` reached a 4th copy in #5260; consolidating it is in scope here.

## Key files

- `packages/api/src/lib/db/seed-builtin-datasource-catalog.ts` — `:492` target, `:499` subtraction, `:16`/`:431` docstrings
- `packages/api/src/lib/db/seed-builtin-knowledge-catalog.ts` — the merged reference implementation

## Acceptance criteria

- [ ] A row whose slug is held under a foreign id is reported as blocked, never as preserved
- [ ] Classification is on SQLSTATE; a `23505` on a non-slug constraint propagates
- [ ] The `:16` and `:431` docstrings no longer describe the unqualified target as intentional
- [ ] `PG_UNIQUE_VIOLATION` has one definition, not four
[...]

Issue #5239, opened 2026-08-15
fix(brain): the built-in knowledge catalog seeder reports success for a row it never wrote


msywulak:
Split out of #5082 (PR #5235), where it was found and deliberately deferred rather than fixed inline.

## The defect

`seedBuiltinKnowledgeCatalog` (`packages/api/src/lib/db/seed-builtin-knowledge-catalog.ts`) inserts with an **unqualified** `ON CONFLICT DO NOTHING`, so the conflict target covers the `slug` unique index as well as the `id` primary key.

If any row already holds one of the built-in slugs under a **different id**, the built-in insert does nothing, `rows.length === 0`, `insertedSlugs` stays empty, and the pass returns `{ kind: "seeded", inserted: false }` with an `info` log listing every slug — for a row that does not exist under its canonical id and never will.

"Not inserted" is currently overloaded across *"present and correct"* and *"blocked, and the row is missing"*, with nothing distinguishing them.

## Why it compounds

Any migration keyed on the canonical id then correctly finds nothing. #5082's migration 0201 is the live example: it would report zero rewritten for a row that is simply absent, and `/admin/knowledge` would never list the row either. Two silences stacking.

The window is narrow — `slug` is settable only at create (`lib/integrations/catalog-crud.ts` has no `slug` in `buildCatalogUpdateSql`), so it needs an operator-created row that predates the built-in ever being seeded — but #5235 added a fixture on the grounds that an operator can create a catalog row through the CRUD path and `name` is not unique, so the shape is production-reachable.

## Key files

- `packages/api/src/lib/db/seed-builtin-knowledge-catalog.ts` — the `INSERT … ON CONFLICT DO NOTHING` and the outcome reporting. The file header already records this as a known blind spot (added by #5235).
- `packages/api/src/lib/db/__tests__/seed-builtin-knowledge-catalog.test.ts`

## Two candidate fixes

1. **Qualify the target** — `ON CONFLICT (id) DO NOTHING`. A slug collision under a foreign id then raises `23505` instead of silently no-op'ing. This is the honest fix, but it needs a **per-row try/catch that logs and continues**, because the loop is sequential and one collision would otherwise abort every remaining row.
2. **Probe after the pass** — one `SELECT id FROM plugin_catalog WHERE id = ANY($1)` and `log.warn` on anything missing. Costs a round trip per boot; does not change insert semantics.

(1) is preferred — it is not the "extra boot-time round trip" the original deferral note described.

[...]

Issue #5000, opened 2026-08-03
bug(brain): conflict detection is exact-string on predicate, so LLM phrasing variance leaves contradictions coexisting silently


msywulak:
Found in the **prod soak of M2/M3** on 2026-08-03, immediately after `v0.2.3`. Two contradictory facts about the same subject now sit in the prod brain with nothing marking them as in tension.

## What happened

Same channel, same audience, ~5 minutes apart, both extracted from live Slack:

```
draft | Business tier | priced at    | 499 a month
draft | Business tier | is priced at | 599 a month
```

No `in-tension-with` edge. The predicates differ by one word — a copula the extractor added on the second call — so the collision lookup saw two unrelated claims.

The control worked in the same batch: `Series A / has target raise of / 6M` vs `8M` produced a correct tension edge, because that time the LLM happened to emit a byte-identical predicate.

## Root cause

`lib/brain/reconcile.ts` — both lookups are raw string equality:

```sql
TENSION_CANDIDATES_SQL:    AND subject = $2  AND predicate = $3  AND object <> $4
CORROBORATION_LOOKUP_SQL:  AND subject = $2  AND predicate = $3  AND object  = $4
```

There is no normalization beyond `candidate.predicate.trim()`. `subject` is exact-match too, so casing or article drift ("the Business tier") defeats it identically.

**So M2's entire conflict apparatus is gated on the extractor reproducing a byte-identical subject AND predicate across independent LLM calls** — and it just proved it doesn't. Corroboration has the same exposure in the opposite direction: a rephrased restatement mints a duplicate row instead of a second provenance edge.

## Why the test suite cannot see this

Every M2 conflict test hand-writes both sides of the collision, so the predicates agree **by construction**. The fixtures encode the assumption under test. No amount of unit coverage reaches it — only two independent extractions of the same claim do.

## Severity: it fails in the expensive direction

Over-flagging costs a reviewer a glance. Missing a conflict leaves two contradictory facts both looking authoritative, both publishable, and the answer the agent gives depends on retrieval order. For a product whose pitch is *authoritative* company facts, a silently-held contradiction is the worst shape.

## Key files
- `packages/api/src/lib/brain/reconcile.ts` — `TENSION_CANDIDATES_SQL`, `CORROBORATION_LOOKUP_SQL`, and the `trim()` at ~606
- `packages/api/src/lib/brain/extract.ts` — where a canonicalization step would belong if the fix goes upstream
[...]

Issue #3167, opened 2026-06-04
Enforce static-bot routing-id uniqueness atomically (close the concurrent-install race)


msywulak:
## Context
Surfaced by CodeRabbit + Codex during review of #3163 (PR for #3154 GAP 2).

The static-bot install handlers (telegram/discord/gchat/whatsapp/teams) gain a cross-workspace routing-id ownership pre-check (`assert*UnboundElsewhere` — a `SELECT … WHERE config->>'<id>' = $1 AND workspace_id <> $2`) before persisting. This narrows the cross-tenant window but is **not transactionally fused** with the cap-gate UPSERT: the gate's advisory lock is keyed by `workspace_id`, and `workspace_plugins_singleton` is unique only on `(workspace_id, catalog_id)`. Two *different* workspaces installing the **same** routing id concurrently can both observe no row and both UPSERT.

The inbound resolvers in `lib/chat-plugin/executeQuery.ts` then see `rows.length > 1` and fail closed for **both** workspaces — no data leak, but an availability/griefing outage for those two workspaces until an admin disconnects one side.

This residual is documented inline in every guard ("Read-only pre-check … first writer wins") and was the "full ownership-proof flow" #3154 referenced as deferred. #3154 shipped the pre-check (its stated deliverable); this issue tracks closing the race.

## Key files
- `packages/api/src/lib/integrations/install/{telegram,discord,gchat,whatsapp,teams}-static-bot-handler.ts` — the `assert*UnboundElsewhere` guards.
- `packages/api/src/lib/billing/enforcement.ts` — `checkChatIntegrationLimitAndInstall` (the advisory-locked UPSERT).
- `packages/api/src/lib/chat-plugin/executeQuery.ts` — the read-side resolvers that fail-close on duplicates.

## Acceptance criteria
- Concurrent installs of the same routing id by two workspaces cannot both succeed.
- Options: a partial UNIQUE index on `(catalog_id, (config->>'<routing_id>'))` WHERE `enabled` + pillar='chat' (per-platform routing key), OR a routing-id-scoped advisory lock spanning the pre-check + UPSERT. Surface the conflict as the same actionable "already connected elsewhere" error.
- `my_customer` (gchat self-install alias) must stay exempt — it is caller-relative, not a global id.

## Dependencies
Follow-up to #3154 / #3163. No blocker.

PR #5276, opened 2026-08-16
fix(db): unique-violation classification was dead on every Effect path — five sites, measured (#5272)

### LINKED ISSUE #5272: fix(db): two routes classify a unique violation on a FLAT SQLSTATE, but internalQuery wraps the driver error under .cause

Surfaced by the #5271 review panel (type-design axis, HIGH) and verified as far as reading allows; the settling experiment needs a `-pg` harness and is named below.

## The gap

Two routes classify a Postgres `unique_violation` by reading a **flat, top-level** `code` off the caught error:

- `packages/api/src/api/routes/admin-prompts.ts` — `isUniqueViolation`, which turns a duplicate collection name into a 409
- `packages/api/src/api/routes/sub-processor-subscriptions.ts` — the `kind: "duplicate"` arm, which turns a duplicate webhook URL into a 409

Both write through `internalQuery`. Once the Effect Layer has booted — i.e. always, in production — `internalQuery` takes the `_sqlClient` branch (`lib/db/internal.ts:768-773`) and rejects with a `FiberFailure` wrapping `SqlError` wrapping the pg `DatabaseError`. There is **no top-level `code`** on that shape. `lib/integrations/install/routing-id-conflict.ts` walks the `.cause` chain for precisely this reason, and its docstring says so explicitly.

If that holds, both 409 arms are dead in production: a duplicate collection name and a duplicate sub-processor URL each 500 instead.

## Why the tests do not see it

`admin-prompts-audit.test.ts` and `sub-processor-subscriptions.test.ts` both set `err.code = "23505"` directly on a hand-built rejection. The fixture agrees with the flat assumption by construction, so it cannot discriminate the two shapes — the same class as #5000 and the `-pg` half of #5266.

## The settling experiment

In a `-pg` test with `_sqlClient` set, insert a prompt collection whose name collides case-insensitively through the same `internalQuery` seam, catch the rejection, and print `Object.keys(err)` alongside `err.cause?.code`. If `err.code` is `undefined`, the classification is dead.

## Approach

If confirmed, `lib/db/pg-errors.ts` gains a SECOND named classifier that walks `.cause` — not a replacement for `asUniqueViolation`. That module's header already argues the two must stay distinct: a chain walk applied to the seeders would classify a wrapped violation from an unrelated layer as a benign collision, and a flat read applied to the Effect path classifies every real collision as an unhandled throw. Two named classifiers, each used on the path it models.

## Key files

- `packages/api/src/lib/db/pg-errors.ts` — where a `asWrappedUniqueViolation` would live
- `packages/api/src/lib/db/internal.ts:768-773` — the `_sqlClient.unsafe` branch
- `packages/api/src/lib/integrations/install/routing-id-conflict.ts` — the existing chain walk to mirror
- `packages/api/src/api/routes/admin-prompts.ts` · `sub-processor-subscriptions.ts` — the two call sites, each carrying a ⚠️ note pointing here

## Acceptance criteria

- [ ] A `-pg` test establishes which shape `internalQuery` actually rejects with when the Layer has booted — the answer is recorded either way, because "we assumed flat" is what produced this
- [ ] If wrapped: both routes classify on the wrapped shape and their 409s are proven reachable by a test that seeds a REAL conflicting row rather than hand-building the rejection
- [ ] If flat after all: the ⚠️ notes in both routes and in `pg-errors.ts` are corrected rather than left standing as an unresolved doubt
- [ ] The two hand-built-rejection fixtures are replaced or supplemented so they can tell the two shapes apart
- [ ] Sweep the other `internalQuery` callers that classify on SQLSTATE — `starter-prompts/favorite-store.ts` and `suggestions/approval-store.ts` read a flat `code` too

## Dependencies

None. The ⚠️ notes referencing this issue merged in #5271.

msywulak:
#5272 asked whether `internalQuery`'s rejection carries a flat SQLSTATE. **It does not**, and the real shape is worse than the issue guessed. Measured against real Postgres through a real `PgClient.layerFromPool` client:

```
FiberFailureImpl                      ← own props: message, name, stack
  [Symbol(effect/Runtime/FiberFailure/Cause)] = { _tag: "Fail", error: SqlError }
      SqlError.cause = DatabaseError { code: "23505", constraint, detail }
```

No top-level `code` **and no `.cause` on the wrapper** — the Cause is symbol-keyed. The flat read missed it; a plain chain walk missed it too. Every affected arm turned a routine duplicate into a 500.

## Five sites, not four

The issue named four. `persist-form-install.ts` awaits `internalQuery(...).catch(raiseWriteError)` — a **promise** catch — so `isRoutingIdUniqueViolation` was equally blind, and #3167's concurrent-install race guard could never fire on that path: the workspace losing the race got a raw 500 instead of "already connected elsewhere". Stopping at the issue's list would have missed it.

| Site | Was returning |
|---|---|
| `admin-prompts.ts` | 500 instead of 409 on a duplicate collection name |
| `sub-processor-subscriptions.ts` | 500 instead of 409 on a duplicate webhook URL |
| `starter-prompts/favorite-store.ts` | 500 instead of `DuplicateFavoriteError` |
| `suggestions/approval-store.ts` | 500 instead of the duplicate-suggestion path |
| `install/routing-id-conflict.ts` | 500 instead of #3167's actionable message |

## The fix

One shared `unwrapFiberFailure` in `pg-errors.ts` used by both classifiers, so the two don't each grow a copy of the same first step. `asWrappedUniqueViolation` unwraps then walks `.cause` — it is the **sibling** of `asUniqueViolation`, not a replacement: the raw-`Pool` seeders must keep the flat read, or a wrapped violation from an unrelated layer would classify as a benign collision.

`routing-id-conflict.ts` gets the unwrap in front of its existing walk. **Additive by construction** — unwrapping only lets the walk see more links, and the `CHAT_ROUTING_ID_UNIQUE_INDEX` check still gates what counts, so it cannot start mislabelling an unrelated `23505`.

## Why no existing test could see this

All five sites' tests hand-build the rejection as `Object.assign(new Error(...), { code: "23505" })` — flat *by construction*, so they passed whether or not production produces that shape. Fixtures agreeing with the assumption they exist to check. That is why this needed a `-pg` harness rather than an argument, and it's the same class as #5000 and #5266's `-pg` half.

`internal-query-error-shape-pg.test.ts` records the answer so nobody re-runs the experiment: the SQLSTATE is **not** on the rejection, **not** on its `.cause` chain, reachable **only** through the symbol-keyed Cause, and classified by the new helper — plus a `42P01` through the same transport that must still not classify.

## Falsifiers — three, each measured RED, each independently necessary

| Mutation | Result |
|---|---|
| drop the FiberFailure unwrap | 1 RED |
| drop the `.cause` walk after it | 1 RED |
| drop routing-id's unwrap | 1 RED |

## Verification

12/12 affected files green, **plus every registry-reached consumer run explicitly** — five chat handlers, both catalog seeders, both stores, both routes, `persist-form-install`. `--affected` under-reports here, as it did on #5239. `lint`, `type`, `lint:type-aware` at 0 errors.

Closes #5272

### DIFF
--- a/packages/api/src/api/routes/admin-prompts.ts
+++ b/packages/api/src/api/routes/admin-prompts.ts
@@ -10,7 +10,7 @@ import { Effect } from "effect";
 import type { Context as HonoContext } from "hono";
 import { createRoute, z } from "@hono/zod-openapi";
 import { createLogger } from "@atlas/api/lib/logger";
-import { asUniqueViolation } from "@atlas/api/lib/db/pg-errors";
+import { asWrappedUniqueViolation } from "@atlas/api/lib/db/pg-errors";
 import { logAdminAction, ADMIN_ACTIONS } from "@atlas/api/lib/audit";
 import { runEffect } from "@atlas/api/lib/effect/hono";
 import { RequestContext, AuthContext } from "@atlas/api/lib/effect/services";
@@ -32,22 +32,13 @@ const log = createLogger("admin-prompts");
  * violation into a typed result so create/rename routes can return a 409
  * instead of leaking a generic 500.
  *
- * ⚠️ **READS A FLAT TOP-LEVEL `code`, AND THIS ROUTE WRITES THROUGH
- * `internalQuery`, WHICH MAY NOT PRODUCE ONE.** Once the Layer has booted,
- * `internalQuery` goes through `Effect.runPromise(_sqlClient.unsafe(…))`
- * (`db/internal.ts:768-773`), which rejects with a `FiberFailure` wrapping
- * `SqlError` wrapping the pg `DatabaseError` — no top-level `code`. That is
- * the shape `integrations/install/routing-id-conflict.ts` walks the `.cause`
- * chain for. If it holds here, this 409 is dead in production and a duplicate
- * name 500s instead. The existing tests cannot see it: they set `err.code`
- * directly on a hand-built rejection, a fixture agreeing with the assumption.
- * Surfaced by the #5271 review panel and NOT fixed there — it is pre-existing,
- * needs a `-pg` harness to settle, and changing four call sites' error
- * classification (two routes, two stores) is machinery this PR has no round
- * left to review. Tracked in #5272.
+ * ⚠️ #5272 SETTLED: `internalQuery` rejects with a `FiberFailure` whose Cause
+ * is symbol-keyed, so the flat read this used to do missed EVERY collision and
+ * a duplicate surfaced as a 500. `asWrappedUniqueViolation` unwraps it. Do not
+ * "simplify" back to `asUniqueViolation` — that is the raw-`Pool` sibling.
  */
 function isUniqueViolation(err: unknown): boolean {
-  return asUniqueViolation(err) !== undefined;
+  return asWrappedUniqueViolation(err) !== undefined;
 }
 
 // ---------------------------------------------------------------------------
--- a/packages/api/src/api/routes/sub-processor-subscriptions.ts
+++ b/packages/api/src/api/routes/sub-processor-subscriptions.ts
@@ -24,7 +24,7 @@ import { AuthContext, RequestContext } from "@atlas/api/lib/effect/services";
 import { hasInternalDB } from "@atlas/api/lib/db/internal";
 import { isSafeExternalUrl } from "@atlas/api/lib/sandbox/validate";
 import { createLogger } from "@atlas/api/lib/logger";
-import { asUniqueViolation } from "@atlas/api/lib/db/pg-errors";
+import { asWrappedUniqueViolation } from "@atlas/api/lib/db/pg-errors";
 import { errorMessage } from "@atlas/api/lib/audit/error-scrub";
 import { createSubscription } from "@atlas/api/lib/sub-processor-publisher";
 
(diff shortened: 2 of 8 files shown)


msywulak:
Closing as **superseded by #5279**, with the good parts carried across — not discarded.

Two sessions worked #5272 in parallel without knowing about each other. This PR was **first** (15:49 vs ~16:35) and reached the same conclusion independently, including that `routing-id-conflict.ts` was dead on the same paths — which is decent evidence the finding is real rather than one session talking itself into it.

**Why #5279 is the one that lands, and it is not "mine is better":**

- This branch is **red** on `api-tests (3/4)` and `ci`: `SyntaxError: Export named 'Runtime' not found in .../effect@3.22.0/.../esm/index.js`. That is the `import { Cause, Runtime } from "effect"` barrel form at `pg-errors.ts:51`. It type-checks, passes `lint:type-aware`, and passes `pg-errors`' own tests, then reds any suite that partially mocks the barrel while linking this file (`admin-marketplace.test.ts`). #5279 hit the same wall and fixed it with subpath imports (`effect/Cause`, `effect/Runtime`).
- Its worktree `atlas-wt-5272` has **no live session** — no process holds it as a cwd, last write 11:47. There was nobody to hand the fix back to, so the reconciliation happened on the branch that had an owner.

**What was ported from here into #5279** (commit `deac4c8b9`):

1. **`Runtime.isFiberFailure(err)`** in place of a hand-rolled symbol `in` check. Strictly better: a runtime change to how the wrapper is marked becomes a type error instead of a silently-undefined lookup, and a silently-undefined lookup here reintroduces the exact defect the PR fixes.
2. **The concrete route in** — `persist-form-install.ts`'s `.catch(raiseWriteError)`, a *promise-level* catch, so it receives the `FiberFailure` rather than the `SqlError` an in-program catch sees. #5279 never identified this and its docstring was weaker for it. Verified before copying (`persist-form-install.ts:408`, applied at 423 and 430) rather than taken on faith.
3. **The positive half of the shape assertion** — squash the symbol-keyed `Cause`, expect `SqlError`, expect its `.cause.code`. #5279 only had the negatives ("no top-level `code`", "no `.cause`"), which a refactor could satisfy by breaking the error entirely.

**What #5279 adds on top:** the subpath-import fix; a test that `MAX_CAUSE_DEPTH` is *enforced* rather than merely declared (20-link chain, assert the `23505` at the bottom is NOT found); a self-referential `cause` chain test; and a shared `pgErrorLinks` traversal so `routing-id-conflict.ts` and the SQLSTATE classifier cannot drift apart.

Falsifiers were re-measured against the merged implementation, since the earlier measurement described code that no longer existed after the port.

Nothing here is lost. Thanks to whichever session wrote it — the `isFiberFailure` guard and the `persist-form-install` trace both made the landed fix better.

> *This was generated by AI.*

Issue #5289, opened 2026-08-17
fix(db): two purge-path follow-ups from #5176, fixed in sequence


msywulak:
Two residue items from #5176 (PR #5261), both landing in `hardDeleteWorkspace` in `packages/api/src/lib/db/internal.ts`. Each was named in that PR body and deliberately not taken in-round, because #5176's brief scopes out behaviour changes: *"This is a typing and de-duplication change; the purge's behavior must be identical before and after."*

They are bundled because they are **not independent in the way #5247's four were** — they edit the same function, and two parallel workers on two worktrees would collide on one file. **Fix them in sequence, one commit per item**, so a bad one can be dropped without losing the other.

**Order is load-bearing: item 1 before item 2.** Item 2 may widen a DELETE and add a read of `stripe_purged_subscriptions` inside the same transaction — new SQL that can raise a SQLSTATE, which lands in exactly the catch block item 1 rewrites. Done in this order the new statement inherits the actionable abort message for free; done the other way round it ships behind the bare 500 item 1 exists to remove.

Re-verify both premises against `main` before starting — an issue body is a claim about the tree at authoring time.

---

## 1. A purge abort on any SQLSTATE but 42P01/42703 reaches the operator as a bare 500

Full statement of the defect, its live prod path (`23502` from the #3468 tombstone `INSERT … SELECT` when `@better-auth/stripe` has written a NULL `stripeSubscriptionId` at checkout initiation), and its five acceptance criteria are in #5265. Nothing about that issue changes here.

**Acceptance:** #5265's acceptance criteria, unmodified.

The two worth restating because they are the easy ones to lose in a bundle: the raw pg message must **not** be interpolated into the operator-facing string on the new arm (a pg error can echo column values, and this is an erasure path), and `requestId` stays on the response.

## 2. Orphan `stripe_webhook_events` ledger rows survive a workspace purge and report as complete

Full statement in #5269. `stripe_webhook_events` declares no FK to `subscription` by design, so a ledger row whose subscription row never existed for this org — or was already removed — is matched by neither the Phase-3b DELETE nor the #3468 tombstone SELECT, and the response still reports `complete: true`.

**Acceptance:** #5269's acceptance criteria, unmodified — which explicitly admit **either** outcome. Widening the delete and documenting it as accepted residue are both defensible; the deliverable is that the choice is recorded in `purge-scope.ts`'s `stripe_webhook_events` reason rather than made by omission.

[...]

Issue #5267, opened 2026-08-16
fix(agents): the review panel is specified read-only and enforced nowhere — four agents hold Bash


msywulak:
`/review-panel` specifies its agents as read-only in two places — "Each is read-only/advisory" and "Read-only. The panel reports; it never edits code" — and every agent definition repeats it ("You review and advise; you do not edit code"). Nothing enforced it.

## The gap

Four of the five panel agents held `Bash`:

| agent | grant |
|---|---|
| `comment-analyzer` | Read, Grep, Glob, **Bash** |
| `pr-test-analyzer` | Read, Grep, Glob, **Bash** |
| `silent-failure-hunter` | Read, Grep, Glob, **Bash** |
| `type-design-analyzer` | Read, Grep, Glob, **Bash** |
| `fix-vs-finding` | Read, Grep, Glob |

`Edit` and `Write` were correctly withheld, then `Bash` handed all of it back — `sed -i`, `>`, `git checkout`, `scripts/mutate.ts`. No agent body explains what Bash is for; it appears only in the frontmatter, so it is an unjustified grant, not a considered one.

**It fired.** During PR #5260 a panel agent mutated a file to measure a falsifier; its restore reverted the author's **uncommitted in-flight edits** along with its own mutant, and left the mutant behind. Caught by `git status` before any commit — but only by luck of looking. `git` cannot distinguish a reviewer's mutant from an author's unstaged edit; both are just unstaged changes. This is the same restore trap already recorded twice from `git checkout`-based mutation cleanup.

## Why the reviewers wanted to write

The motive is legitimate and does not go away: **you cannot prove a test is able to fail by reading it.** The panel's own evidence standard demands measurement — "Apply the mutant, watch it [go red]", "a `0` in a mutation table is a CLAIM, not a note". An agent holding `Bash` and asked whether the tests can fail was obeying the deeper rule by breaking the shallower one. Any fix has to answer that need, not just remove the capability.

## Approach

- Drop `Bash` from all four. The caller already passes the base ref and changed-file list, so `git diff` is not needed.
- State the split explicitly: **the panel names the mutation, the author runs it.** A reviewer that wants a measurement it cannot take reports the finding as UNVERIFIED and names the one experiment that settles it.
- A named mutation is worth more than a measured one regardless: it lands in a `scripts/mutations/*.mutations.ts` spec and keeps working, whereas a measured-and-reverted one vanishes on restore.
[...]

Issue #5037, opened 2026-08-03
fix(brain): correction.ts inherits the slot keys from its target instead of re-deriving them


msywulak:
Slice E of ADR-0037. `correction.ts` is called the one **immune** producer because it *"carries identity from the target row rather than re-deriving it from text"* — but that is true only because identity == surface today. **The instant keys are computed at the reconcile seam, `correction.ts` stops carrying identity and starts re-deriving from an inherited surface**, which is the operation ruled out for every other producer.

**Spec: the resolution comment on #5015 (T10) §3.** ADR-0037 §8 is the *why*.

**What breaks.** The stamp is id-based (`SUPERSEDE_STAMP_SQL`, `:1349`), so it **always fires**. The replacement's key is `alias_now(lexicalNorm(target.subject))`, equal to `target.subject_key` only if the vocabulary has not moved. Three ways it has: an **alias removal**; **before the drift rewrite has run**; and **an imported row**, which now carries keys from a *foreign* vocabulary. On divergence: `valid_to` is stamped, `INSERT_SUPERSEDES_EDGES_SQL` records new→old, and the replacement lands in a **different slot** — the belief retired and its successor **unreachable from the slot every future collision joins on**. The audit trail says "superseded by X"; the slot says empty.

## Key files
- `packages/api/src/lib/brain/correction.ts` — `TargetRow` at `:587-608` (surfaces only, no key slot); `:1349` the id-based stamp; `:1374-1382` where target surfaces are passed into `reconcileFacts`; `:1360-1362`'s *"the SAME seam every producer does"*; `:1383` declares `producer: "correction"`

## Acceptance criteria
- [ ] `TargetRow` gains `subjectKey` / `predicateKey`
- [ ] The candidate carries them on a **distinguished inherit channel** — typed so it cannot be filled from thin air: inherited-from-a-row-id, not free strings
- [ ] **The slot is inherited; the object is derived fresh.** The correction is *about this claim*, while the object is new and human-authored
- [ ] The inherit channel gets an allowlist entry in `check-brain-fact-promotion.sh`
- [ ] Falsification: a correction whose target was written under a **since-changed** vocabulary lands in the **target's** slot. This passes vacuously against an unchanged vocabulary, which is exactly how it would otherwise be written — the fixture must actually move the vocabulary between the write and the correction

## Dependencies
- Depends on #5020, #5022, #5023 — its falsification requires *moving* the vocabulary between the write and the correction, which needs the decide seam
[...]

PR #5298, opened 2026-08-17
chore(agents): burn the workflow layer, keep the evidence

msywulak:
Continues the harness cleanup started in #5251. Net **-6,265 lines**.

## Why

The audit that opened this branch found three false claims in the harness, and the third named the pattern:

1. `/research`'s 49-row module map had **three rows pointing into the wrong package** (`lib/semantic.ts` is a directory; `ChatErrorCode` moved to `packages/types`; `cli/bin/enrich.ts` is `lib/semantic/enrich/`) and two counts off by 3x.
2. `.claude/agents/README.md` described a tool grant that had not existed for two days — false in both halves after #5267.
3. ⭐ **`/review-panel` Step 6 contradicted itself in four places.** Its body retired the "name every round, build in the closing one" split after #5037 measured that deferral is a bet on which round is last — while its heading, cost paragraph, RUN-the-mutant rule and Rules summary all still stated the retired version.

## The diagnosis

The recurring "round 2 finds defects inside round 1's fixes" problem was audited against evidence rather than assumed. **Two suspects are ruled out.** The vendored reviewers are not drifted forks — diffed against `pr-review-toolkit` they keep upstream's full methodology and add the Atlas standards on top, and are shorter only because they drop upstream's fake `<example>` dialogues and Sentry/`errorIds` references. And the round machinery works: the ratio stop and 3-round cap fired correctly on #5148 and #5170.

What is actually happening is that **each countermeasure gets satisfied nominally, and the defect moves up a level**:

| | |
|---|---|
| #5027 | fixes shipped with no falsifier — 11 probed, 11 zeros → Step 6 added |
| #5170 | five of the author's own assertions measured inert |
| #5289 | **four falsifiers that could not fail** — a ROLLBACK-count assertion whose fixture rolls back *successfully*; a mock failing both COMMIT and ROLLBACK so neither condition was independently falsifiable |

Every rule was obeyed. More prose was the move that had already failed three times.

## What this does

- **`fix-vs-finding` gains a second object**, not a second mode: a falsifier that cannot fail is *closed the instance, left the class* applied to the instrument. Dispatched from a new **Step 6b**. It holds no `Bash`, so it **names** the mutation and the author runs it — the same panel-names/author-measures split as everywhere else.
- **Deletes 26 commands, 5 review agents and `docs/agents/`** (~6,250 lines), replaced by one page: `docs/agents/practices.md`.
- **Keeps eight operational runbooks** — `release`, `publish`, `deploy`, `ci`, `verify-mcp-cli`, `verify-prod-signup`, `dev`, `deps-update`. No finding implicated them, and `scripts/ci-local.sh`, `scripts/lib/ci-local-report.sh` and ADR-0008 cite two of them from live code.

The line, recorded in `practices.md` so the next pass need not re-derive it: **delete the practice whose enforcement was your own diligence; keep the sequence whose enforcement is an external system that will fail loudly.**

## Not a loss

`.claude/research/ROADMAP.md` holds every measured incident in **more** detail than the commands carried — #5170's round-2 composition and #5289's four dead falsifiers are both there, fuller. The commands were a lossy copy of the record, not the record.

## Verification

- No CI gate scans the deleted paths; the only references from `scripts/` are comments citing `ci.md`, which is restored.
- Restored commands re-checked for dangling pointers into the deleted 26 — one (`ci.md` → `/ship-issue`) found and removed.
- `.claude/rules/**` untouched: product-domain rules enforced by named guards, implicated by no finding.

⚠️ A fresh session no longer has the workflow commands. `docs/agents/practices.md` is the entry point; engineering skills come from `claude plugin install mattpocock-skills`.


msywulak:
**Merged `main` in and resolved three shared files with #5300.** Recording the resolution because two of the three had a correct answer that neither side held alone.

- **`CLAUDE.md`** — true same-line conflict. #5300 bumped the gate count 36 → 38; this branch removed a pointer to `/ship-issue`, which it deletes. Taking either side alone ships a defect: theirs keeps a pointer to a command that no longer exists, mine keeps a count that is now wrong. Resolved to **38 gates *and* no dead pointer**.
- **`docs/agents/workflow.md`** — modify/delete. Stays deleted; #5300's 36 → 38 edit there is superseded, and the fact survives in the two files this branch carries forward.
- **`.claude/commands/ci.md`** — auto-merged, then corrected (below).

⚠️ **#5300 updated the gate count in two of its three homes.** `CLAUDE.md` and `ci.md`'s frontmatter went to 38; the **body of that same file** still said *"all 35 of them"* and its enumeration was missing three real gates — `brain-settings-doc`, `gate-fixtures-wired`, `lighthouse-report-paths`. Counted mechanically off the `run_fg`/`launch` invocations in `scripts/ci-local.sh`, which `ci.md` itself names as the authority: **38**. The derived arithmetic was wrong in the same direction and is recomputed (`CI_LOCAL_NO_NET=1` → 36, `CI_LOCAL_NO_TEST=1` → 37, both → 35).

One fact in three places, two of them updated, **disagreeing within twenty minutes**. That is the argument for #5299 covering counts and rosters rather than paths alone — a roster that names its own authority file is mechanically checkable against it.

Issue #5297, opened 2026-08-17
chore(api): bundle 3 — an unobserved seed Layer and a refusal that outlives its log, fixed in sequence (#5273, #5112)


msywulak:
Two api-side items bundled so **one** review panel reads the combined diff. They share a theme rather than a file: each takes something whose only record today is a log line or a stub, and makes it observable. **One commit each, in the order below** (the #5289 / #5290 / #5247 pattern), so a bad one can be dropped without losing the other.

No file overlap between them, so the ordering is by size rather than dependency: the smaller harness first, so the panel's first read is the simpler diff.

## Order

1. **#5273** — the first Live-seed-Layer test harness in the repo.
2. **#5112** — refused alias edges durable at the source region.

## Members

- **#5273** — `BuiltinDatasourceCatalogSeedLive` and `BuiltinKnowledgeCatalogSeedLive` have no tests; `layers.test.ts` only ever supplies them as `Layer.succeed(...)` stubs, and a stub is not an exercise of the Live layer. The `case "seeded"` arm is the ONLY path by which `blockedSlugs` — the field #5266 exists to produce — reaches a health consumer, and replacing the three forwarded fields with `...zeroCounts` is green today. Needs a pattern that does not exist in-repo yet: stub `InternalDB` + `Migration` and intercept the dynamic `await import(...)` inside `Effect.tryPromise`.
- **#5112** — the refusal PAYLOAD (the whole edge, everything needed to re-author a dropped human decision) exists only in `mergeApprovedEdges`' `log.warn` in the TARGET region, while the SOURCE region is the one that schedules the cleanup which DELETEs its own `brain_vocabulary_edge` rows after a 7-day grace period. So the party owning the irreversible act has a count, and the record that would let anyone undo it lives in another region's log retention. Land it durably (`region_migrations` is the natural home — classified `platform`, so cleanup never deletes it), route the counts through `logMigrationEvent`, surface them on the CLI (`migrate-import.ts` never printed vocabulary counters at all), and add a correlation token so two attempts at the same bundle stop emitting byte-identical line sets.

## Key files

- `packages/api/src/lib/effect/layers.ts` + `__tests__/layers.test.ts` — #5273
- `packages/api/src/lib/db/__tests__/seed-builtin-datasource-catalog.test.ts` — the boot-wrapper forwarding test to mirror one level up (#5273)
[...]

Issue #5273, opened 2026-08-16
test(effect): the two built-in catalog seed Layers have no tests — blockedSlugs' entire production route is unobserved


msywulak:
> *This was generated by AI during triage.*

Surfaced by the #5271 review panel (test-coverage axis) in round 2. Deferred from #5271 because closing it means building the first Live-seed-layer test harness in the repo, which is new machinery arriving in a closing round with nothing left to review it.

## The gap

`BuiltinDatasourceCatalogSeedLive` and `BuiltinKnowledgeCatalogSeedLive` (`lib/effect/layers.ts`) have **no tests**. `layers.test.ts` only ever supplies them as `Layer.succeed(...)` stubs for unrelated wiring tests — a stub is not an exercise of the Live layer.

That matters more since #5266/#5239, because the Layer's `case "seeded"` arm is the ONLY path by which `blockedSlugs` reaches a health consumer:

```ts
case "seeded":
  return {
    insertedSlugs: result.insertedSlugs,
    preservedSlugs: result.preservedSlugs,
    blockedSlugs: result.blockedSlugs,
    outcome: "seeded",
  } satisfies BuiltinDatasourceCatalogSeedShape;
```

Replace those three forwarded fields with `...zeroCounts` and nothing goes red. `blockedSlugs` is the field #5266 exists to produce, and its entire production route is unobserved.

#5271 closed the identical hole one hop below, at the boot wrapper (`runBuiltinDatasourceCatalogSeedBoot`), with a test that forwards 1 blocked / 1 preserved / 7 inserted. The Layer above it kept the hole. Both catalogs are affected, so this is class-wide rather than a one-off.

## Why it is not trivial

The Layer resolves its wrapper through a dynamic `await import(...)` inside `Effect.tryPromise`, and depends on `InternalDB` and `Migration`. A test needs to stub both upstream layers and intercept the dynamic import — a pattern that does not exist anywhere in the repo yet. Getting it right once makes both catalogs testable and unblocks every future seed Layer.

## Key files

- `packages/api/src/lib/effect/layers.ts` — `BuiltinDatasourceCatalogSeedLive`, `BuiltinKnowledgeCatalogSeedLive`, and their two `Shape` types
- `packages/api/src/lib/effect/__tests__/layers.test.ts` — where the stubs live today
- `packages/api/src/lib/db/__tests__/seed-builtin-datasource-catalog.test.ts` — the boot-wrapper forwarding test to mirror one level up

[...]

Issue #5112, opened 2026-08-10
feat(residency): make a refused alias edge durable at the source, not a log line in another region


msywulak:
Follow-up to #5036, which merges two vocabularies on import: it unions the approved edges, refuses any that would close a cycle or take a second parent, logs every refusal, and recomputes the closure. That slice is scoped to **merge semantics**, and its review surfaced a gap one layer out that is genuinely new machinery rather than a defect in the merge.

## The gap

The refusal **payload** — the whole edge, everything needed to re-author a dropped human decision — exists in exactly one place: `mergeApprovedEdges`' `log.warn`, in the **target** region's process (`packages/api/src/lib/brain/vocabulary.ts`). At the seam that could make it durable, `admin-migrate.ts` reads `vocabularyMerge.refusals.length` and throws the array away, so only a **count** crosses the wire.

Meanwhile the **source** region is the one that cuts over and schedules the cleanup which DELETES its own `brain_vocabulary_edge` rows after the grace period (`CLEANUP_GRACE_PERIOD_DAYS = 7`).

So the party that owns the irreversible act has a number, and the record that would let anyone undo it lives in another region's log retention — which only barely outlives the window.

#5036 closed the actionable half of this: the source now warns with the count, names the delete timer, and points at **this region's own** `brain_vocabulary_edge` rows, which are still present for the whole grace period. That is a real remedy and it needs no new machinery. What it does not give is a durable record that survives the cleanup, or one that reaches an operator who never opens a log stream.

## Acceptance criteria

- [ ] The refusal payloads reach the **source** region — either carried on the import response (a bounded `refusalDetails` on `ImportResult`'s vocabulary section; the `_SchemaMatchesWireType` pin added in #5036 already forces the schema and the type to move together) or re-emitted from the route after `COMMIT`
- [ ] They land somewhere that **outlives the grace period** — the `region_migrations` row is the natural home: it is classified `platform`, so the cleanup never deletes it, and `cleanup.ts` re-reads it under `FOR UPDATE` at the moment of deletion
- [ ] `refused` counts reach the module's **designated audit channel** — `logMigrationEvent("region_migration_completed", …)` and `region_migration_cleanup_scheduled`, which fires for the exact 7-day timer the disclosure names. Today the disclosure is a bare `log.warn` sitting beside an audit event built for this
[...]

PR #5307, opened 2026-08-17
docs(roadmap): record bundle 3 — #5273 + #5112, and the six lessons the review produced

msywulak:
ROADMAP entry for PR #5301 (`2d53ce756`), which shipped #5273 + #5112 as bundle #5297.

Records what `/ship-issue` Step 7 asks for — `rounds: 1` with per-round agent minutes and the finding curve (28, all new surface, zero defect-in-prior-fix) — plus the six lessons the review produced, each with the measurement behind it:

1. **Widening review scope to the enclosing FILE** found a live production defect in a Layer the issue never named — the #5239 unscrubbed-`error` asymmetry for the third time, now behind a lexical guard whose negative control immediately caught the matcher flagging a *deliberately* raw producer.
2. **A scrubber can be a floor rather than a filter.** `errorMessage` rewrites `scheme://user:pass@` and nothing else, so `ECONNREFUSED 10.0.3.7:5432`, pg row fragments and column spellings survive it onto a field returned verbatim to a workspace admin. Answer: keep driver text off the field, not scrub harder.
3. **A VALUE export added to `@useatlas/types` breaks every scaffold**, and only `Deploy Validation` can see it — the scaffold installs the *published* package at `^0.7.0` while `Standalone Example Build` compiles against the local one. Distinct from T3's types-vs-schemas gotcha: this is value-vs-type in either package.
4. **Nine of the author's own comment claims measured FALSE**, plus three structural docstring defects the new code introduced.
5. **The quotation trap, twice in one PR** — a `"WILL DROP"` lookup matched the confirmation that *quotes* the phrase, so a join-key assertion compared a line to itself; root-caused to a static import hoisting a module above its logger mock.
6. **Both halves of a feature's coverage can be hand-written statements ABOUT the same columns**, so a renamed column stayed green while breaking every migration that exercises it.

Docs only — no code, no tests, no generated artifacts.


msywulak:
**Parked on the GitHub Actions outage, not on anything in this diff.**

All 24 checks are `queued` at the workflow level — GitHub has not assigned runners. The [critical incident](https://www.githubstatus.com/) opened 2026-08-17T13:40Z is still unresolved.

Not merging: `ci`, `Analyze (javascript-typescript)`, `Deploy Validation`, `fork-pr-gate` and `Image Scan` are required, and a required check that has not run is a stop sign rather than an override invitation (CLAUDE.md § Merge discipline). An agent admin-merging past it is forbidden even for a docs-only change.

Nothing to re-trigger — the workflows are queued, not failed, so they should start on their own once capacity returns. If they end up cancelled rather than started, a fresh push is what re-triggers CodeQL default setup (`gh run rerun --failed` does not work for `event: dynamic`).

Context: the work this entry records is already merged — #5301 @ `2d53ce756`, 35/35 green, closing #5273, #5112 and #5297. This PR only adds the ROADMAP record.

Issue #5233, opened 2026-08-15
bug(brain): the entity store has no reaper, and a bridge-window object_cmp may stamp valid_to


msywulak:
Two deferrals recorded during #5043's review, each with a source comment pointing here. Both turn on a question nobody has answered yet, which is why they are still here rather than in the queue.

> **Split, 2026-08-16.** This issue originally carried four deferrals. The two that need no undecided design call — the entity-edge change-detector and the `EntityEdgeOutcome` union — moved to #5277 and are `ready-for-agent`. The two below stayed, because an agent handed either one would have to invent an answer to a question this issue exists to settle.

## 1. A store reaper — entries for entities no longer in reach

`writeEntityEntries`' DELETE is what makes an entry a SNAPSHOT rather than an accumulation. It only runs inside an entity's reconcile transaction, so it is **not** reached when:

- the entity produced no candidates (a truncated table, a primary-key column that became unsurfaceable);
- the entity was refused (`snapshot-failed`, `row-cap-exceeded`, `entity-unreadable`);
- the entity was un-enrolled entirely — nothing ever deletes its entries.

The store then answers for rows that no longer exist. Recoverable (an entry resolves to an id no live fact carries; nothing over-matches, because ids are unique per row), which is why it was not fixed inline.

⚠️ **It is also what un-poisons an imported store.** `writeEntityEntries` deletes by `(workspace_id, entity)`, so if two regions' semantic layers name the entity differently (`accounts` vs `crm_accounts`), the imported rows are never deleted, both sets go live with the same canonical norms and different ids, and `resolvableIds` poisons every shared norm — the store resolves **nothing** for that entity, permanently. `entityEdgesAmbiguous` (#5043) is what makes that visible; this is what fixes it.

The design question that kept it out of #5043: what does "no longer in reach" mean when a datasource is merely down? Deleting on a transient failure is the same silent data loss the reaper exists to prevent.

## 2. Untraced: can a bridge-window `object_cmp` reach a `valid_to` stamp?

Region-imported store entries carry ids minted over the **source** workspace. Extraction consults the store, so facts written during the bridge window (before the destination's first producer run) carry those ids; that run then retires them for freshly minted ones.

[...]

PR #5322, opened 2026-08-18
feat(brain): the rename verdict, a per-entity success record, and the disagreement census (#5316, #5317, #5318)

### LINKED ISSUE #5316: chore(brain): falsify the rename trigger — does a semantic-layer rename re-mint entity ids and orphan store entries?
Parent: [issue 5233](https://github.com/AtlasDevHQ/atlas/issues/5233). This ticket decides the trigger set for the three remedy tickets in this milestone, so it goes first.

[PR 5315](https://github.com/AtlasDevHQ/atlas/pull/5315) confirmed the supersession hazard at the join: two facts about one entity whose comparables carry different ids supersede each other, even when the object surface is byte-identical. That trace framed the trigger as the **region-migration bridge window**.

There is reason to think the trigger is far more ordinary, and it has not been tested.

`entity_id` is `wh_<sha256 of (workspace, entity, primary key)>`. The entity component is `semantic_entities.name`. So **renaming an entity in the semantic layer should re-mint every id under it**, with no region migration involved. And because `writeEntityEntries` deletes by `(workspace_id, entity)`, the old name's entries are never targeted by any future run — they should orphan permanently, which is the same poisoning shape the parent attributes to a cross-region name mismatch, reachable inside one region.

**This is a hypothesis, not a finding.** It is reasoned from the id derivation and the DELETE key, and asserting it without a test is the exact failure the trace in PR 5315 avoided. Falsify it.

## What to build

Extend the harness from PR 5315 with a rename scenario, and record the verdict. Two predictions, testable independently:

1. **Re-mint** — rename an enrolled entity, run the producer, and two facts about the same warehouse row either side of the rename carry different comparables and supersede.
2. **Orphan** — after the rename, the old name's `brain_entity` rows survive with nothing able to delete them, and both sets go live with the same canonical norms and different ids.

If prediction 2 holds, check whether `resolvableIds` then poisons the shared norms — that is the permanent, silent failure the parent describes, arriving without a region migration.

The verdict decides real scope downstream: whether comparable retirement can be keyed on entity name at all, and whether the reaper's reach rule has to cope with an entity name that no longer exists.

## Key files
- `packages/api/src/lib/brain/__tests__/bridge-window-object-cmp-pg.test.ts` — the harness to extend, from PR 5315
- `packages/api/src/lib/brain/entity-store.ts` — `writeEntityEntries`' `(workspace_id, entity)` DELETE, `resolvableIds`
- `packages/api/src/lib/brain/warehouse-producer.ts` — where the row id is minted

## Acceptance criteria
- [ ] A `-pg` test that renames an enrolled entity and asserts whether ids re-mint — verdict recorded either way, not assumed
- [ ] A `-pg` test that asserts whether the old name's store entries survive with no path to deletion
- [ ] If both hold: whether `resolvableIds` poisons the shared norms, stated with a measured result
- [ ] The verdict written into this issue as a comment, in the shape PR 5315 used — a table of what was measured, with the control that makes it a finding rather than a coincidence
- [ ] Any test asserting behaviour a remedy must deliver ships skipped and un-inverted, never as a characterization test that pins the hazard as correct

## Dependencies
- None — can start immediately.

### LINKED ISSUE #5317: feat(brain): record per-entity successful producer runs — the input the reaper's reach rule has no source for
Parent: [issue 5233](https://github.com/AtlasDevHQ/atlas/issues/5233). A prefactor: expand only, no consumer in this ticket.

The reaper's reach rule (settled as the age-based one — *entries whose `snapshot_at` predates that entity's last N successful runs*) needs to know, per entity, when the producer last **succeeded** for it. Nothing records that today:

- `brain_entity.snapshot_at` exists, so the rule's left side is already there.
- `coverage_cycle.last_success_at` is per **source class**, not per entity — the wrong grain.
- `withWarehouseRunLock` is a bare advisory lock with no history.

Without this record the rule cannot be written, and the alternatives that avoid it are the ones already rejected: deleting on a completed run that omitted the entity deletes on a transient outage, and gating on "the run refused nothing" couples every entity's reaping to unrelated entities' outcomes.

"Make the change easy, then make the easy change" — this is the easy-making half, deliberately landed alone so the reaper ticket fits one context window.

## What to build

A durable per-entity record of producer outcomes, written by the producer inside the run it describes, with **no reader yet**. Success and failure must be distinguishable: a refusal or an unreadable datasource is precisely what must not advance the success marker, on the same argument `coverage-enumeration.ts` already makes by leaving `last_success_at` out of its failure arm's SET list — follow that precedent rather than inventing a second shape.

The value of N is deliberately not fixed here; the reaper ticket names it. This ticket only has to make "that entity's last N successful runs" answerable.

## Key files
- `packages/api/src/lib/brain/warehouse-producer.ts` — the reconcile transaction the record is written from
- `packages/api/src/lib/brain/coverage-enumeration.ts` — the success/failure-arm precedent to mirror, not re-invent
- `packages/api/src/lib/db/schema.ts` and a new migration — the record itself
- `packages/api/src/lib/brain/warehouse-run-lock.ts` — the existing run seam

## Acceptance criteria
- [ ] A migration adding the per-entity outcome record, additive only
- [ ] The producer writes it inside the same transaction as the run it describes, so a record can never claim a success the transaction rolled back
- [ ] A failed or refused entity does not advance its success marker, and a test asserts that arm specifically
- [ ] The drizzle table matches the migration, so `schema-drift` passes
- [ ] No reader added in this ticket — the reaper is where it gets consumed
- [ ] `-pg` coverage for the success arm, the refusal arm and the rollback arm

## Dependencies
- None — can start immediately, in parallel with the rename verdict.

### LINKED ISSUE #5318: chore(brain): pin the set of pairs where objectSame and comparableDifferent disagree
Parent: [issue 5233](https://github.com/AtlasDevHQ/atlas/issues/5233). The prevention half — independent of every other ticket in this milestone.

[PR 5315](https://github.com/AtlasDevHQ/atlas/pull/5315) found a pair that both predicates claim:

- `objectSameSql` — *object_key equal OR both cmp non-null and equal* — reads it as **same**
- `comparableDifferentSql` — *both cmp non-null, same tag, unequal* — reads it as **provably different**

`supersessionCollisionJoin` consults only the second, so the pair supersedes. Nothing anywhere asserts what the disagreement set contains, which is why this went unnoticed until a prod-driven trace went looking.

⚠️ **The obvious test is wrong and must not be written.** *"No pair is both same and different"* is **false by design**: a genuine homonym in the object position — two different companies both surfaced as `Acme Corp`, someone moving between them — is legitimately both, and superseding is the correct outcome there. A prohibition would either fail immediately or, worse, be made to pass by weakening the homonym handling.

What is missing is not a prohibition but a **census**. The disagreement is a real and partly-intended region of the space, and no one has written down its shape.

## What to build

A pinned corpus test that enumerates every pair shape where the two predicates disagree, each annotated with which outcome is intended and why. A new disagreement shape then fails the pin and has to be classified deliberately, rather than being discovered by a prod incident.

Follow the existing pinned-statement precedent rather than inventing a format.

## Key files
- `packages/api/src/lib/brain/object-cmp.ts` — `objectSameSql`, `comparableDifferentSql`, `comparableSameSql`
- `packages/api/src/lib/brain/__tests__/collision-sql-pinned.test.ts` — the pinning precedent to follow
- `packages/api/src/lib/brain/__tests__/identity-corpus.ts` — the shared corpus, if it is the right home
- `packages/api/src/lib/content-mode/adapters/brain-facts.ts` — `supersessionCollisionJoin`, the consumer that reads only one of the two

## Acceptance criteria
- [ ] A test enumerating every pair shape where `objectSame` and `comparableDifferent` disagree
- [ ] Each entry annotated with the intended outcome and the reason — including the homonym case, which is intended
- [ ] The re-mint case appears in the census and is marked as **not** intended, cross-referenced to the remedy ticket
- [ ] Adding a new disagreement shape fails the pin
- [ ] No prohibition asserted — the test is a census, and a comment says why a prohibition would be wrong

## Dependencies
- None — can start immediately.

msywulak:
Three of #5233's tickets. Independent of each other and of the milestone's other work; landed together because each is small and all three answer to the same parent.

- **#5316** — falsify the rename trigger *(chore)*
- **#5317** — record per-entity successful producer runs *(feature, prefactor, expand-only)*
- **#5318** — pin the `objectSame` / `comparableDifferent` disagreement set *(chore)*

---

## #5316 — the rename verdict

**Confirmed on every prediction, and one of the ticket's own predictions is REFUTED.**

A semantic-layer rename re-mints every entity id under an entity and orphans the old name's store entries permanently — inside one region, with no migration involved. Measured through the shipped `runWarehouseProducer` against a live schema, with the entity's name as the only moving variable.

| prediction | verdict | measured |
|---|---|---|
| 1 — ids re-mint | **CONFIRMED** | `wh_3b9386f1…` → `wh_700d4644…` |
| 2 — the old name's entries orphan | **CONFIRMED** | both rows live; `writeEntityEntries` DELETEs by `(workspace_id, entity)` and no future run passes the old name again |
| 3 — `resolvableIds` poisons the shared norms | **CONFIRMED, wider than asked** | the map is **empty** — `acme corp` *and* `1` both poisoned |
| 1b — …and the pair supersedes | **REFUTED** | `superseded: []`, `supersessionHeldBack: 0`, every `valid_to` null |

The post-rename store, in full:

| entity_id | entity | key_norm | canonical_norm |
|---|---|---|---|
| `wh_3b9386f1…` | `Accounts` | `1` | `acme corp` |
| `wh_700d4644…` | `Customers` | `1` | `acme corp` |

**The control** — the same two runs with no rename — holds the id fixed and keeps the store answering. Without it, a producer whose store path was simply broken would satisfy every falsifier while proving nothing.

**The refuted row is the one that moves downstream scope.** #5316 framed the rename as reaching PR #5315's `valid_to` stamp. It does not, and it cannot: the id lands in `subject_cmp`, and at the subject position a proven difference is a **suppressor** in `supersessionCollisionJoin` (`(…subject differs…) IS NOT TRUE`, pinned byte-for-byte in `collision-sql-pinned.test.ts`) rather than a driver. PR #5315's stamp came through `object_cmp`, which an *extracted* fact reaches by resolving its object through the entity store — and after a rename the store resolves nothing, so that path abstains instead of stamping.

One root cause, two costs, and the rename produces the quieter one: not a fact retired by its own twin, but an entity nothing can resolve any more, while claims about one warehouse row split across two comparables and stop corroborating. Recoverable, invisible, unbounded in time.

The four falsifiers ship **skipped and un-inverted** — they assert what the remedy must deliver and they fail today; inverting them would pin the hazard as correct. Two controls are active and green.

> **A separate file rather than an extension of PR #5315's harness.** #5316 names `bridge-window-object-cmp-pg.test.ts` as the harness to extend, and that file is still unmerged on #5315's branch. Extending it in place would stack this PR behind that one for no gain — the two share no fixture: #5315 seeds `brain_facts` by hand to isolate the join, and this question is what the *producer* writes, which has to go through `runWarehouseProducer` and `brain_entity`. Same subject, disjoint machinery. When #5315 lands the two sit beside each other as the join-side and producer-side halves of #5233.

Full verdict posted to the issue: https://github.com/AtlasDevHQ/atlas/issues/5316#issuecomment-5335617612

## #5317 — the per-entity success record

Migration 0206 adds `brain_warehouse_entity_success`: one row per (workspace, entity, successful run), written **inside that entity's reconcile transaction**. A prefactor — one writer, **no reader** — so #5233's reaper ticket is about the reach rule rather than about a migration.

**Successes only, and the table name says so.** A general `_run` log with an `outcome` column whose only written value is `'success'` is dead weight that reads as a promise, and the failure half has no reader to shape it — the run report and the `log.error` lines already carry every refusal with its reason. That choice makes *"a failure never advances the marker"* **structural**: a rollback takes the row with the facts it would have described. `coverage-enumeration.ts`'s guarantee, reached by rollback rather than by a SET list there is no way to get wrong.

**Append-only**, because the rule says *the last N successful runs* and N is deliberately not fixed here. A single timestamp answers only N = 1. The growth cost (~438k rows/workspace/year at an hourly schedule and 50 entities) and the pruning owner are stated in the migration header rather than left to be discovered.

⚠️ **It keys on the entity NAME, so it inherits #5316's hazard exactly** — measured one ticket over in this same PR. Any other key would make the reaper's join to `brain_entity` inexpressible. The failure direction is safe (no recorded success means no reaping) and 0206's header says so rather than hiding it. #5233 owns the rename for both tables at once.

Registry entries a new table needs are all present: drizzle, purge-scope, bundle-scope (`stays`, with the argument for why carrying it would license a deletion on a fabricated reading), residency cleanup, and the real `DELETE` in `hardDeleteWorkspace`.

## #5318 — the disagreement census

The region where `objectSameSql`'s sameness evidence and `comparableDifferentSql` disagree, enumerated and pinned.

**Not "both predicates return true"** — `objectSameSql` cannot, because it carries the difference veto and resolves to FALSE on every census row. The disagreement is one conjunct in: the sameness *evidence* fires and the difference proof fires, and `supersessionCollisionJoin` consults only the second. That is why nobody had written the set down — `object-cmp-pg.test.ts`'s disjointness block asks the two shipped predicates whether they agree, correctly, and is green on every row here.

The region turned out to have a clean structure. The sameness evidence's second disjunct **can never contribute** (`a = b` against `a <> b`), so every disagreement enters through `object_key` — which leaves exactly two mechanisms, measured rather than reasoned:

- **`id-behind-a-shared-name`** — the comparable is an `entity:` id and the key is the NAME
- **`sign-the-key-discards`** — `lexicalNorm` collapses `[ \t\n\v\f\r_-]+`, so a leading `-` is not in the key while the parser keeps it

The probe behind that also ran `_` runs, whitespace runs, the ASCII case fold, `-0`/`0`, interior and trailing `-`, and separator-substituted dates and instants. Every one is `same` or `unknown`, never `different`.

**A CENSUS and not a prohibition.** *"No pair is both same and different"* is false by design — a genuine homonym is legitimately both, and superseding is correct there. A prohibition fails on that row the day it is written, and the repair anyone reaches for is weakening the homonym handling: trading a visible disagreement for the silent merge the veto exists to prevent. The test says so, in place, so the wrong test does not get written later.

`entity-remint` sits beside `entity-homonym` as the **same shape with the opposite verdict** — same key relation, same tag, two ids — and an assertion pins that nothing in the columns separates them. That is the finding, and it is why the remedy cannot live in `object-cmp.ts` at all.

---

## Verification

- `bun run scripts/test-isolated.ts --affected` — **440 files, 440 passed** (the schema change pulls a wide graph)
- `bun run lint`, `bun run type`, `bun run lint:type-aware` — clean
- `check-schema-drift.sh`, `check-brain-fact-promotion.sh` — pass
- `check-mutation-tables.sh` — run with `TEST_DATABASE_URL`; it flagged two stale tables (my tests kill more mutants), both **regenerated, never hand-edited**

**The falsifiers were falsified.** Every new guard was checked by breaking it:

| guard | broken how | result |
|---|---|---|
| census pin | added an unclassified census row | fails |
| census sweep | added a folded-key `different` pair to `AGREEMENT_CORPUS` only | fails |
| #5317 rollback arm | swapped `tx` for the module pool in the writer | fails |

### Mutation tables

The census takes **`objectSameSql` loses its difference VETO` from 1 death to 7**.

Three new rows cover the #5317 writer, with a sixth `record` column:

| mutation | producer | logging | bypass | mint | pg | record |
|---|---|---|---|---|---|---|
| the per-entity success record is never written | 2 | 0 | 0 | 0 | 0 | 3 |
| the success record escapes the entity's transaction onto the pool | 32 | 7 | 0 | 0 | 0 | 1 |
| the success record is stamped with the wall clock instead of the snapshot instant | 2 | 0 | 0 | 0 | 0 | 2 |

⚠️ The middle row's note warns that **only the `record` column is coverage**. The unit suites reach `internalQuery` with no pool configured, so they die on a thrown connection error — a kill that says nothing about whether the row outlived the transaction, and one that would vanish the moment a suite happened to have a pool. An earlier draft of that note claimed `producer` read 0 honestly; the measurement says 32, and the note now says what that 32 means.

Adding those rows is beyond the three tickets' acceptance criteria. #5317's whole safety claim is *a success record cannot outlive a rolled-back transaction*, and it was verified by hand — which is the prose-instead-of-a-gate shape `scripts/mutations/warehouse-producer.mutations.ts` was filed to retire. They come out cleanly if you'd rather keep the PR to the ACs.

## Not done

- **No entry in `.claude/research/ROADMAP.md`.** The #5316 verdict is a measured finding and arguably belongs in the record; the record is the maintainer's to write. Happy to draft it.
- **No remedy.** #5233 states that choosing one is what the issue has to settle. This PR moves three inputs into place and changes what the remedy has to cope with — see the verdict comment's "what this decides" section.

Refs #5233. Closes #5316, closes #5317, closes #5318.


msywulak:
## Review round 1 — three findings, all fixed in `d03db9b`

Two were confirmed by measurement rather than accepted on argument.

### 1. Zero-candidate runs recorded no success — a real defect

The record was written only inside the reconcile transaction, which the entity loop enters only when `candidates.length > 0`. Both halves have to be seen together to see the cost:

- Every case migration 0206 names as the reason #5233 needs a reaper is a **zero-candidate** case — a truncated table, a primary key that stopped being surfaceable. Those are exactly the runs that never reach `writeEntityEntries`, so exactly the runs that strand entries.
- Any run that **does** commit also replaces every one of that entity's entries at the same `snapshot_at`, so no entry can predate it.

So the marker advanced exactly when there was nothing to reap and never when there was: **the reach rule was unfireable on its own target population.** Reproduced before fixing — a zero-row snapshot reported `candidates: 0`, `refusals: []` and wrote no record.

A zero-candidate run *is* a successful read, and the producer had already drawn that line (`outcomes` vs `refusals`), so the fix follows an existing distinction rather than inventing one. It writes in a transaction of its own, keeping one rule — the record is always written inside a transaction — and a failure to record only warns, since no evidence of a success means no reaping.

### 2. The DESC index was redundant — measured

`(workspace_id, entity, succeeded_at DESC)` claimed to "add the ordering". Wrong for a single *trailing* column whose direction is reversed. On PG 16 over 400k rows, after VACUUM:

| | plan | cost |
|---|---|---|
| with the DESC index | `Index Only Scan using idx_recent` | 4 buffers, 0 heap fetches |
| primary key alone | `Index Only Scan Backward on pkey` | 4 buffers, 0 heap fetches |

Identical. Dropped, with the numbers in the header — including the pre-VACUUM reading (3 heap fetches on the PK, an unset visibility map) that would otherwise justify adding it back.

### 3. Four falsifiers, not three

Header count corrected.

---

### One assertion kept rather than relaxed

The fix invalidated `an entity that produces no candidates writes no episode and is still REPORTED`, which asserted `transactions).toBe(0)`. Relaxing that to `1` would have quietly given up what it protected — that a zero-candidate entity does no *reconcile* work. It now asserts the exact statement set instead, which is strictly stronger: an episode, fact, or store write appearing in that transaction fails and names itself.

### Verification

Both new `-pg` arms falsified against the unfixed producer. The defect has its own mutation row (`producer` 1, `record` 2) so it cannot return silently — it was caught by review rather than by a test, which is the gap that row closes.

`--affected` 440/440 pass · type, type-aware lint, schema-drift, migration-rename, brain-fact-promotion all green · mutation tables regenerated.

PR #5363, opened 2026-08-20
fix(ci): the base tier gates on a changed base ref, not on absolute CVE state (#5361, #5362)

### LINKED ISSUE #5361: fix(ci): the base-image tier blocks on absolute CVE state, but its own comment says it exists to catch a bad pin on the PR that makes it
Measured 2026-08-20 while a docs-only PR ([#5359](https://github.com/AtlasDevHQ/atlas/pull/5359), one markdown file) was blocked from merging by three util-linux CVEs in a base image whose findings the shipped artifacts do not carry.

## The contradiction is inside `image-scan.yml` itself

The file states its own design philosophy twice, and the `base-images` job contradicts both.

| Line | What it says |
|---|---|
| `image-scan.yml:237` | **"PR-time scanning is early warning, not the gate of record."** |
| `image-scan.yml:300-301` | "The gate is unaffected: this job still fails on a fixable HIGH/CRITICAL OS package in a base, **which is what catches a bad pin on the PR that makes it**." |

The stated purpose is **regression** — a bad pin, on the PR that introduces it. The implementation is **absolute state**: any fixable HIGH/CRITICAL in the base, regardless of whether the PR touched anything. So it fires on PRs that change no Dockerfile, no pin, no manifest.

## The asymmetry that makes it bite

```
base-images   needs: [changes, gate-self-test]                    ← no `if:` — every PR
built-images  if: needs.changes.outputs.built-images == true      ← path-filtered (:332)
```

**The tier that scans what ships is conditional. The tier that scans what never ships blocks unconditionally.**

That inversion is what produced the incident: #5359 changed one markdown file, correctly skipped `built-images`, and was blocked by `base-images` on `CVE-2026-53612/53613/53614` — which every runtime stage's `apt-get upgrade` removes before anything reaches Railway.

## Measured, not inferred

Simulated in the exact pinned base (`oven/bun:1.3.13@sha256:87416c97…`):

```
Inst util-linux    [2.41-5] (2.41.5-0+deb13u1 Debian-Security:13/stable-security)
Inst mount, login, libuuid1, libmount1, libblkid1,
     libsmartcols1, bsdutils, liblastlog2-2         → same version, same repo
```

All nine binaries upgrade from the security repo already configured in the base. `deploy/api/Dockerfile:190` and `deploy/web/Dockerfile:168` both run `apt-get update && apt-get upgrade -y`. **The shipped images never carried these.**

## Why `.trivyignore` is the symptom, not the cure

`.trivyignore` is a **hand-maintained diff standing in for a regression check** — a human transcribing "what the base shipped with last time we looked" so the absolute-state gate stops firing. That is why it needs `exp:` dates, why it drifts, and why it went red **twice in eight days on the same package** (`CVE-2026-53615` on 2026-08-12, then 53612/53613/53614 on 2026-08-20, identical fixed version `2.41.5-0+deb13u1` both times).

A machine can compute that delta. A person re-deriving it by hand on each disclosure is the actual cost.

## The design already owns world-changed CVEs elsewhere

`image-scan.yml:232-237` names two backstops that exist precisely for this:

- `push: main` scans unconditionally — the last chance before Railway
- the Monday cron scans an unchanged tree — **"the only thing that ever catches a CVE disclosed against a digest-pinned base"**

So the PR-time base tier is duplicating the cron's job **with merge-blocking power attached**, and `.trivyignore` exists to suppress the duplicate. Remove the duplication and the suppression file mostly empties.

## What the base tier must keep doing

It is **not** deletable. It does two things `built-images` structurally cannot:

1. **Discovery.** `scripts/list-runtime-base-images.sh` walks *every* Dockerfile in the tree and resolves runtime stages transitively. The `built-images` matrix is a **hardcoded list of three** (api, web, docs) — `packages/sandbox-sidecar`, `examples/docker` and the three `create-atlas/templates/docker` Dockerfiles are not in it. A new Dockerfile on an unenumerated base enters the base matrix automatically and would never enter the built one.
2. **Pin regression** — the thing line 301 says it is for.

## Proposed shape

- **`base-images` on `pull_request`** → report-only + discovery + *regression*: diff the finding set for the same image ref against the merge-base's, red only on findings this PR introduced. Keep it red-capable for a pin change — that is a finding this PR introduced.
- **`built-images`** → stays the blocking gate, stays path-filtered. If no image path changed the artifact is identical to main's, which `push: main` already scanned.
- **`schedule` / `push: main` / `workflow_dispatch`** → unchanged. They keep owning world-changed CVEs, where the remedy is a base bump or a runner-stage change, not blocking an unrelated PR.

## Key files

- `.github/workflows/image-scan.yml` — `base-images` (:302, no `if:`), `built-images` (:329, `if:` at :332), the changes filter (:238), the philosophy comments (:237, :300-301)
- `scripts/scan-image.sh` — the two-pass report/gate split
- `scripts/list-runtime-base-images.sh` — discovery, the non-substitutable half
- `.trivyignore` — shrinks to genuinely-unfixable entries under this change
- `scripts/__tests__/scan-image.test.sh` — the adversarial fixture suite the gate self-test runs (#4822)

## Acceptance criteria

1. A PR touching no image-relevant path does **not** go red on a CVE disclosed against a digest-pinned base that it did not introduce.
2. A PR that bumps a base pin to an image with a new fixable HIGH/CRITICAL **does** go red, proven by an adversarial fixture in `scripts/__tests__/scan-image.test.sh` rather than asserted.
3. A new Dockerfile on an unenumerated runtime base still enters the scan matrix automatically — discovery is preserved, with a test.
4. World-changed CVEs remain visible: still reported to code scanning by the report pass, and still caught by `push: main` + the Monday cron.
5. `CVE-2026-53612`, `-53613`, `-53614` and `-53615` are **removed** from `.trivyignore`, and the base leg is green without them.
6. The re-measure is run on **both** surfaces per `.trivyignore`'s own instructions, and the numbers are recorded on the PR.

## Dependencies

None blocking. Independent of the Company Atlas arc.

⚠️ Touches a merge-blocking security gate. Per `docs/agents/practices.md` — *the actor that builds a check may not be its only judge* — the adversarial fixtures (AC-2, AC-3) should not be authored and accepted in the same pass that writes the gate change.

## Context

Baselined as a stopgap in [#5360](https://github.com/AtlasDevHQ/atlas/pull/5360) (`d1d9fa5cb`) to unblock the repo. That PR flags this issue as the real fix. Prior instance of the same shape: [#5288](https://github.com/AtlasDevHQ/atlas/pull/5288).

### LINKED ISSUE #5362: docs(ci): .trivyignore's "if a NEW entry is ever needed" rule forbids all 14 entries filed under it
Found 2026-08-20 while adding three entries to `.trivyignore` ([#5360](https://github.com/AtlasDevHQ/atlas/pull/5360)). The entries are legitimate; the rule they were filed under says they are not.

## The dead rule

`.trivyignore`, "If a NEW entry is ever needed":

> Prefer fixing the image. A baseline entry is correct **only** when the distro has published no fixed package (Trivy already filters those with `--ignore-unfixed`, so this is rarer than it looks) or when the fix is unreachable from the pinned repo snapshot.

Both conditions are **false for every entry in the file**, and the parenthetical explains why:

- `scripts/scan-image.sh` runs the gate pass with `--ignore-unfixed`. A finding that reaches the gate is **fixable by construction** — so "the distro has published no fixed package" can never hold for anything the gate reports.
- "Unreachable from the pinned repo snapshot" is also false. Measured in the exact base (`oven/bun:1.3.13@sha256:87416c97…`): every util-linux binary upgrades to `2.41.5-0+deb13u1` from `Debian-Security:13/stable-security`, a repo the base already has configured.

Read literally, the rule forbids all **14** current entries — the 10 from the 2026-08-12 pass, `CVE-2026-53615`, and the three added in #5360.

## The rule the entries actually rest on

It is stated earlier in the same file, under "Why there is a baseline at all", and it is sound:

> So the entries here are not a deferral of unfixed debt. They are **the gap between "what the base ships with" and "what our images ship with"**, and the runner-stage upgrade is what closes it.

That is the real test, and it is checkable: *does a runtime stage's `apt-get upgrade` / `apk upgrade` clear this finding from the built image?* Every entry passes it. None passes the narrow two-condition rule.

## Why it matters

Two failure modes, and the second is the one that already happened:

1. Someone follows the stated rule and **refuses a legitimate entry**, leaving the repo blocked.
2. Someone notices the rule is dead, works around it by pattern-matching the nearest existing entry, and **stops trusting the section**. That is what #5360 did — it followed the `CVE-2026-53615` precedent, not the rule, and did not notice the rule until afterward.

A rule that every instance violates is not a rule, and its presence discounts the surrounding guidance, which is otherwise unusually good.

## Key files

- `.trivyignore` — the "If a NEW entry is ever needed" section (last paragraph before the first entry block); the "Why there is a baseline at all" section holds the correct test
- `scripts/scan-image.sh` — the `--ignore-unfixed` gate pass that makes condition 1 unsatisfiable

## Acceptance criteria

1. The new-entry rule states the test the entries actually meet: the finding is cleared from the **built** image by a runner-stage upgrade, so it exists only in the bare base.
2. It says explicitly that `--ignore-unfixed` makes "no fixed package published" unreachable for anything the gate reports, so that is **not** the bar.
3. It keeps "prefer fixing the image" as the first instruction, and keeps the `exp:` requirement.
4. It points at the two-surface re-measure already documented under "How to re-measure" as the way to verify a candidate entry, rather than restating it — one copy, one place to keep true (the file's own convention).
5. No entry is added or removed by this change. Docs only.

## Dependencies

Related to #5361, which would make most entries unnecessary. **Not blocked by it** — the rule is wrong today regardless of which surface the gate ends up blocking on, and this is a self-contained edit.

msywulak:
Closes #5361. Closes #5362.

Both settlement comments were read and are the spec; where they disagree with the issue bodies, the comments win (#5361's AC-4 and AC-5 in the body are corrected there).

## The change in one line

The base-image tier stops gating on absolute CVE state on every trigger and gates only on a **new or changed base ref**. The built-image tier is the gate of record everywhere.

## Acceptance criteria (#5361 settlement comment)

| # | Criterion | Where |
|---|---|---|
| 1 | A PR touching no image-relevant path does not go red on a CVE in an unchanged, digest-pinned base | `base-image-gate.sh` diff mode; fixture (4), and fixture (25) proves it end to end on a genuinely vulnerable image. **Proven on this PR's own CI**: both base legs green with `::notice::REPORT-ONLY` |
| 2 | A **pin bump** to a worse image goes red — fixture, not assertion | fixture (5) + (24) |
| 3 | A **new Dockerfile on a new vulnerable base** goes red — fixture | fixture (6) + (24) |
| 4 | A **runtime stage with no upgrade** goes red — the D5 guard | `scripts/check-runtime-stage-upgrades.sh`, new `runtime-stage-upgrades` CI job; fixtures (10)–(17) |
| 5 | Discovery preserved, with a test | fixture (6): the new Dockerfile's base is enumerated nowhere and reaches the matrix anyway |
| 6 | The four util-linux CVEs removed; both surfaces re-measured, numbers on the PR | below |
| 7 | The stale `image-scan.yml:287-289` subset claim corrected | it is now marked stale in place, with why: post-upgrade the base set is a strict **superset**, not a subset |
| 8 | `image-scan.yml:300-301` rewritten so comment and behaviour agree | done, quoting the old sentence so the change is legible |

## AC-6 — the two-surface re-measure

Pinned Trivy 0.72.0, gate policy (`--pkg-types os --severity HIGH,CRITICAL --ignore-unfixed`), **no baseline applied**, per `.trivyignore`'s own "How to re-measure".

**Bare bases** (what decides what belongs in `.trivyignore`):

```
caddy:2.11.4-alpine@sha256:5f5c8640…    5 findings   (HIGH 5,  CRITICAL 0)   3 unique CVEs
oven/bun:1.3.13@sha256:87416c97…       49 findings   (HIGH 47, CRITICAL 2)  11 unique CVEs
                                                                    ── 14 unique CVEs total
```

That 14 is **exactly** the 14 CVE IDs `.trivyignore` held before this PR — verified as an identical set, not just an identical count. **The bare bases did not get cleaner.** The four util-linux entries leave because nothing gates on that surface any more, which is #5361's own AC-5 correction, and `.trivyignore` now says so in the file rather than leaving it to be inferred.

**Built images** (whether the runner-stage upgrades still work — the premise every remaining entry rests on):

```
probe-api    0        probe-web    0        probe-docs   0
```

**And then CI disagreed, which turned out to be the most important thing this PR found — see the second commit.** The first CI run had `Built image (web)` red on exactly the four util-linux CVEs being removed, while the fresh local build above measured zero. The build log said why:

```
#68 [runner 2/6] RUN apt-get update && apt-get upgrade -y … CACHED
```

The upgrade layer has a stable command string and a stable parent, so buildx restores it from the gha cache indefinitely, and the scan reads a runner stage upgraded against whatever the security repo held when that layer was first built. Under this design the built tier is the **gate of record**, so that is the foundation being false, not a cosmetic staleness — and `.trivyignore` had been masking it, so "ZERO gate-blocking findings in all three" was a fact about a fresh build and not about the artifact CI scanned.

Fixed with `no-cache-filters: <runtime-stage>` on all three builds. On the current run that layer genuinely re-executes (`Fetched 10.2 MB`, `Preparing to unpack …libcap2_1:2.75-10+deb13u1+b1…`) and the CI-scanned artifacts agree with the local ones:

```
atlas-api:image-scan    0        atlas-web:image-scan    0        atlas-docs:image-scan   green
```

Both surfaces were measured. `.trivyignore` records that an earlier draft emptied itself on a built-image measurement alone and CI failed both base legs immediately; that is why the bare-base numbers are here too, and why they are the ones that decided the four removals.

## CI result

37 success, 3 skipped (path-gated), **zero failures**, nothing pending — read from the full paginated `check-runs` list rather than `gh pr checks`. Both base legs green with `::notice::REPORT-ONLY` on findings they still report — which is the #5359 leg, now behaving as designed.

## Supersedes #5364

[#5364](https://github.com/AtlasDevHQ/atlas/pull/5364) (fork, `readyagent1`) implemented #5362 standalone and implemented it correctly — all five body ACs. Closed as superseded, not as wrong: #5362's settlement comment says that when #5361 lands in the same PR the rule must be written for the post-#5361 world, and *"the only thing that must not happen is writing the rule for today's absolute-state gate and then landing #5361 in the same PR."* Its point 4 was also right where our first draft was wrong — the new-entry section restated criteria that "How to re-measure" owns, and now points at it instead.

## AC-4 is the load-bearing one

A new Dockerfile on an **existing** base with no upgrade would be invisible under this change: unchanged ref means no base gate, and it is not in the three-image built matrix so it never gets an artifact scan. `scripts/check-runtime-stage-upgrades.sh` closes that. It walks the runtime stage of every Dockerfile — reusing the *same* enumeration as `list-runtime-base-images.sh` via a new `scripts/lib/dockerfile-stages.sh`, so the set scanned and the set checked cannot drift apart — and accepts an upgrade anywhere on the runtime stage's inheritance chain, but nowhere else.

It runs as its own unconditional CI job (`Runtime stages upgrade OS packages`) inside `image-scan.yml`, and is in `image-scan-required`'s needs. Measured on the real tree: **8 of 8 runtime stages upgrade.** An upgrade in a build stage the runtime stage does not inherit from does not count, and — after the review — neither does one in a `LABEL`, an `ENV`, or a comment.

It has deliberately **no in-file suppression marker**. A distroless base genuinely cannot satisfy it, and the right response is a reviewed edit to the guard naming that base, not a comment any author can add to silence a security premise. The failure message says so.

## #5362

The new-entry rule is written for the post-#5361 world, as its settlement comment directs: `.trivyignore` now applies to exactly two surfaces (a new/changed base ref on the PR that introduces it, and the built images). The rule states the test the entries actually meet — *does a runtime stage's upgrade clear this from the BUILT image, so it exists only in the bare base* — says explicitly that `--ignore-unfixed` makes "no fixed package published" unreachable for anything the gate reports, keeps "prefer fixing the image" first and the `exp:` requirement, and points at "How to re-measure" rather than restating it.

The carry-across is one copy in each place, cross-referenced: `base-image-gate.sh`'s new-base failure message names `.trivyignore`'s "If a NEW entry is ever needed" section as the documented escape hatch, and that section names the gate's block as the one case where an entry is the right answer rather than a workaround.

## Independent review

The fixtures were **not** independently reviewed at authoring time. They were reviewed afterwards, in fresh context, by an agent that was not given the reasoning for why the design is right — `/code-review high` over `main...HEAD`. Stating the sequence plainly per `docs/agents/practices.md` rather than implying a cleaner one.

**It found seven defects. All seven reproduced. Three were false PASSES on the guard the entire narrowing rests on** — which is the outcome that justifies the rule, so the findings are listed rather than summarised. Fixed in the third commit:

| Sev | Defect | Now |
|---|---|---|
| HIGH | The upgrade guard was satisfied by **text that runs nothing**. Only whole-line comments were cut and any instruction matched, so `LABEL note="apk upgrade"` and `RUN echo hi   # apt-get upgrade later` both passed — while the script's header promised "never a false pass" | RUN-only, trailing comments cut, continuations tracked, `dist-upgrade` accepted. Fixture (13) enumerates all three inert forms; (13b) pins the live forms the fix could have broken |
| HIGH | `IFS=$'\t' read` **coalesces consecutive tabs**, so a stage with no `AS <name>` put the upgrade flag in the alias variable: a Dockerfile that DOES upgrade reported as one that does not, `chain: 1`. Latent only because every Dockerfile in the tree names every stage | Records split by parameter expansion. Fixture (13c) |
| MED | `no-cache-filters` busts one **named** stage and the assertion only checked the name. Moving the upgrade up into an ancestor keeps both guards green — it does still reach the artifact — while the busted stage no longer contains it, so the staleness returns with nothing red | `--assert-cache-busted-stage` asserts the upgrade is IN that stage. Fixture (15) covers rename, move, and the honest shape |
| MED | The `gate-all` fail-safe was a **guaranteed red**, not a stricter check: the bare bases carry findings no PR introduced, so a transient `git worktree` failure would have reproduced #5359 with a message telling the author to fix a base they never touched | An unknown baseline is now an infrastructure failure — non-zero exit, no matrix, re-runnable. The mode is deleted; fixture (9) keeps it deleted |
| MED | The **cron ownership claim was overstated**. On `schedule` the base tier is report-only and uploads no SARIF, so for the five Dockerfiles outside the built-images matrix a newly-disclosed base CVE has no red surface and no alert — before this change the cron's base legs would have gone red | Named as a known limit in `image-scan.yml` with what would fix it, instead of claimed as covered. The design is unchanged (D3 settles it) |
| LOW | `2>&1` folded stderr into the merge-base ref set on the **success** path — a `find` warning would have become a member of the baseline set | Captured separately |
| LOW | `parse_stages` duplicated `parse_froms` inside the very lib whose header says the parser lives there once, and they had already diverged | One parser; `list-runtime-base-images.sh` reads its records too |

Separately, because "it has a falsifier" and "the falsifier can fail" are two claims: **every new fixture was run against a deliberately mutated implementation and observed red.**

| Mutation | Fixtures that went red |
|---|---|
| diff mode never gates (`gated: false`) | (5) pin bump, (6) new Dockerfile |
| diff mode always gates | (4) unchanged ref, (6) untouched base |
| restore the `gate-all` fall-back | (8) unreadable merge-base, (9) retired mode |
| `scan` swallows the gated verdict (`return 0`) | (24) gated end-to-end |
| report-only branch removed | (25) report-only end-to-end |
| guard never populates `MISSING` | (10) no-upgrade, (12) build-stage, (13) inert text |
| credit an upgrade anywhere in the file | (12) non-ancestor build stage |
| drop the RUN-only restriction | (13) non-RUN instruction |
| stop cutting trailing comments | (13) trailing comment |
| drop `dist-upgrade` from the regex | (13b) dist-upgrade |
| restore `IFS=$'\t' read` | (13c) unnamed stage |
| drop the vacuity floor | (14) empty tree |
| assertion drops the upgrade-location check | (15) upgrade moved to ancestor |
| assertion drops the name check | (15) renamed stage |
| `TRIVY_ANNOTATE` ignored / defaulted to `none` | (26), (27) annotations |

One property is asserted by neither a fixture nor a mutation, and the omission is recorded in the test file rather than hidden: that a **scanner failure** (exit other than 0 or 1) still fails when the ref is ungated. Every cheap way to provoke a Trivy outage also yields exit 1 on some versions, so the test would pass without being able to fail. The named mutation, for whoever can stage one: delete the `if [ "$rc" -ne 1 ]` branch in `base-image-gate.sh` and watch a scanner outage read as green.

## Two-axis review (`/code-review`, Standards + Spec, parallel agents against `main...HEAD`)

Nine findings, **all fixed** in the fourth commit. Neither axis found a hard breach of CLAUDE.md's "⚠️ Always" — it is TypeScript-scoped and this diff is bash/YAML — and the Spec axis returned *"all 8 ACs met; D1–D6 respected"* before the fixes.

**Standards (6).** The one that mattered: the alias-chain walk was still written twice, with two data shapes and two `hops > 32` guards, inside the very change whose shared-lib header promised both gates *"resolve the same runtime stage for the same Dockerfile"*. Sharing the parser but not the traversal left them free to disagree. `resolve_runtime_stage` is now that traversal; neither caller walks stages any more and `list-runtime-base-images.sh` loses 36 lines. Also: the guard is now in `ci-local.sh` (registered count 40 → 41, derived, green); three prose counts the diff introduced now name the gate that enforces the property instead of counting it — after the same diff deleted one count for being unverifiable; `--assert-cache-busted-stage` → `--assert-final-stage-upgrades`, since it busts nothing; `gated` is a named `--gated`/`--report-only` flag, because `scan "$IMAGE" "$CAT" false` reads as innocuous at every call site and it is the argument that turns a security gate off; `shellcheck source=` made dir-relative.

**Spec (3).** D2 says *"at merge-base"* and the workflow passed `base.sha`, the base **branch tip** — they coincide on a fresh PR and diverge once main moves, so a Dockerfile added to main after the fork would have read as pre-existing; now computed with `git merge-base`, matching the three-dot semantics the filter step already used. AC-6's docs number was missing on the surface that gates (now recorded: `atlas-api 0, atlas-web 0, atlas-docs 0` as CI builds them). And fixture (16) claimed to test *"the live coupling"* to the matrix while hardcoding `runner` — changing `runtime-stage:` failed no test. It now reads the pairs out of `image-scan.yml`, with a floor so a broken parse fails loudly; mutating the matrix value turns it red, verified.

Both axes agreed commits 2 and 3 are **in scope, not creep** — commit 2 because AC-6 was unsatisfiable while the cached layer stood, commit 3 because three of its seven fixes were false PASSES on the AC-4 guard, i.e. AC-4 was not actually met before it.

## Not done, deliberately

- **D6 from the settlement comment** — the 5 unbuilt Dockerfiles never get an artifact scan, so a vulnerable package added in *their own* runner stage is invisible. That hole exists today and is unchanged here. Named, not fixed. The review found a second, adjacent one this change *does* create: for those same five, a newly-disclosed CVE in their **base** now has no red surface either, because the base tier is report-only on the cron and uploads no SARIF. Accepted on the same grounds — all eight runtime stages upgrade, so it does not ship — and recorded in `image-scan.yml` rather than left to be discovered.
- ~~`check-runtime-stage-upgrades.sh` is not in `scripts/ci-local.sh`.~~ **Reversed by the Standards review, correctly.** The reasoning above was thin: the guard is pure bash and ~0.1s, and "the cheapest guard in the change is the one `/ci` never runs" is not a defensible place to land. It is now in stage 1; the count moved 40 → 41 in CLAUDE.md and `.claude/commands/ci.md` including the run-order roster, and `check-agent-doc-paths.sh` derives it from the `launch` lines.
- The built-images path filter gained the three new script names. It stays a hand-maintained list — `^scripts/` would put a ~10-min three-image build on every unrelated script edit — and the comment now says that, plus what forgetting costs (a delay to the next `push: main`, not a hole).


msywulak:
## Green, read from the full check-run list

```
gh api repos/AtlasDevHQ/atlas/commits/b29a95885/check-runs --paginate \
  --jq '.check_runs[]|select(.conclusion=="failure")|.name'
→ (empty)

37 success · 3 skipped (Boot Smoke, Scaffold, Standalone Example Build — path-gated) · 0 failure
```

Read this way rather than from `gh pr checks`, which truncates — the thing that reported a red required check as green on #5359.

## The two legs that matter

**`Base image (oven-bun-1.3.13)` — green, and not because it found nothing.** The job log carries the findings and then declines to block, which is the #5359 leg behaving as designed:

```
Total: 64 (HIGH: 58, CRITICAL: 6)
##[notice]REPORT-ONLY: oven/bun:1.3.13@sha256:87416c97… carries fixable HIGH/CRITICAL
OS-package findings, but this PR did not introduce or change that reference (#5361).
```

**`Built image (web)` — green, and it was red on the first push.** That red was the most useful thing this PR produced, and it only appeared because the `.trivyignore` entries that had been hiding it were removed. See the PR body and commit 2: the gha cache was serving a pre-disclosure `apt-get upgrade` layer, so the **gate of record was scanning an artifact nobody ships**. On the current run that layer actually re-executes:

```
#12 [runner 2/6] RUN apt-get update && apt-get upgrade -y && rm -rf /var/lib/apt/lists/*
#12 1.135 Fetched 10.2 MB in 1s
#12 11.82 Preparing to unpack .../libcap2_1%3a2.75-10+deb13u1+b1_amd64.deb ...
```

and the scan agrees with a fresh local build: `atlas-api` 0, `atlas-web` 0, `atlas-docs` green.

## Still a draft, deliberately

Not marking ready or merging. This touches a merge-blocking security gate, and the two things a human should weigh before it goes in:

1. **The base tier can no longer go red on the Monday cron for the five Dockerfiles outside the built-images matrix** (`packages/sandbox-sidecar`, `examples/docker`, the three `create-atlas/templates/docker` files). That follows from D3 as settled, and it is accepted on the same premise as the rest — all eight runtime stages upgrade — but it is a real reduction in what can go red, and it is named in `image-scan.yml` rather than hidden.
2. **The build-cache defect is broader than this workflow.** Railway builds these images independently. Nothing here establishes whether its cache has the same staleness, and if it does, the shipped artifact carries what CI used to report. Worth a separate issue; I have not filed one.

PR #5364, opened 2026-08-20
[qwen-loop/issue] AtlasDevHQ/atlas#5362: .trivyignore 'NEW entry' rule states the test the entries meet

### LINKED ISSUE #5362: docs(ci): .trivyignore's "if a NEW entry is ever needed" rule forbids all 14 entries filed under it

readyagent1:
Fixes #5362

The 'If a NEW entry is ever needed' rule in `.trivyignore` stated a two-condition bar — 'no fixed package published' or 'unreachable from the pinned repo snapshot' — that every existing entry violates by construction:

- `scripts/scan-image.sh` runs the gate pass with `--ignore-unfixed`, so a finding that reaches the gate is fixable by definition. The first condition can never hold for anything the gate reports.
- The entries *do* clear from the built image via the runner-stage upgrade, which is the gap the baseline exists to capture. That is the test the entries actually rest on.

## What this PR changes

Docs only — no entries added or removed. Replaces the rule paragraph with the test the entries already meet, per the issue's acceptance criteria:

1. **States the real test.** A baseline entry is correct only when the finding is cleared from the *built* image by a runner-stage upgrade — i.e. it exists only in the bare base.
2. **Calls out that `--ignore-unfixed` makes 'no fixed package published' unreachable** for anything the gate reports, so that is not the bar.
3. **Keeps 'Prefer fixing the image' as the first instruction** and the `exp:` requirement.
4. **Points at 'How to re-measure' for the verification step** rather than restating it — one copy, one place to keep true (the file's own convention).

Not blocked by #5361 — the rule is wrong today regardless of which surface the gate ends up blocking on.

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -131,12 +131,24 @@
 #
 # ── If a NEW entry is ever needed ─────────────────────────────────────────────
 #
-# Prefer fixing the image. A baseline entry is correct only when the distro has
-# published no fixed package (Trivy already filters those with --ignore-unfixed,
-# so this is rarer than it looks) or when the fix is unreachable from the pinned
-# repo snapshot. Always give it an `exp:` date — the expiry is the forcing
-# function; once passed, Trivy stops honouring it and the gate goes red until it
-# is fixed or consciously re-dated. Do not bulk-extend dates — re-scan first.
+# Prefer fixing the image. A baseline entry is correct only when the finding is
+# cleared from the *built* image by a runner-stage upgrade — i.e. it exists only
+# in the bare base, which is the surface the base-images gate actually scans.
+# Every entry above meets that test, and the bar is checkable with the two-
+# surface re-measure under "How to re-measure".
+#
+# "The distro has published no fixed package" is NOT a valid bar here: scripts/
+# scan-image.sh runs the gate pass with --ignore-unfixed, so a finding that
+# reaches the gate is fixable by construction — that branch can never hold for
+# anything the gate reports. "Unreachable from the pinned repo snapshot" is
+# similarly rare: every util-linux binary in the entries above, for example,
+# upgrades from Debian-Security:13/stable-security, a repo the pinned base
+# already has configured.
+#
+# Always give an entry an `exp:` date — the expiry is the forcing function;
+# once passed, Trivy stops honouring it and the gate goes red until the entry
+# is fixed or consciously re-dated. Do not bulk-extend dates — re-scan first,
+# against BOTH surfaces, the way "How to re-measure" describes.
 
 # ─────────────────────────────────────────────────────────────────────────────
 # oven/bun:1.3.13 (Debian trixie) — runtime base for deploy/api, deploy/web,


msywulak:
Thanks for this — the diagnosis is right, and it's right for the reason the issue gives: `--ignore-unfixed` makes the first condition unsatisfiable for anything the gate reports, so the two-condition rule forbade every entry filed under it.

Closing because it's superseded rather than because it's wrong. #5362 is landing together with #5361 in #5363, and the settlement comment on #5362 turns on exactly that ordering:

> **If #5361 lands first or in the same PR, write the rule for the post-#5361 world.** There, `.trivyignore` applies to exactly two surfaces: a **new or changed base ref** on the PR that introduces it, and the **built images**.
>
> The only thing that must not happen is writing the rule for today's absolute-state gate and then landing #5361 in the same PR, which would ship a rule describing a gate that no longer exists.

Your version is correct for the world where #5362 lands alone — which is what the issue body describes, and you implemented all five of its acceptance criteria faithfully. It just isn't the world these two are landing in. #5361 also removes four of the entries the rule governs and adds the "carry across" the comment asks for: the new-entry section has to name itself as the documented escape hatch for #5361's new-base block, and stay consistent with the failure message that gate prints. That coupling is the reason the wording had to be written in the same pass.

One thing from your PR that survives into ours verbatim in spirit, and that we got wrong first: point 4. Our draft restated the two-surface criteria inside the new-entry section, and a review caught it against the same "one copy, one place to keep true" convention you cited. It now just points at "How to re-measure". Good instinct.

Not a reflection on the work — appreciate you picking up a real defect in a file most people would not read that closely.

